PR #3, opened 2024-05-02
skeleton for the control cli tool and the connection daemon

### LINKED ISSUE #2: CACTL <-> CD skeleton
Create the CACTL command line tool and a CD daemon and hook them up so that they can communicate.  The cactl should accept the `connect`, `disconnect`, and `status` commands and the CD should fabricate replies and create logfile entries.

Part of this story is also thinking a bit about how configurations will work with cactl.

mkolehmainen:
This is working but I would like to add some unit tests just to see how that works too in rust.


adapter/cd/src/main.rs
```diff
@@ -0,0 +1,88 @@
...
+    });
+
+    if foreground {
+        return cd::tokio_main(config.clone());
+    }
+
+    // Else we go into background.
+
+    let logpath = Path::new(LOG_DIR);
+    if !logpath.exists() {
```
cpacejo: `exists` check is not needed I think?  (`create_dir_all` will succeed if the path already exists, like `mkdir -p`)

adapter/cd/src/main.rs
```diff
@@ -0,0 +1,88 @@
...
+        .append(true)
+        .truncate(false)
+        .open(format!("{}/cd.err", LOG_DIR))?;
+    writeln!(
+        stderr,
+        "=============== cd restarts at {} ==============",
+        chrono::Local::now()
+    )?;
+
+    let daemonize = Daemonize::new()
```
cpacejo, line 75: Rather than "old-school" daemonization I wonder whether it'd be better to support systemd-style daemonization?  (Which IIRC involves, instead of most of this code, just sending a message to systemd saying we've started, e.g.: <https://docs.rs/libsystemd/latest/libsystemd/>)

adapter/cd/src/cd/zpr.rs
```diff
@@ -0,0 +1,400 @@
...
+            }),
+        }
+    }
+
+    // If a configuration exists with the same path, we overrite the existing one but only if it is disconnected.
+    pub fn add_configuration(&self, c: Configuration) -> Result<(), std::io::Error> {
+        let mut found = false;
+        let mut state = self.shared.configurations.lock().unwrap();
+        for (conf, state) in &*state {
+            // XXX <------- RUST WTF IS THIS "&*" ?
```
cpacejo: The `*` is needed to dereference the `MutexGuard`; the `&` is then needed to iterate by reference rather than copy.

adapter/cd/src/cd/zpr.rs
```diff
@@ -0,0 +1,400 @@
...
+                    return Err(Error::new(
+                        ErrorKind::Other,
+                        "Configuration already exists and is not disconnected",
+                    ));
+                }
+            }
+        }
+        if found {
+            // Remove the existing configuration
+            state.retain(|(conf, _)| conf.path_name != c.path_name);
```
cpacejo: Should this be `HashMap` instead of a `Vec`?

adapter/cd/src/cd/zpr.rs
```diff
@@ -0,0 +1,400 @@
...
+    configurations: Mutex<Vec<(Configuration, ConfigState)>>,
+}
+
+impl Configuration {
+    pub fn get_name(&self) -> &str {
+        self.profile.name.as_str()
+    }    
+}
+
+// TODO: Rust what is this "default" stuff?
```
cpacejo: I'm guessing used to create a struct when no initializers are supplied.

adapter/cd/src/cd/zpr.rs
```diff
@@ -0,0 +1,400 @@
...
+    };
+    c.path_name = path.to_string();
+    Ok(c)
+}
+
+
+
+#[derive(Debug, Clone)]
+pub struct Zpr {
+    shared: Arc<Shared>,
```
cpacejo, line 81: I don't have super strong feelings here, but my gut tells me that putting `shared` on the heap here opens the door for "zombie" tasks (i.e. tasks which outlive the `Zpr` struct itself).  Whereas if `shared` is defined as `shared: Shared` instead, then we can lean on Rust's lifetime checking to ensure the invariant that the `Zpr` struct outlives the tasks.

adapter/cd/src/cd/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+};
+
+use tokio_util::{
+    sync::CancellationToken,
+    task::TaskTracker,
+};
+
+
+#[tokio::main]
+pub async fn tokio_main(config: Arc<Config>) -> io::Result<()> {
```
cpacejo, line 34: Why require `config` to live on the heap here?

adapter/cd/src/cd/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+        }
+        let _ = sig_shutdown_tx.send(());
+    });
+
+    let (cs_shutdown_tx, mut cs_shutdown_rx) = oneshot::channel();
+    let cs_config = config.clone();
+    let cs_token = token.clone();
+    tracker.spawn(async move {
+        match command_server(cs_config, zpr.clone(), cs_token).await {
+            Ok(()) => {
```
cpacejo, line 48: Why would this exit gracefully before the cancellation token is sent?

I feel like the loop on line 80 should be its own task (or merged into the task above which listens for signals).  Else we block on like 79 waiting for the command server to exit _before_ we start that loop (and have the ability to send it a cancellation token).

adapter/cd/src/cd/command_server.rs
```diff
@@ -0,0 +1,261 @@
...
+        AsyncWriteExt,
+    },
+    net::UnixListener,
+};
+use tokio_util::sync::CancellationToken;
+
+pub use crate::cd::config::Config;
+pub use crate::cd::zpr::{load_configuration, Zpr, ConfigState};
+
+pub async fn command_server(config: Arc<Config>, zpr: Zpr, token: CancellationToken) -> io::Result<()> {
```
cpacejo, line 22: Again here I feel `config` should be a `&` not an `Arc<>`.  `command_server` doesn't make sense to me to be the last one "responsible" for the config should all other references be dropped.

adapter/cd/src/cd/command_server.rs
```diff
@@ -0,0 +1,261 @@
...
+        tokio::select! {
+            _ = token.cancelled() => {
+                info!("command server is cancelled");
+                return Ok(());
+            }
+            result = listener.accept() => {
+                match result {
+                    Ok((stream, _)) => {
+                        info!("accepted command connection");
+                        let zpr = zpr.clone();
```
cpacejo, line 35: This feels wrong… I think we should be passing references to `zpr`, no?

adapter/cd/src/cd/command_server.rs
```diff
@@ -0,0 +1,261 @@
...
+                        tokio::spawn(async move {
```
cpacejo, line 36: Should these tasks be collected into a `TaskTracker`?

adapter/cd/src/cd/command_server.rs
```diff
@@ -0,0 +1,261 @@
...
+    line.clear();
+    let n = reader.read_line(&mut line).await?;
+    if n > 0 {
+        let line = line.trim();
+        if line.is_empty() {
+            error!("empty line received");
+        } else {
+            let parts: Vec<&str> = line.split_whitespace().collect();
+            match parts[0] {
+                "status" => handle_status(Arc::clone(&writer), zpr).await?,
```
cpacejo, line 80: Why `Arc<>` and `Mutex<>` for `writer`, vs. passing `&mut` references?  It's only ever used sequentially (i.e. not concurrently) and its lifetime doesn't outlive this function from what I see?

cpacejo (approved):
Looks good mostly I have comments on using `Arc` and `Mutex` where it seems references would suffice.  Also what looks like a deadlock in graceful shutdown.  Approving though for now not to block you.

adapter/cd/src/cd/command_server.rs
cpacejo: Ah now having tried to do this, I understand more :)  Tokio wants all tasks to have lifetime of `'static` which really puts a damper on lifetime hygiene IMO.  What I've done for the time being is dereference the `Arc` first thing in the async block which calls my functions, so I don't have to pass the `Arc` around everywhere.  I don't love it though...

adapter/cd/src/main.rs
mkolehmainen: Yes good idea. I've made a task to explore this later.

PR #5, opened 2024-05-16
add startmeup support to the nascent CD

### LINKED ISSUE #4: CD can perform start-me-up with node
A connection is initiated with CACTL and the CD performs start-me-up sequence with the dock.  This story is complete with the CD can emit to its log all the relevant connection information required to bring up a WG tunnel with the dock.  

(Does not actually need to bring up the tunnel -- that is a follow on story).

There does need to be a way for the CD to report success or failure of the exchange.  Maybe just writing to a log file for now?  Note that the prototype has ability to write status messages over a local socket which could then be monitored by a process.  Probably that is overkill for now -- but that was super helpful when building the GUI.

mkolehmainen:
This is just a proof of concept of start-me-up.  This now will connect to a listening node and do the start-me-up exchange.  This does not yet do anything with the response.

Talks to the node, gets the response and checks a signature.

### DIFF
--- a/adapter/cd/src/cd/startmeup.rs
+++ b/adapter/cd/src/cd/startmeup.rs
@@ -1,8 +1,8 @@
 
 use std::net::Ipv6Addr;
-use std::io::{Error, ErrorKind, Cursor, SeekFrom, Read, Seek};
+use std::io::{Cursor, Error, ErrorKind, Read, Seek, SeekFrom, Write};
 use std::time::SystemTime;
-use byteorder::{BigEndian, WriteBytesExt, ReadBytesExt}; 
+use byteorder::{BigEndian, WriteBytesExt}; 
 use base64::prelude::*;
 use ring::signature;
 use tracing::info;
@@ -10,10 +10,9 @@ use tracing::info;
 
 use crate::cd::zpr::Configuration;
 
-use tokio::{
-    net::TcpStream,
-    io::AsyncWriteExt,
-};
+use tokio::net::TcpStream;
+use tokio::io::{AsyncWriteExt, AsyncReadExt};
+
 
 
 const NOISE_KEY_LEN: usize = 32;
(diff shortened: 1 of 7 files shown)


adapter/cd/src/cd/startmeup.rs
```diff
@@ -1,5 +1,216 @@
...
+    let mut stream = TcpStream::connect(format!("{}:{}", config.get_dock_host(), config.get_dock_startup_port())).await?;
+
+    stream.write_all(&msg).await?;
+    stream.shutdown().await?; // shut down write side of the connection.
+
+    // TODO: This assumes we get the entire response in a single read.
+    let mut resp_buffer = vec![0; 1024];
+    loop {
+        stream.readable().await?;
+        match stream.try_read(&mut resp_buffer) {
```
cpacejo: Use [`AyncReadExt::read_exact`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#method.read_exact), it does all this for you, including handling the case of a partial read.

adapter/cd/src/cd/startmeup.rs
```diff
@@ -1,5 +1,216 @@
...
+            ErrorKind::Other,
+            "Dock response too short",
+        ))
+    }
+
+
+    let mut rdr = Cursor::new(&resp_buffer);
+
+    rdr.seek(SeekFrom::Start(OFFSET_WG_PORT as u64))?;
+    let wg_port = ReadBytesExt::read_u16::<BigEndian>(&mut rdr)?;
```
cpacejo: `rdr.read_u16::<BigEndian>()` should work, no?

adapter/cd/src/cd/startmeup.rs
```diff
@@ -1,5 +1,216 @@
...
+    let mut rdr = Cursor::new(&resp_buffer);
+
+    rdr.seek(SeekFrom::Start(OFFSET_WG_PORT as u64))?;
+    let wg_port = ReadBytesExt::read_u16::<BigEndian>(&mut rdr)?;    
+
+    // read IPv6
+    rdr.seek(SeekFrom::Start(OFFSET_IP_ADDR as u64))?;    
+    let mut ip6 = [0u16; 8];
+    for i in 0..8 {
+        ip6[i] = ReadBytesExt::read_u16::<BigEndian>(&mut rdr)?;
```
cpacejo: [`.read_exact()`](https://doc.rust-lang.org/nightly/std/io/trait.Read.html#method.read_exact) followed by [`Ipv6Addr::from`](https://doc.rust-lang.org/nightly/core/net/struct.Ipv6Addr.html#impl-From%3C%5Bu8;+16%5D%3E-for-Ipv6Addr)?

adapter/cd/src/cd/startmeup.rs
```diff
@@ -1,5 +1,216 @@
...
+    if key_len as usize != NOISE_KEY_LEN {
+        return Err(Error::new(
+            ErrorKind::Other,
+            format!("Invalid key length {}", key_len),
+        ))
+    }
+
+    let mut nonce = [0u8; START_ME_UP_NONCE_LEN];
+    for i in 0..START_ME_UP_NONCE_LEN {
+        nonce[i] = resp_buffer[OFFSET_NONCE+i];
```
cpacejo: `.read_exact` here and below should work…

adapter/cd/src/cd/startmeup.rs
```diff
@@ -1,5 +1,216 @@
...
+
+    if kbuf.len() > std::u16::MAX as usize {
+        return Err(Error::new(
+            ErrorKind::Other,
+            "Public key too long",
+        ))
+    }
+    let key_len:u16 = kbuf.len() as u16;
+    let mut msgbuf = Cursor::new(vec![0; 12+key_len as usize]);
+    let _ = std::io::Write::write(&mut msgbuf, &[0x01, 0x00]); // transport_type, signature_type
```
cpacejo, line 181: `msgbuf.write(&[0x01, 0x00])` should work if you `use std::io::Write`; same w/ `write_u16` below

cpacejo (approved):

adapter/cd/src/cd/startmeup.rs
mkolehmainen: That is there to disambiguate the read_exact call.

adapter/cd/src/cd/startmeup.rs
mkolehmainen: yes! In fact now I see that read_u16 on the cursor defaults to big_endian so no generic arg.

adapter/cd/src/cd/startmeup.rs
mkolehmainen: ah.

adapter/cd/src/cd/startmeup.rs
mkolehmainen: I'm doing this, but don't see why we need async at all on this buffer which is already in memory.

adapter/cd/src/cd/startmeup.rs
mkolehmainen: Ehh. The complier wants to use async in this non-async routine so I'm leaving these alone.

adapter/cd/src/cd/startmeup.rs
cpacejo: I'm not following?  I mean, instead of all this logic involving `readable` and `try_read` and looping, you can make a single call.

adapter/cd/src/cd/startmeup.rs
cpacejo: I'm confused, I don't see any async here?

adapter/cd/src/cd/startmeup.rs
cpacejo: Oh!  OK yes namespace conflict between Tokio and std.

PR #7, opened 2024-05-28
buffer pool/stack implementation

cpacejo:
Simple buffer pool/stack implementation.  Future work could include caching these at the OS thread level to reduce contention, and maintaining pools of different sizes (e.g. tiny, normal, jumbo) to allow for better memory usage.


mkolehmainen:
Looks great. Maybe consider adding a unit test or two to the buffer holder thingy?

cpacejo:
Oh good call.  I haven't added any unit tests yet but that's a great next
task.

PR #8, opened 2024-05-30
Cpacejo develop ph

cpacejo:
Here I'm pulling in the "Assembly" code which structures the flow of packets in the system.  (Also removed some dead code from BufferStack which I noticed.)

The "Assembly" defines the inter-stage interface used for shuttling packets through the packet handler.  All inter-stage and shared-resource communication and access is mediated through the assembly.

The "Packet" is the primary means of referencing a packet traveling through the packet handler.

This commit doesn't actually include any of the processes which service the inter-stage queues.

### DIFF
--- a/adapter/ph/src/assembly.rs
+++ b/adapter/ph/src/assembly.rs
@@ -1,13 +1,34 @@
-// Interface to full assembly of all stages.
-
 use crate::config;
 use crate::buffer_stack::BufferStack;
 use crate::queues::*;
 
+// Interface to full assembly of all stages.
+
+// This is the "public interface" that all stages of the system use to talk
+// to each other (via queues), and to shared resources (e.g. the buffer stack).
+
+// All queues and shared resources here should be bounded, so that
+// backpressure can flow from any processing stage all the way back to the
+// kernel network ingest queues, and that service time of any packet
+// transiting the system is not permitted to grow indefinitely under
+// pressure.
+
+// The intention is that there are no hidden unbounded queues in the system
+// (such as a mutex held over a blocking operation).  If a resource is
+// highly contended resulting in a bottleneck, that should result in some
+// visible queue becoming full.
+
 pub struct Assembly<'pktbuf> {
+    // Shared resources.  These may be accessed by any part of the system.
     pub buffer_stack: BufferStack<'pktbuf, { config::PACKET_BUFFER_SIZE }>,
+
+    // Inbound (dock->adapter) agent packet path.  Keep these topologically
+    // sorted according to expected packet flow.
     pub inbound_processor: InboundProcessor<'pktbuf>,
     pub inbound_send: InboundSend<'pktbuf>,
+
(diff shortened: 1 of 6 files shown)


adapter/ph/src/queues.rs
```diff
@@ -0,0 +1,68 @@
...
+        Self{ sender }
+    }
+
+    pub async fn enqueue(&self, packet: Packet<'pktbuf>) {
+        self.sender.send(packet).await.unwrap();
+    }
+}
+
+
+pub struct InboundSend<'pktbuf> {
```
mkolehmainen, line 36: So what is meant by InboundSend and why is it the only one to take an array of Senders?

adapter/ph/src/assembly.rs
```diff
@@ -0,0 +1,13 @@
+// Interface to full assembly of all stages.
+
+use crate::config;
+use crate::buffer_stack::BufferStack;
+use crate::queues::*;
+
```
mkolehmainen, line 4: It would be helpful for me if you had a comment on this struct that shows the expected packet flow.

adapter/ph/src/main.rs
```diff
@@ -69,8 +75,39 @@ fn main() -> ExitCode {
         tun_fds.push(unsafe { BorrowedFd::borrow_raw(rfd) });
     }
 
+    let inbound_recv_batch_size = 16;
```
mkolehmainen, line 82: A comment on these magic numbers would be useful too.  Why these values and what is impact of changing them?

mkolehmainen:
I think it looks good. I have some requests for in-line clarification.  I do like your use of descriptive lifetime annotations like `'buf` which I take it are helpfully indicating what things lifetime we are talking about (If I'm understanding this, that is...)

adapter/ph/src/main.rs
cpacejo: Right now they are essentially placeholders… I need to do some measurement and tuning before they are meaningful…

adapter/ph/src/queues.rs
cpacejo: I'll add some commentary to all these… to answer this question specifically though, InboundSend is the packet-send stage of the inbound pathway.  It's the only _necessarily_ vectored queue because it's the only one associated with a vector of resources… the TUN FDs in this case.

(TUN in multiqueue mode supplies multiple FDs, one for each outbound queue… it's to be seen how much performance gain multiqueue mode gets us but I included support for it here to make sure I wasn't boxing myself in with the design.  From my point of view it's mainly a way to parallelize _receive_ from the TUN device, since the TUN devices do not support `recvmmsg`.)

(OutboundSend has only the single socket FD -- that FD being a socket, _does_ support `recvmmsg` -- therefore I doubt we need multiple inbound queues for performance there and haven't built out that mechanism.)

Note, the inbound receive stage is also associated with a vector of resources (again, the TUN FDs), but it's not associated with a queue (since all its work comes from the TUN device, not from another stage), so it doesn't make an appearance in this commit.

PR #11, opened 2024-05-31
add extension modules for vectored and async I/O

cpacejo:
These are needed for future code import.  These are all things that ideally would be provided by the standard library or Tokio but currently aren't.

### DIFF
--- a/adapter/ph/src/ext.rs
+++ b/adapter/ph/src/ext.rs
@@ -1,0 +1,3 @@
+// Various "extensions" to external crates.
+pub mod std;
+pub mod tokio;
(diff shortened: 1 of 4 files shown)


mkolehmainen:
Looks good to me. 

What is the meaning of the `_inner` designation on those functions in your tokio file?

mkolehmainen (approved):
(I just noticed this "add your review" button and am trying it out...)

Looks good to me!

cpacejo:
> What is the meaning of the `_inner` designation on those functions in your tokio file?

They are private "inner functions" which are called in a loop by the corresponding public "outer functions" via `try_io()`.  The alternative is to write them inline as a closure.

PR #12, opened 2024-06-04
import inbound_recv_worker

cpacejo:
The inbound receive worker has one task: it reserves buffers from the buffer pool, pulls batches of packets out of the tether socket into these buffers, and enqueues them into the packet processing stage.  There is necessarily one inbound receive worker, to ensure packets remain ordered when entering the system.  Fanout occurs during enqueue to the packet processing stage.

Future work: @svjaffer has suggested we may want to consider explicitly head-dropping (or otherwise managing the packet drop behavior) from the socket when the processing queue is full.  This would be a simple change to try to enqueue each packet (and drop if failed), rather than block to enqueue it.  (Or ideally the kernel lets us configure UDP socket drop behavior but I couldn't find anything in a cursory search.)

### DIFF
--- a/adapter/ph/src/inbound_recv_worker.rs
+++ b/adapter/ph/src/inbound_recv_worker.rs
@@ -1,0 +1,68 @@
+use core::future::Future;
+use std::io::IoSliceMut;
+use tokio::net::UdpSocket;
+use crate::ext::std::vec::VecExt;
+use crate::ext::tokio::net::*;
+use crate::assembly::Assembly;
+use crate::packet::Packet;
+
+#[derive(Copy, Clone)]
+pub struct Config {
+    pub batch_size: usize
+}
+
+async fn worker(
+    config: &Config, asm: &Assembly<'_>, socket: &UdpSocket
+) {
+    let mut bufs = Vec::new();
+    let mut iovs_outer = Vec::new();
+    //let mut iovs_slices_outer = Vec::new();  // TODO; see below
+    let mut msgs_outer = Vec::new();
+
+    loop {
+        // grab some buffers from the pool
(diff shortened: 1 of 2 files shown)


adapter/ph/src/inbound_recv_worker.rs
```diff
@@ -0,0 +1,68 @@
...
+        msgs_outer = msgs.recycle();
+    }
+}
+
+pub fn launch<'pktbuf, AsmRef: 'pktbuf, UdpSocketRef: 'pktbuf>(
+    config: &Config, asm: AsmRef, socket: UdpSocketRef)
+-> impl Future<Output = ()> + Send + 'pktbuf
+    where AsmRef: std::ops::Deref<Target = Assembly<'pktbuf>> + Send + Sync,
+        UdpSocketRef: std::ops::Deref<Target = UdpSocket> + Send + Sync
+{
```
mkolehmainen, line 65: INSANE function signature ! ;)

mkolehmainen: Looks good to me.  But also at my limit of rust type annotation comprehension, so I may have missed something.

cpacejo: The reason for this whole mess is because I want this function to accept _either_ a `&'static Assembly<'pktbuf>`, _or_ an `Arc<Assembly<'pktbuf>>` -- i.e., a reference I can pass into a thread which ensures the referencee outlives the thread.  Not sure if there's a shorthand for Deref+Send+Sync…

svjaffer (approved):
I can't vouch for the Rust syntax, but the structure looks good

PR #44, opened 2024-06-06
added counter to the ph

### LINKED ISSUE #43: [PH] packet drop counting
Assembly ([code](https://github.com/org-zpr/zpr-core/blob/main/adapter/ph/src/assembly.rs)) is where all the shared structures live.  You'll want to add one here, call it like Counters or something.  It will be the home for all shared counters we might eventually want.  It should have an interface which allows adding things to counters, resetting counters, and reading out counters.  (Probably each counter has a separate function to keep it extensible?)  All the counters should be 64-bit.  You'll want to use atomics to minimize contention: https://doc.rust-lang.org/std/sync/atomic/

Then in place of my TODO above you can increment this counter.

Actually while we're here let's add a counter for all received packets, regardless of whether they're dropped.

owenrhysthomas:


adapter/ph/src/config.rs
```diff
@@ -1,4 +1,4 @@
 // Static system configuration.
 
 // Size of a packet buffer.
-pub const PACKET_BUFFER_SIZE: usize = 16384;
+pub const PACKET_BUFFER_SIZE: usize = 16;
```
cpacejo: Leftover testing change! :)

adapter/ph/src/counter.rs
```diff
@@ -0,0 +1,115 @@
...
+    number: AtomicU64
+}
+
+impl Counter {
+    pub(crate) fn new() -> Self {
+        Self { number: AtomicU64::new(0) }
+    }
+
+    pub(crate) fn reset(&self) {
+        self.number.store(0, Ordering::Relaxed); // TODO ordering type
```
cpacejo, line 15: For statistical counters like this "relaxed" is the correct ordering.  All that we need is the race-freedom which atomics provide independent of ordering.

adapter/ph/src/counter.rs
```diff
@@ -0,0 +1,115 @@
...
+    pub(crate) fn increase_by(&self, amount: u64) {
+        self.number.fetch_add(amount, Ordering::Relaxed);
+    }
+
+    pub(crate) fn decrease_by(&self, amount: u64) {
+        self.number.fetch_sub(amount, Ordering::Relaxed);
+    }
+
+    pub(crate) fn get_count(&self) -> u64 {
+        self.number.fetch_add(0, Ordering::Relaxed) // TODO look for more elegant way to get value
```
cpacejo: [`load()`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU64.html#method.load) is what you want.  (Why the mutating methods use "fetch" and this is called "load" is just for historical reasons I think.)

adapter/ph/src/assembly.rs
```diff
@@ -30,5 +31,11 @@ pub struct Assembly<'pktbuf> {
     // Outbound (adapter->dock) agent packet path.  Keep these topologically
     // sorted according to expected packet flow.
     pub outbound_processor: OutboundProcessor<'pktbuf>,
-    pub outbound_send: OutboundSend<'pktbuf>
+    pub outbound_send: OutboundSend<'pktbuf>,
+
+
+    // counters[0] represents the total number of packets recieved
```
cpacejo, line 37: Would be better to use an enum to represent these values, but let's do that in a separate commit and we'll use `EnumMap` to link the array size to the enum too.

adapter/ph/src/main.rs
```diff
@@ -5,6 +5,8 @@ use tokio::io;
 use tokio::net::UdpSocket;
 use tokio::sync::mpsc;
 use tokio::task::JoinSet;
+use std::io::Error;
+use std::process;
```
cpacejo: These look unused?

adapter/ph/src/assembly.rs
```diff
@@ -30,5 +31,11 @@ pub struct Assembly<'pktbuf> {
...
+    // counters[1] represents the number of dropped packets
+    pub counters: [Counter; 2]
```
cpacejo, line 39: Hm what I was imagining in my head was that there was a single `Counters` struct, which had an `[AtomicU64; 2]` as its sole member.  BUT I don't think there's anything wrong with this, except it possibly may be unwieldy to do actions like "reset all counters" or "dump all counters".  Let's keep in mind to revisit if we bump into issues but no need to change right now.

adapter/ph/src/main.rs
```diff
@@ -112,11 +116,16 @@ fn main() -> ExitCode {
     let (os_inq, os_outq) = mpsc::channel(outbound_send_batch_size * 2);
     let outbound_send = OutboundSend::new(os_inq);
 
+    let counters = [Counter::new(), Counter::new()];
```
cpacejo, line 118: When switching to `EnumMap` in the future, you should just be able to write `enum_map! { _ => Counter::new() }`.

cpacejo (changes requested):
Looks good!  Needs two fixes; just remove the change of the packet buffer size, and also there's what look like a couple unused imports in main.  Great first PR!!

cpacejo (approved):

cpacejo:
Looks good thanks!!

PR #47, opened 2024-06-06
added stderr messages displaying packet counts when SIGUSR1 and SIGTERM signals are recieved

### LINKED ISSUE #35: [PH] packet count reporting
This is the more involved part -- how to read the packet counts out to some debug tool.  It's pretty open-ended how we'll communicate with the PH as we build out the CA -- so let's not go down that route and keep it simple instead.  We'll install a signal handler (https://docs.rs/signal-hook/latest/signal_hook/low_level/fn.register.html) when the PH starts up, say for SIGUSR1, which when triggered, causes the packet counts to be emitted on stderr.  (I wouldn't worry about resetting the counters.  I think that's just something we might want to do internally in response to e.g. a configuration change in the future.)

Note there are some subtleties with writing to stderr (and doing other things) while in a signal handler… we can discuss those at that point!

And a suggestion from Sarah – we should also install the signal handler on the termination signals – SIGTERM, SIGSEGV, SIGBUS, SIGABRT at least.  So we get packet counts printed out on process death in most cases.

owenrhysthomas:


adapter/ph/src/main.rs
```diff
@@ -41,6 +44,13 @@ fn set_fd_nonblocking<T: AsRawFd>(fd: T) -> io::Result<()> {
     Ok(())
 }
 
+fn emit_counts(counts_arr:&[Counter]) {
```
cpacejo, line 47: When we figure out how to handle signals directly let's move this into another file -- maybe to live with the counters directly.  No need to change right now though (I've got some random functions hanging out here too I know).

cpacejo (approved):

PR #50, opened 2024-06-07
added first stage of RPC handling to program

### LINKED ISSUE #49: PH control channel: accept a single connection
The packet handler needs some way for us to control & query it from another process (e.g. the client adapter, or debug tools).

"Ideally" the PH would present some sort of RPC interface (e.g. Thrift or gRPC) which the CA can use.  But I'm hesitant on this for two reasons:

(1) These tend to add a lot of code bloat (i.e. dependencies… gRPC for one requires an HTTP2 server!).  For the PH, being data path, we want to minimize extra "stuff" that's in the process.

(2) We will likely have very few RPC calls, all very simple, so we don't gain too much from them.

So instead, until one of the above changes, let's go with the "old school" Unix-domain sockets.  Unix-domain sockets are byte stream sockets like TCP, but instead of running over IP/Ethernet, they connect directly through the kernel on the local machine, and are addressed via the filesystem hierarchy.  They also have the benefit that kernel objects such as file descriptors can be sent through them (not something we need right now though).

For now, our RPC mechanism will simply be:
1) Client (the CA or a debug tool) opens a new connection to the socket; server accepts the connection.
2) Client sends a one-line text command, and then is free to shut down its half of the socket.
3) Server (the PH) does whatever is requested.
4) Server optionally sends a (possibly multiline) reply.
5) Server sends "OK" on a single line.
6) Server shuts down its half of the socket.

(For reference -- this isn't the most robust design for obvious reasons… if we need something more robust though that is future work!)

BTW – one other benefit of using Unix-domain sockets is that you can again use `nc` as your debug tool.  `nc -U /path/to/socket` connects to the socket, you can type the command, hit enter, and see the reply.

Anyway -- this initial task entails modifying the PH to:
1) Accept a new argument (let's just make it the 1st argument for now -- doing proper argument parsing is a separate task) which is the path to use for the socket.
2) Bind a listening Unix-domain socket to that path, using [tokio::net::UnixListener::bind()](https://docs.rs/tokio/latest/tokio/net/struct.UnixListener.html#method.bind).  You'll want to do this in the same spot where we connect the UDP socket.
3) Create a new worker stage, call it `rpc_worker` or such.  You can copy from `inbound_recv_worker` and delete the body of its `worker()` function.  (You'll just want to replace the `UdpSocket` argument with a `UnixListener`.  Keep the `Assembly` argument though, you'll need it when we actually implement RPC calls.)
4) Launch the worker stage from `main()` just like how the other workers are launched.
5) In a loop in your new `worker()`, wait for & accept each new connection using [UnixListener::accept()](https://docs.rs/tokio/latest/tokio/net/struct.UnixListener.html#method.accept).  You can then print out a message using `eprintln!()` that you got a connection, and for now simply immediately close it using `.shutdown()`.

You should be able to test the above by running the PH like before with the launcher script, but now with an extra argument which is the path of the socket you want to create (say, `/tmp/ph.sock`).  Then run `nc -U /tmp/ph.sock` and you should see the connection come through in the PH's window.  You should be able to do this multiple times.

(Aside – I think `bind()` requires that the path not exist – but it creates a socket file there and never deletes it.  So you might need to add a call to delete whatever's at the target path before calling `bind()`.)

owenrhysthomas:


adapter/ph/src/main.rs
```diff
@@ -124,14 +132,23 @@ fn main() -> ExitCode {
         .block_on(async {
             let socket = Box::leak(Box::new(UdpSocket::bind(self_addr).await.expect("unable to bind to self addr")));
             socket.connect(peer_addr).await.expect("unable to connect to peer addr");
+            
+            fs::remove_file(&sock_path);
+            let unix_socket =  Box::leak(Box::new(UnixListener::bind(sock_path).unwrap())); //TODO not sure if this needs the Box leak wrapper
 
-            let mut js = JoinSet::new();
+            let mut js1 = JoinSet::new();
+            let mut js2 = JoinSet::new(); //TODO not sure if these have to be different
```
cpacejo: Yes, you should re-use the same JoinSet.  I think of it as "all top-level tasks running in the application".

adapter/ph/src/rpc_worker.rs
```diff
@@ -0,0 +1,42 @@
...
+use crate::ext::std::vec::VecExt;
+use crate::ext::tokio::net::*;
+use crate::assembly::Assembly;
+use crate::packet::Packet;
+use tokio::net::UnixListener;
+use tokio::net::UnixStream;
+use tokio::io::AsyncWriteExt;
+
+#[derive(Copy, Clone)]
+pub struct Config {
```
cpacejo: No need for `Config` here – in the other workers this is just a way to pass configuration info to them but this worker isn't using any so you can just delete it and all references to it.

cpacejo (changes requested):
Looks good, only changes needed are to use only the single JoinSet, and delete the unused Config stuff.

cpacejo (approved):
Looks great, thank you!

PR #53, opened 2024-06-07
51 ph control channel respond to echo command (created branch based on branch 49 from previous task)

### LINKED ISSUE #51: PH control channel: respond to ECHO command
Now we'll want to build out the command/response infrastructure of the management path, with the first deliverable being the ability to respond to an "ECHO" command.  The steps here are:

1. First in the RPC worker file define a method `async fn echo(_asm: &Assembly<'_>) -> String`.  (Other management methods we implement will go here.)  Implement `echo()` to return just some arbitrary string; it won't do anything with the Assembly (hence the underscore, to avoid a warning).  This is the first of what will later become several management functions, so we want them all to follow the same pattern.
2. At the point where you've accepted a new connection from the Unix-domain socket, you'll have a [`UnixStream`](https://docs.rs/tokio/latest/tokio/net/struct.UnixStream.html).  We want to read a line from it – this is awkward with the default `AsyncRead` interface it supplies because that requires us to know ahead of time how many bytes we want to read, and to handle partial reads, etc.  So you'll want to wrap it in a [`BufReader`](https://docs.rs/tokio/latest/tokio/io/struct.BufReader.html#method.new), which provides the [`read_line`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncBufReadExt.html#method.read_line) method.  Use this to read the command from the stream.
3. Now we're done reading, we now want a [`BufWriter`](https://docs.rs/tokio/latest/tokio/io/struct.BufWriter.html) to send back a response (the buffering isn't as important for writing but it's nice to keep things symmetrical).
4. Now we want to dispatch based on that string, using a `match` expression.  If it's "ECHO", call `echo`, if it's something else, we can send an error back on the socket (see below) and then close it.
5. The return value of `echo().await` we'll want to send back through the socket.  This we can write back with the [`write_all()`](https://docs.rs/tokio/latest/tokio/io/trait.AsyncWriteExt.html#method.write_all) method (you'll need to use [`to_bytes()`](https://doc.rust-lang.org/std/primitive.str.html#method.as_bytes) on the string to encode it as bytes).
6. If the dispatch call succeeded (i.e. we ran `echo()`), now we can send back "OK".  If it failed (unknown command), send back "ERR".
7. Finally -- we need to `.flush()` the buffer so it makes it to the wire.  Then `.shutdown()` the stream.

You should be able to test this all again with `nc -U /path/to/socket`, this time sending the string "ECHO" and looking for the proper response back, or sending some nonsense and getting an error back.  Either way the PH should close the connection after.

owenrhysthomas:


adapter/ph/src/rpc_worker.rs
```diff
@@ -0,0 +1,70 @@
+use core::future::Future;
+use std::io::IoSliceMut;
+use tokio::net::UdpSocket;
+use crate::ext::std::vec::VecExt;
+use crate::ext::tokio::net::*;
+use crate::assembly::Assembly;
+use crate::packet::Packet;
```
cpacejo: A lot of these seem unused, IoSliceMut, Packet, VecExt, UdpSocket.  The compiler usually warns about these.

adapter/ph/src/rpc_worker.rs
```diff
@@ -0,0 +1,70 @@
...
+            Ok((mut stream, _addr)) => {
+                eprintln!("Connection recieved");
+                let mut str_message = String::new();
+                let mut split_buf = stream.split(); // split stream into read/write streams
+                let mut buf_reader = BufReader::new(split_buf.0);
+                let mut buf_writer = BufWriter::new(split_buf.1);
+                buf_reader.read_line(&mut str_message).await;
+                str_message.pop(); // Removes \n from end of string
+
+                // TODO remove \n from end of message?
```
cpacejo: Check first that there _is_ a '\n'… (i.e. that the result of `pop() == Some('\n')`).  There won't be if the client closed the connection before sending a complete line.  In this case I would simply shut down the connection without even sending an error back.  (Thinking of it like… we didn't get a complete command to process, so we shouldn't even move on to the response step.)

You can test this with `nc` by typing Ctrl+D once on an empty line to close the connection without sending anything, or Ctrl+D twice after typing some stuff to send that stuff without a newline and then close the connection.

cpacejo (changes requested):

cpacejo (approved):
Looks great thanks!!

PR #57, opened 2024-06-10
52 ph control channel respond to counters and counters reset commands

### LINKED ISSUE #52: PH control channel: respond to COUNTERS and COUNTERS RESET commands
This is a follow-on to the ECHO command.  The framework should be in place that you can easily add two new commands, let's call them "COUNTERS" and "COUNTERS RESET".  You'll need to add async functions which implement them (like for `echo()`), and add them to the dispatch loop.  Then you should be able to test them as a combination of how you tested the "ECHO" command and the counters.

(Of course this requires the counters code to be checked into main, I'm on that!!)

owenrhysthomas:


adapter/ph/src/rpc_worker.rs
```diff
@@ -0,0 +1,91 @@
...
+
+async fn echo(_asm: &Assembly<'_>) -> String {
+    return "echo\n".to_string(); // TODO change the return value of echo
+}
+
+// TODO not sure if just printing is what we want this function to do
+async fn counters(asm: &Assembly<'_>) -> String {
+    for p in 0..2 { // TODO replace 2 with some global var that represents # of packets
+        let num = asm.counters[p].get_count();
+        println!("{num}");
```
cpacejo, line 66: We'll want to have this print a description of what each counter means too… e.g.:
```
Packets received: 123
Packets dropped: 45
```
or such.  Let's hold off on that until we switch to using an enum to identify the counters.  Then you can have a centralized function which names each counter.  (I'll make a task for that today.)

adapter/ph/src/rpc_worker.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+    return "counters\n".to_string(); // TODO change the return value of counters
+}
+
+async fn counters_reset(asm: &Assembly<'_>) -> String {
+    for p in 1..2 {
+        asm.counters[p - 1].reset();
```
cpacejo: This doesn't seem right… why not `0..2` and `[p]` like in the `counters()` function?

cpacejo (changes requested):
Something seems off in `counters_reset()`, otherwise looks good!

adapter/ph/src/rpc_worker.rs
```diff
@@ -0,0 +1,91 @@
...
+    for p in 0..2 { // TODO replace 2 with some global var that represents # of packets
+        let num = asm.counters[p].get_count();
+        println!("{num}");
+    }
+    return "counters\n".to_string(); // TODO change the return value of counters
+}
+
+async fn counters_reset(asm: &Assembly<'_>) -> String {
+    for p in 0..2 {
+        asm.counters[p - 1].reset();
```
cpacejo: Still this I think should be `[p]`?

cpacejo (changes requested):
`[p - 1]` I think should be `[p]` still?

cpacejo (approved):
Looks great thank you!

PR #60, opened 2024-06-10
move packet metadata onto heap

### LINKED ISSUE #48: [PH] move packet metadata onto heap

cpacejo:
Not sure I love the interface I've created.


svjaffer (commented):
From the internet, you can do a static assertion like this

`const _: () = assert!(FOO == BAR, "assert_foo_equals_bar");`

svjaffer (approved):

PR #63, opened 2024-06-10
54 ph control channel accept concurrent connections

### LINKED ISSUE #54: PH control channel: accept concurrent connections
Right now the management RPC implementation is single-threaded.  It's unable to process two commands at the same time.  Since all our commands are currently short-running, and our clients well-behaved, this is not a problem.  This won't always be the case.  (Particularly, it can be an issue if one client gets stuck while sending its command; no other clients will be able to connect!)

To fix this -- after accepting a connection on the Unix-domain socket, we'll want to spawn a new task which runs the RPC code in an `async` block, so we can immediately get back to the business of accepting connections.

We could just do this with [`spawn(async { ... })`](https://docs.rs/tokio/latest/tokio/task/fn.spawn.html) but it would be nice to keep track of the active tasks, so we could e.g. query what's active, or limit the number of active tasks.  Though as a first step using plain `spawn()` is good to test that it does what you want (the first test below should pass).

So instead we'll want to create a [`JoinSet`](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html).  This has its own `.spawn()` method which, after spawning a task, adds it to the set to keep track of it.

(Actually you might want to use [`spawn_local()`](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html#method.spawn_local) since (a) we only need concurrency, not parallelism, and (b) that will put fewer restrictions on what data you can pass into the `async { }` block.)

The last thing you'll need to do is, on the line where you `accept()` new connections, you'll need to instead use [`select!()`](https://docs.rs/tokio/latest/tokio/macro.select.html) to _either_ accept a new connection _or_ "join" a task from the `JoinSet` (i.e. collect the state of a completed task) -- that you do with [`join_next()`](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html#method.join_next).  We don't really care about the fact that the task completed, but we need to join it, else its return code just sits in the `JoinSet` forever.  You can print a debug statement out here to make sure it's joining tasks correctly when they complete.

The easy way to test this is -- launch _two_ instances of `nc -U /path/to/socket`… in the second one, type "ECHO" and hit enter.  Without the change, it should just hang.  Now in the first one, type "ECHO" and hit enter – you should get a response from the PH right away, and then the second one should complete.  With the change, the order doesn't matter: both should respond instantly.

owenrhysthomas:


cpacejo (approved):
Looks great!  Merge it in :)

PR #65, opened 2024-06-10
[No merge] Test out the rust formatter

svjaffer:


nonrenewable (commented):
fwiw nothing bothers me about this formatter so i would probably go all in left to my own devices but i'm curious what folks who have actually written more rust than me think!

(cc @cpacejo @mkolehmainen )

cpacejo:
I like it!!!  I'm thinking… let's add `rustfmt --edition 2021 --check --config skip_children=true <modified files>` as a [Git hook](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks)?  Client pre-commit and server update hooks maybe?  To reject pushes which modify a file and leave it in an unformatted state.  I can assign myself that task if that sounds good to y'all…

PR #67, opened 2024-06-10
basic DTLS connection

### LINKED ISSUE #46: PH stage for setting up, maintaining, & transiting data across a DTLS connection
This will sit between the raw network recv/send stages, and the processing stages.  It will be a _single_ stage (not an inbound/outbound pair), which adapts the OpenSSL DTLS implementation to sit in our async framework (using the OpenSSL BIO mechanism… a.k.a. [SslStream::new()](https://docs.rs/openssl/latest/openssl/ssl/struct.SslStream.html#method.new) with our queues as the read/write stream… running in nonblocking mode and then wrapped appropriately for tokio…)  It will be slightly complex but self-contained.

Future work entails seeing if there's a way to parallelize this, as the cryptography will be CPU-intensive.  Being a datagram protocol I'd expect there's _some_ degree of parallelization we can do here.  Worst case we need to open multiple streams.

Particularly -- the session setup can occur outside the main assembly -- though I suspect there are DTLS management messages which would need to be handled within the assembly.

cpacejo:
Does not do certificates or set any security parameters particularly.

Replaces the current `inbound_recv` stage, and also the `outbound_send` stage from the experiments repo.


adapter/ph/src/dtls_worker.rs
```diff
@@ -0,0 +1,108 @@
...
+use crate::config;
+use crate::packet::{packet_body_buffer, Packet};
+use crate::udp_stream::UdpStream;
+
+// NOTE: Packet buffers *must* be at least 16384 bytes, to match TLS maximum
+// record size.  This is because OpenSSL read functions provide no way to
+// determine whether the provided read buffer was too small to contain a
+// full record.  So to ensure correct behavior we must be prepared to accept
+// the maximum size record.
+const _: () = assert!(config::PACKET_BUFFER_SIZE >= 16384, "packet buffers too small for OpenSSL DTLS");
```
cpacejo: Oh!!  This isn't correct!!  Because we now keep metadata in the packet buffer too.  I need to rewrite this assert in terms of the actual available packet body space.  Which means all our packet buffers must grow to 20480 bytes at least.  Which isn't the end of the world.

cpacejo: Fixed now.

svjaffer (approved):

PR #77, opened 2024-06-12
implement wrapper for DTLS_get_data_mtu

### LINKED ISSUE #71: write Rust wrapper for DTLS_get_data_mtu
This function isn't included in the `openssl` crate for some reason.  Make a wrapper for it.

cpacejo:
This is needed so we can reject outbound packets which would otherwise be fragmented by OpenSSL.  Resolves #71.


mkolehmainen (approved):
I do note lack of comments.  What is an SSL-DATA-MTU and why do we  care?

cpacejo:
Ah sorry, it's just a wrapper for [`https://www.openssl.org/docs/man1.1.1/man3/DTLS_get_data_mtu.html`](https://www.openssl.org/docs/man1.1.1/man3/DTLS_get_data_mtu.html)… I'll add a comment generally that that's what lives in here.  MTU = Maximum Transmission Unit, this MTU specifically is the maximum size message that the DTLS connection can send without fragmenting.

PR #80, opened 2024-06-12
wrapper for querying socket MTU

### LINKED ISSUE #79: add support to query network MTU
Doesn't exist in tokio or std.

cpacejo:
Resolves #79.

Modeled after API of [`ttl()`](https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.ttl) method.

### DIFF
--- a/adapter/ph/src/ext/tokio.rs
+++ b/adapter/ph/src/ext/tokio.rs
@@ -62,3 +62,22 @@ pub mod io {
         }
     }
 }
+
+pub mod net {
+    use std::io;
+    use nix::sys::socket;
+    use tokio::net::UdpSocket;
+
+    pub trait UdpSocketExt {
+        fn mtu(&self) -> io::Result<u32>;
+    }
+
+    impl UdpSocketExt for UdpSocket {
+        fn mtu(&self) -> io::Result<u32> {
+            match socket::getsockopt(self, socket::sockopt::IpMtu) {
+                Ok(mtu) => Ok(mtu as u32),
+                Err(errno) => Err(io::Error::from(errno))
+            }
+        }
+    }
+}


adapter/ph/src/ext/tokio.rs
```diff
@@ -62,3 +62,22 @@ pub mod io {
         }
     }
 }
+
+pub mod net {
```
nonrenewable, line 66: this is kind of cute! essentially multiple "modules" (?) in the same file where each seems to list its own dependencies for containing scope?

adapter/ph/src/ext/tokio.rs
```diff
@@ -62,3 +62,22 @@ pub mod io {
...
+    use std::io;
+    use nix::sys::socket;
+    use tokio::net::UdpSocket;
+
+    pub trait UdpSocketExt {
+        fn mtu(&self) -> io::Result<u32>;
+    }
+
+    impl UdpSocketExt for UdpSocket {
```
nonrenewable, line 75: this is the oddest syntax i've ever see but i find it relatable (unsure whether as a human or programmer 😂 )

question for usage of this UdpSocketExt behavior - does it get attached to the global UdpSocket class/struct/whatever definition so everyone importing `tokio::net::UdpSocket` now gets the functionality or does the instance need to be initialized specially to take into account this logic/feature attachment?

adapter/ph/src/ext/tokio.rs
```diff
@@ -62,3 +62,22 @@ pub mod io {
...
+        fn mtu(&self) -> io::Result<u32> {
+            match socket::getsockopt(self, socket::sockopt::IpMtu) {
```
nonrenewable, line 77: i imagine if the IpMtu reference was wrong, building would fail here (once i set up the runner, i'll know if you're lying to me about it!)

adapter/ph/src/ext/tokio.rs
```diff
@@ -62,3 +62,22 @@ pub mod io {
...
+                Ok(mtu) => Ok(mtu as u32),
```
nonrenewable, line 78: i imagine this is how error handling simply works in rust, no exceptions thrown?

nonrenewable (approved):
I enjoyed that very much, thank you @cpacejo 😁

adapter/ph/src/ext/tokio.rs
cpacejo: Yes I like that Rust allows modules to be separate files or nested within one file!  OCaml is similar.

adapter/ph/src/ext/tokio.rs
cpacejo: Neither :) You import the `UdpSocketExt` symbol to get the functionality!  i.e. `use crate::ext::tokio::net::UdpSocketExt`.  Then all these methods are available directly on any `tokio::io::UdpSocket` in that file.  Very clean and avoids "mystery methods" from appearing in unrelated files.

adapter/ph/src/ext/tokio.rs
cpacejo: Yes all undefined symbols are caught at compile time!

adapter/ph/src/ext/tokio.rs
cpacejo: Yes correct!  There's the notion of a "panic" which is somewhere between "process abort" and "exception" which I don't understand fully yet.  But explicit error returns are what's expected.

adapter/ph/src/ext/tokio.rs
nonrenewable: ooooooh!!! i like that, that's way better logically than either of what i was imagining 🎉

nonrenewable (commented):
Thank you for all these notes, super helpful 🙇

PR #81, opened 2024-06-12
Clean up most warnings

### LINKED ISSUE #58: Clean up warnings
We've got a lot of warnings let's clean them up.

Particularly let's figure out how we want to manage warnings for unused code which we _intend_ to be used at some point (e.g. part of a complete but internal API).

cpacejo:
Resolves #58

Just cleaning up warnings quick so we can get to a clean build for GitHub automation purposes.  I've ignored the warnings pertaining to the outbound packet path (which is in PR) and the uninspected error codes in `rpc_worker` (which I've opened a task for) but the rest should be cleaned up.


owenrhysthomas (approved):

owenrhysthomas:
Doesn't currently accept connections because of the DTLS launcher being at the top, I assume that will be resolved in the merge with main.

cpacejo:
Merged manually.

PR #85, opened 2024-06-12
Changed Assemnly counts member to an enum_map, addressed some warning…

### LINKED ISSUE #74: use enum to identify counters
Currently, the packet counters are identified by numeric array index (0, 1, 2, etc.)

Let's give these names!

Steps here are:

1) Declare an enum type (call it say `CounterType`) with one value for each counter.  E.g. InboundPacketsReceived, InboundPacketsDropped, InboundPacketsSent, OutboundPacketsReceived, OutboundPacketsDropped, OutboundPacketsSent…

2) We now want to use these enum values as array indices.  To do that we'll need the [`enum-map` crate](https://docs.rs/enum-map/latest/enum_map/index.html).  Add that as a dependency in `Cargo.toml`.

3) Now you can mark the enum as `#[derive(Enum)]`.  (You'll have to use `use` to make `Enum` visible of course, see the example in the `enum-map` documentation.)  This marks it as an enhanced sort of enum that works with the `enum-map` crate.

4) Now instead of creating a plain array of counters, you can use the [`enum_map!` macro](https://docs.rs/enum-map/latest/enum_map/macro.enum_map.html) to initialize it as an `EnumMap`, which is just an array which uses enum values as its indices.  (Under the hood it's still just a plain array, this is purely type magic to make things safer/more readable.)

4-alt) Alternatively – you can implement the [`Default` trait](https://doc.rust-lang.org/std/default/trait.Default.html) for the `Counter` type, to just create a zeroed counter – and then use `EnumMap::default()` (instead of using the `enum_map!` macro).  Whichever method you feel is cleaner!

5) Now you'll have to update all references to individual counters which currently use numeric indices (e.g. `[1]`) to use enum references instead (e.g. `[CounterType::InboundPacketsDropped]`).

You'll probably want to put the enum type in a new `counters.rs` file so it can be accessible from other modules.

owenrhysthomas:
…s in rpc_worker


cpacejo (approved):
Nice!!  Glad to see EnumMaps working for us here.

PR #86, opened 2024-06-12
Changed Assemnly counts member to an enum_map, addressed some warning…

### LINKED ISSUE #75: give externally-visible names to counters
Let's write a function to map counter indexes (enum values) to human-readable names, and then use this when printing out counters on the management channel.

Previously blocked on:
- org-zpr/zpr-core#74

owenrhysthomas:
…s in rpc_worker


adapter/ph/src/counters_enum.rs
```diff
@@ -0,0 +1,30 @@
...
+    OutPacksRec,
+    OutPacksDrop,
+    OutPacksSent,
+}
+
+// pub fn new_counters_array() {
+//     let enum_map: EnumMap<CounterType, Counter> = enum_map! { _ => Counter::new(), };
+// }
+
+pub fn name_counters(count_num: CounterType) -> String {
```
cpacejo, line 17: Hm… instead of a free function, let's make this an implementation of the `Display` trait (see [here](https://doc.rust-lang.org/std/fmt/trait.Display.html) for examples).  Then, instead of calling `name_counters(x)`, you can call `x.to_string()` or even just pass an enum value directly into a format string like `println!("{}: {}", x, value)`.

(Actually, you could probably even keep `name_counters()` as is, just make it private, and implement the `Display` trait just by calling out to `name_counters()`.  Might make for slightly cleaner code that way anyway.)

Feel free to do this in a separate PR if that's easier.

cpacejo (approved):
Looks good!  But I think that the enum->string function we can implement as the `Display` trait… feel free to do that in a separate PR, I'll approve this for merging.

PR #88, opened 2024-06-13
Moved async block that handles single connection into error function …

### LINKED ISSUE #82: proper error handling in rpc_worker
In `rpc_worker` there's a lot of I/O which can error out which we currently ignore (and the compiler warns about).  We could add `.unwrap()` to these, which causes the runtime to panic if any errors are encountered.  But that would bring down the data path if an error occurs, which we want to avoid.  So we'll want to do some proper error handling.

Probably the best way to do this is the following:

1) Factor out the code which handles a single connection (i.e., the `async move {}` block) into a separate async function which returns a `std::io::Result`.  So the signature will be something like `async fn handle_connection(...) -> std::io::Result<()>`.  Now instead of `async move {}` you can call this function.

2) Inside `handle_connection()`, all calls to `.unwrap()` you can replace with `?` (so like `write("foo").await?`), which will cause `handle_connection()` to return with an error return if that function call returns an error.

3) In the branch of the `select!` which joins completed tasks, we can inspect whether the task completed successfully or not.  (This is the argument to `Some(_)` -- i.e. if we make it `Some(ret)`, `ret` will be either `Ok(())` or `Error(err)` indicating whether the task completed successfully.)  If the task errored out we can print something to stderr but otherwise probably it's safe to ignore.

owenrhysthomas:
…for error handling

### DIFF
--- a/adapter/ph/src/rpc_worker.rs
+++ b/adapter/ph/src/rpc_worker.rs
@@ -9,59 +9,66 @@ use tokio::io::BufWriter;
 use tokio::task::JoinSet;
 use tokio::task::LocalSet;
 use tokio::io::AsyncBufReadExt;
+use std::io::Error;
 
 async fn worker(
     asm: &'static Assembly<'static>, socket: &UnixListener
 ) {
-    let mut set = JoinSet::new();
+    let mut set = JoinSet::<Result<(), Error>>::new();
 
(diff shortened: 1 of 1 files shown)


cpacejo (approved):
Looks great!

PR #92, opened 2024-06-13
Implemented clap to add flags for parsing command line arguments

### LINKED ISSUE #89: proper command line parsing in PH
We're about to add a bunch of parameters to the PH, and the basic command line parsing we're doing is getting messy.  Let's integrate with the [clap](https://docs.rs/clap/latest/clap/) crate to pass arguments using named command-line parameters, rather than the sequence we have right now.  We can name the arguments like this:

* `--control-path <path>` path to control socket
* `--dock-addr <addr>` address of dock
* `--self-addr <addr>` our own address; now with DTLS though we can do away with this but I'll do that in a separate task
* `--tun-fd <FD>` the TUN FD, this should accept multiple values

owenrhysthomas:


cpacejo (approved):
Oh wow!!!!!!!!  This is so much cleaner, I didn't realize how simple this would end up looking.  Perfect!!

PR #94, opened 2024-06-13
90 validate dtls peer certificate

### LINKED ISSUE #90: validate DTLS peer certificate
When we make the DTLS connection in the PH, we want to validate the certificate provided by the peer against our well-known CA certificate, and reject the connection if it's not valid.

We'll want to add an option to the PH, `--ca-file`, which can be used to point to a CA certificate file.

We'll want to configure OpenSSL to use this file using the [`SslContextBuilder::set_ca_file()`](https://docs.rs/openssl/latest/openssl/ssl/struct.SslContextBuilder.html#method.set_ca_file) method.

May also need to use `SslContextBuilder::set_verify()` to enable verification.

Verify that this all works by providing a self-signed cert vs. a cert signed by the well-known CA to `openssl s_server`.  The self-signed cert should cause the PH to reject connecting to the server; the CA-signed cert should cause it to succeed.

owenrhysthomas:


adapter/ph/src/main.rs
```diff
@@ -153,7 +148,12 @@ fn main() -> ExitCode {
     ssl_context_builder.set_options(
         ssl::SslOptions::NO_COMPRESSION |
         (ssl::SslOptions::NO_SSL_MASK & !ssl::SslOptions::NO_DTLSV1_2));
+
+    //eprintln!("{}", ca_file.unwrap());
+    ssl_context_builder.set_ca_file(ca_file.unwrap()).unwrap();
```
cpacejo: `ca_file.unwrap()` here will cause a panic if `--ca_file` isn't specified.  So we either want to make `--ca_file` mandatory or do something different here if it's not supplied.  But I think defaulting to "no verification" is dangerous.  So for now let's make it mandatory.

An enhancement we could make though, to allow to run without verification, would be to have `--ca_file` be mandatory, but force the user to specify `--disable-certificate-verification` to allow running in a mode without any verification.  (I intentionally made that argument long because it ought to be clearly visible.)

cpacejo (changes requested):
Looks good but see my comment on making `--ca_file` mandatory so we don't panic when it's not supplied.

cpacejo (approved):
Looks good thanks!

PR #95, opened 2024-06-13
implement BufMut variants of async FD read extension methods

cpacejo:
These I will use in the outbound recv worker and the DTLS inbound path, to eliminate/simplify the logic of populating a packet with this data.

### DIFF
--- a/adapter/ph/src/ext/std.rs
+++ b/adapter/ph/src/ext/std.rs
@@ -1,3 +1,9 @@
+pub mod mem {
+    pub unsafe fn slice_assume_init_mut<T>(slice: &mut [std::mem::MaybeUninit<T>]) -> &mut [T] {
+        unsafe { &mut *(slice as *mut _ as *mut [T]) }
+    }
+}
+
 pub mod vec {
     pub trait VecExt<T> {
         fn recycle<U>(self) -> Vec<U>;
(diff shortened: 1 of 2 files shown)


adapter/ph/src/ext/std.rs
```diff
@@ -1,3 +1,9 @@
+pub mod mem {
+    pub unsafe fn slice_assume_init_mut<T>(slice: &mut [std::mem::MaybeUninit<T>]) -> &mut [T] {
+        unsafe { &mut *(slice as *mut _ as *mut [T]) }
+    }
```
cpacejo, line 4: This is backported from Rust nightly stdlib [here](https://doc.rust-lang.org/nightly/src/core/mem/maybe_uninit.rs.html#988-992).  It's short enough and "does the one thing in the obvious and only simple way" that I don't believe it triggers any licensing clauses.

adapter/ph/src/ext/tokio.rs
```diff
@@ -13,6 +15,9 @@ pub mod io {
             //async fn read(&self, buf: &mut [u8]) -> io::Result<usize>;
             fn try_read(&self, buf: &mut [u8]) -> io::Result<usize>;
 
+            //async fn read_buf<B: buf::BufMut>(&self, buf: &mut B) -> io::Result<usize>;
+            fn try_read_buf<B: buf::BufMut>(&self, buf: &mut B) -> io::Result<usize>;
+
```
cpacejo, line 20: These are modeled after [these](https://docs.rs/tokio/latest/tokio/io/trait.AsyncReadExt.html#method.read_buf) tokio functions.

svjaffer (approved):

PR #98, opened 2024-06-13
Draft: Break builds if warnings are present

nonrenewable:
cc @cpacejo i think you said you wanted this but now i'm doubting myself. Anyway, looked it up this seems like the simplest implementation. Seems to have "buyer beware" warnings around compatibility long term but i am not sure their first alternative (env variable) is different; we might want to consider listing the warning codes to deny though instead, that i'd be more on board with but it'll be uglier 😅 

https://github.com/rust-unofficial/patterns/blob/main/src/anti_patterns/deny-warnings.md

### DIFF
--- a/adapter/cd/src/main.rs
+++ b/adapter/cd/src/main.rs
@@ -1,3 +1,4 @@
+#![deny(warnings)]
 use daemonize::Daemonize;
 use std::env;
 use std::fs::OpenOptions;
--- a/adapter/ph/src/main.rs
+++ b/adapter/ph/src/main.rs
@@ -1,3 +1,4 @@
+#![deny(warnings)]
 use std::fs;
 use std::net::SocketAddr;
 use std::os::fd::{AsRawFd, BorrowedFd, RawFd};


cpacejo:
I do want this to happen in our CI pipeline, but not in local builds.  (I'll often temporarily have warnings while developing, but I don't think we should be committing code with warnings.)

I'll submit a separate PR with a way I found to accomplish this.  (Since <https://github.com/rust-lang/cargo/issues/8424> is still a thing.)

cpacejo:
PR: https://github.com/org-zpr/zpr-core/pull/102

Oh, you mentioned the `RUSTFLAGS` method, a downside I've read of this is it applies to _all_ crates (i.e. external crates also), and overrides any Rust flag settings they've made.

The long-term stability argument is interesting, we might want to disable this at some point in the future, but right now, I find I'm sometimes drowning in warnings, I can't easily spot warnings related to my own changes.

nonrenewable:
Succeeded by https://github.com/org-zpr/zpr-core/pull/102

PR #99, opened 2024-06-14
91 provide dtls client ca list

### LINKED ISSUE #91: provide DTLS client CA list
The configured CA root cert should also be used as a client CA file via [`SslContextBuilder::add_client_ca`](https://docs.rs/openssl/latest/openssl/ssl/struct.SslContextBuilder.html#method.add_client_ca).  (Probably need to load it using [`X509::from_pem`](https://docs.rs/openssl/latest/openssl/x509/struct.X509.html#method.from_pem).)  Testing can be performed via Wireshark.

owenrhysthomas:


cpacejo (approved):
Perfect, thanks!  (I went ahead and did the merge to make the diff view cleaner.)

PR #102, opened 2024-06-14
add `ci` feature which builds with warnings treated as errors

cpacejo:
With this, `cargo build -F ci` errors out if warnings are present; `cargo build` does not (existing behavior).

### DIFF
--- a/adapter/cd/Cargo.toml
+++ b/adapter/cd/Cargo.toml
@@ -23,3 +23,5 @@ toml = "0.8.12"
 tracing = "0.1"
 tracing-subscriber = "0.3"
 
+[features]
+ci = []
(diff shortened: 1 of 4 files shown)


nonrenewable (approved):
It's funny i did a version like this myself and was like "this is a lot, why would i want CI to behave differently than local builds?!" and truncated it to what i have. Your comment ( https://github.com/org-zpr/zpr-core/pull/98#issuecomment-2168662745 ) though reminds me of how the world works (i just usually work around a lot of the work recently.. i blame julia 😂 )

This looks perfect and once things are ready, you can now just adjust the makefiles (downside of re-using makefiles for ci and local - if you don't want them to behave the same way like in this case... 😓 probably running things locally could always be more ad-hoc than the make files?)

Agreed with your point of "let's just see where this goes over time!"

PR #109, opened 2024-06-17
96 provide dtls certificate

### LINKED ISSUE #96: provide DTLS certificate
The client adapter PH will need to provide a client certificate as part of the DTLS handshake so it can authenticate itself to the node.  We'll want to do this using [`SslContextBuilder::set_certificate_file`](https://docs.rs/openssl/latest/openssl/ssl/struct.SslContextBuilder.html#method.set_certificate_file) and [`set_private_key_file`](https://docs.rs/openssl/latest/openssl/ssl/struct.SslContextBuilder.html#method.set_private_key_file).  Both these files should be specified as command-line arguments.  (Say, `--certificate-file` and `--private-key-file`.)

You can test with Wireshark that these show up in the handshake.  And also `openssl s_server -CAfile my-ca.crt -cert my-service.crt -key my-service.key -Verify 3 -dtls -mtu 1400 -accept 127.0.0.1:12346` should reject any connection which does not provide a client cert signed by the CA.

owenrhysthomas:


cpacejo (approved):
Perfect thanks!  (Again double-check that my merge commit looks correct…)

PR #110, opened 2024-06-17
Implemented Display trait for counter enums

### LINKED ISSUE #101: Implement Display trait for the externally visible names of the counters
Looks good! But I think that the enum->string function we can implement as the Display trait… feel free to do that in a separate PR, I'll approve this for merging.

counter_enums.rs 14-17
Hm… instead of a free function, let's make this an implementation of the Display trait (see [here](https://doc.rust-lang.org/std/fmt/trait.Display.html) for examples). Then, instead of calling name_counters(x), you can call x.to_string() or even just pass an enum value directly into a format string like println!("{}: {}", x, value).

(Actually, you could probably even keep name_counters() as is, just make it private, and implement the Display trait just by calling out to name_counters(). Might make for slightly cleaner code that way anyway.)

Feel free to do this in a separate PR if that's easier.

- org-zpr/zpr-core#75
- https://github.com/org-zpr/zpr-core/pull/86/files/9c44debbb767555a83af54e49cad130a72973454#diff-262d341774cd524911361a2133c6db57191181c63dc0431ae0e41efa19140eed

owenrhysthomas:


adapter/ph/src/main.rs
```diff
@@ -61,11 +61,9 @@ fn set_fd_nonblocking<T: AsRawFd>(fd: T) -> io::Result<()> {
 
 fn emit_counts(counts_map: &EnumMap<CounterType, Counter>) {
     for (key, &ref value) in counts_map {
-        let counter_type = name_counters(key);
-        println!("{counter_type}: {}", value.get_count());
+        println!("{}: {}", key.to_string(), value.get_count());
```
cpacejo: I think you should be able to directly use `key` here (or `"{key}"` in the format string).  Same below.

adapter/ph/src/counters_enum.rs
```diff
@@ -14,7 +15,7 @@ pub enum CounterType {
 //     let enum_map: EnumMap<CounterType, Counter> = enum_map! { _ => Counter::new(), };
 // }
 
-pub fn name_counters(count_num: CounterType) -> String {
+pub fn name_counters(count_num: &CounterType) -> String {
```
cpacejo: I think if you derive `Copy` and `Clone` on the enum above (in addition to `Debug` and `Enum`) you should be able to stick with pass by value here.

cpacejo (approved):
Looks good thanks!  A couple style suggestions but nothing incorrect.

PR #113, opened 2024-06-17
112 change the way counts is implemented in rpc worker

### LINKED ISSUE #112: Change the way COUNTS is implemented in rpc_worker
Change COUNTS from just printing the counts to sending them back on the buffer

owenrhysthomas:


adapter/ph/src/rpc_worker.rs
```diff
@@ -85,11 +84,11 @@ async fn echo(_asm: &Assembly<'_>) -> String {
 
 // TODO not sure if just printing is what we want this function to do
 async fn counters(asm: &Assembly<'_>) -> String {
+    let mut counts: String = "".to_string();
     for (key, &ref value) in &asm.counters {
-        let counter_type = name_counters(key);
-        println!("{counter_type}: {}", value.get_count());
+        counts += &(key.to_string() + ": " + &value.get_count().to_string() + "\n");
```
cpacejo: This will entail a few allocations and copies (looks like 2 temporary `String`s per counter).  Not a critical issue here as this isn't a performance-sensitive path, but in general you can avoid the allocations and copies by `write!`ing directly to `counts`, e.g. `write!(&mut counts, "{key}: {}\n", value.get_count());`.  (This is possible because `String` implements the [`Write` trait](https://doc.rust-lang.org/std/string/struct.String.html#impl-Write-for-String).)

cpacejo (approved):
The code is correct but I feel the string concatenation can be written more cleanly, while avoiding a bunch of small short-lived heap allocations when we've got a lot of counters.

PR #114, opened 2024-06-18
add Packet accessors for simultaneous access of metadata and body

cpacejo:
This fixes our problem of not being able to refer to the metadata and body of a packet simultaneously.

Better still would be a way to avoid using `zerocopy` entirely for the metadata (it _seems like_ `MaybeUninit` ought to let us do this but all its byte array methods seem to go the "wrong way" around) but this should address the most annoying issue.


svjaffer (approved):
Hopefully this cleans up my code a bunch

PR #115, opened 2024-06-18
Created ph-debug tool

### LINKED ISSUE #100: create debug tool for PH
So this is a little more open-ended, in that you are free to design the steps and architecture of this how you want.  Also whether you want to use tokio or not, but I would suggest _not_ using tokio for this as it's probably just unnecessary complexity.

Now that we have a management communication channel, let's make a standalone debug tool which communicates over that management channel.  You use `cargo new` to create a new Rust project next to `ph`, let's call it `ph-debug`.  This should make a new directory which you can run `cargo build` in to build the `ph-debug` tool (which will end up in target/debug/ph-debug).

Goals of this tool
* We'll want to be able to do various debug tasks from it.  So likely, you'll want to just take the first command-line argument as a "subcommand" to run (same as how `git` and `cargo` have subcommands).
* You'll need to accept an argument which says which socket to connect to.
* Each subcommand should use the socket to send the relevant command to the PH, read back the result, and print it out.
* Initially, there's probably a 1-to-1 mapping of each "subcommand" we define in this tool, and each socket command we've defined in the PH.  But this might not be the case going forward.  (Imagine e.g. a debug command that "watches" the counters, printing out the change every ten seconds.  So that would be sending multiple commands on the socket, and parsing and reformatting the results.)  So just keep in mind we can't rely on a direct mapping between them.

For now, the only subcommands will be "echo" and "counters", and those can probably just directly print the results.  I think the very next subcommand can be the "watch" command I described above, but let's do that as a separate PR.

Of course feel free to make multiple PRs for this tool itself too if you'd like feedback at various stages.

Again I would personally not use tokio here as it's unnecessary complexity in this context.  So you can use e.g. [`UdpSocket`](https://doc.rust-lang.org/std/net/struct.UdpSocket.html) from `std` instead of `tokio`.

owenrhysthomas:


owenrhysthomas:
I ended up using UnixStream instead of UdpSocket, which was suggested in the description, because UdpSocket::bind expects a SockAddr, but we have been using a path to connect to the PH in this way, so I used the UnixStream instead. Unlike other modules, I used the standard UnixStream not the tokio UnixStream since you said the tokio added unnecessary complexity.

adapter/ph-debug/src/main.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    Ok(())
+}
+
+// Determines which command to execute
+// Opens and closes UnixStream
+fn handle_commands(command: String, port: String) -> std::io::Result<()> {
+    println!("{command}, {port}");
+    // fs::remove_file(&port).or_else(|e| if e.kind() == ErrorKind::NotFound { Ok(()) } else { Err(e) }).unwrap();
+    let stream =  Box::leak(Box::new(UnixStream::connect(port).unwrap())); //TODO not sure if this needs the Box leak wrapper
```
cpacejo: No no need for the Box here!  Since you're not trying to pass it to a separate thread at all.

cpacejo (approved):
Looks good but no need for the Box!

PR #117, opened 2024-06-18
Created ph-debug tool

### LINKED ISSUE #116: Implemented watch command in the PH debug tool
So essentially for the watch subcommand – you'll want to accept one argument which is the interval between reports.  (It should be optional; 2 seconds is a good default.)  Then what you'll want to do is not simply dump out the packet counts – but you'll want the debug command to print the differences in packet counts from one report to the next.

owenrhysthomas:


adapter/ph-debug/src/main.rs
```diff
@@ -0,0 +1,111 @@
...
+#[command(version, about, long_about = None)]
+struct Args {
+    #[arg(short, long)]
+    command: String,
+
+    #[arg(short, long)]
+    port: String,
+
+    #[arg(short, long, default_value_t = 2)]
+    time: u64,
```
cpacejo, line 26: Hm ideally we only parse this after we determine it's the `WATCH` command.  I wonder if we can check the subcommand first before performing command parsing.

cpacejo (approved):
Looks good!  Feel free to change the argument parsing order in a separate commit.

PR #127, opened 2024-06-20
Walk packet headers to pull IPs, ports, and protocol

svjaffer:


adapter/ph/src/config.rs
```diff
@@ -1,4 +1,5 @@
 // Static system configuration.
 
-// Size of a packet buffer.  Note that this includes packet metadata.
+// Size of a packet buffer.
 pub const PACKET_BUFFER_SIZE: usize = 20480;
+pub const IPV6_ADDRESS_SIZE: usize = 16;
```
cpacejo: Maybe a new file, constants.rs? I.e. keep PH configurables separate from externally-determkned constants?

adapter/ph/src/inbound_processor_worker.rs
```diff
@@ -28,6 +29,8 @@ async fn worker<'pktbuf>(
                     // strip packet header
                     pkt.advance(std::mem::size_of::<ZdpHeader>());
 
+                    classify(&mut pkt);
```
cpacejo: Client PH doesn't need to do this, right? I added a mode flag to main, you should be able to thread it into here to skip classification on the client.

adapter/ph/src/packet.rs
```diff
@@ -1,29 +1,129 @@
...
+
+#[derive(AsBytes, FromZeroes, FromBytes, KnownLayout, Copy, Clone, Hash, Debug, PartialEq)]
+#[repr(C)]
+pub struct IpAddress {
+    pub v6: [u8; config::IPV6_ADDRESS_SIZE],
+}
+
+impl IpAddress {
+    pub fn set_from_v4(&mut self, v4_address: [u8; 4]) {
+        self.v6[0..4].copy_from_slice(&v4_address)
```
cpacejo: Should we use IPv6-mapped addresses here? i.e. ::ffff:a.b.c.d? Then we can just hash this full field blindly without worry about what portion is initialized.

adapter/ph/src/packet.rs
```diff
@@ -1,29 +1,129 @@
...
 #[derive(AsBytes, FromZeroes, FromBytes)]
 #[repr(packed)]
 pub struct PacketMetadata {
-    offset: usize,  // packet offset (must be >= PACKET_BODY_BUFFER_MIN_OFFSET)
-    len: usize,  // packet length
-    pub flow_id: u32  // flow ID for load-balancing purposes; not otherwise meaningful
+    offset: usize,    // packet offset (must be >= PACKET_BODY_BUFFER_MIN_OFFSET)
+    len: usize,       // packet length
+    pub flow_id: u32, // flow ID for load-balancing purposes; not otherwise meaningful
+    src_address: IpAddress,
```
cpacejo, line 49: Any reason not to make these public?

adapter/ph/src/packet.rs
```diff
@@ -1,29 +1,129 @@
...
 // This contains all state of a packet which is moving through the system.
 // TODO: possible we want to keep this stuff on the heap
 
 pub struct Packet<'buf> {
-    buf: &'buf mut [u8; config::PACKET_BUFFER_SIZE]
+    buf: &'buf mut [u8; config::PACKET_BUFFER_SIZE],
+}
+
+#[derive(AsBytes, FromZeroes, FromBytes, KnownLayout, Copy, Clone, Hash, Debug, PartialEq)]
+#[repr(C)]
```
cpacejo: repr(transparent) would allow this to be cast directly to/from an array.

adapter/ph/src/packet.rs
```diff
@@ -1,29 +1,129 @@
...
+        self.protocol = proto
+    }
+
+    pub fn set_src_address_v4(&mut self, src_addr: [u8; 4]) {
+        assert!(self.l3_type != L3Type::Ipv6);
+        self.src_address.set_from_v4(src_addr);
+        self.l3_type = L3Type::Ipv4;
+    }
+
+    pub fn set_src_address_v6(&mut self, src_addr: IpAddress) {
```
cpacejo: It seems asymmetric that for v4 we want an array but v6 we want a struct.

adapter/ph/src/packet.rs
```diff
@@ -1,29 +1,129 @@
...
+    pub fn set_dst_port(&mut self, dport: [u8; 2]) {
+        self.dst_port = NetworkEndian::read_u16(&dport)
+    }
+
+    pub fn set_protocol(&mut self, proto: u8) {
+        self.protocol = proto
+    }
+
+    pub fn set_src_address_v4(&mut self, src_addr: [u8; 4]) {
+        assert!(self.l3_type != L3Type::Ipv6);
```
cpacejo: Hmmm I wonder if there's a different way to express this that both disallows "bad" states at compile time rather than runtime?

adapter/ph/src/packet.rs
```diff
@@ -100,7 +204,7 @@ impl<'buf> Packet<'buf> {
         };
         let offset = md.offset - size_of::<PacketMetadata>();
         let len = md.len;
-        (md, &mut bd[offset..offset+len])
+        (md, &mut bd[offset..offset + len])
```
cpacejo, line 186: Meta nit: it's hard to tell what's just a formatting change and what is substantial. Maybe it makes sense to separate functional PRs from formatting PRs?

cpacejo (commented):
Haven't had a chance to look at the classifier yet but left comments on the rest :)

adapter/ph/src/classifier.rs
```diff
@@ -0,0 +1,676 @@
...
+    UnclassifiedL4,
+    LengthError,
+    FirstFragment,
+    SubsequentFragment,
+}
+
+const IP_VERSION_MASK: u8 = 0xF0;
+const IPV4_HEADER_LENGTH_MASK: u8 = 0x0F;
+
+#[derive(FromZeroes, FromBytes, KnownLayout)]
```
cpacejo: Does this also need `Unaligned`?

adapter/ph/src/classifier.rs
```diff
@@ -0,0 +1,676 @@
...
+    pub tc_upper_and_fl_lower: u8,
+    pub fl_upper: u16,
+    pub payload_length: u16,
+    pub next_header: u8,
+    pub hop_limit: u8,
+    pub src_address: packet::IpAddress,
+    pub dst_address: packet::IpAddress,
+}
+
+const NO_NEXT_HEADER: u8 = 59;
```
cpacejo, line 47: It may be possible to use `open_enum` for this, I think I have an example in zdp.rs.

adapter/ph/src/classifier.rs
```diff
@@ -0,0 +1,676 @@
+use crate::config;
+use crate::packet;
+use std::mem::size_of;
+use zerocopy::{ByteOrder, FromBytes, FromZeroes, NetworkEndian};
+use zerocopy_derive::{FromBytes, FromZeroes, KnownLayout};
+
+#[derive(Debug, PartialEq)]
+pub enum ClassifierResult {
```
cpacejo, line 8: Would it make sense to use Rust's `Result` type here? Then we could leverage all the goodies that provides.

adapter/ph/src/classifier.rs
```diff
@@ -0,0 +1,676 @@
...
+    }
+}
+
+fn classify_ipv4(
+    metadata: &mut packet::PacketMetadata,
+    body: &[u8],
+    mut offset: usize,
+) -> ClassifierResult {
+    // Check that there's enough room in the packet data for the base header (no options)
+    if usize::from(size_of::<IPv4Header>() + offset) > metadata.get_length() {
```
cpacejo: I think `as` keyword might work here instead of using the `from` method.

adapter/ph/src/classifier.rs
```diff
@@ -0,0 +1,676 @@
...
+    match ip_version {
+        4 => classify_ipv4(metadata, body, offset),
+        6 => classify_ipv6(metadata, body, offset),
+        _ => return ClassifierResult::NonIP,
+    }
+}
+
+fn classify_ipv4(
+    metadata: &mut packet::PacketMetadata,
+    body: &[u8],
```
cpacejo, line 99: Instead of slice + offset, would it make sense to pass just a slice which we keep re-slicing (via `[offset..]` syntax)?

adapter/ph/src/classifier.rs
```diff
@@ -0,0 +1,676 @@
...
+}
+
+fn classify_next_header(
+    metadata: &mut packet::PacketMetadata,
+    body: &[u8],
+    offset: usize,
+    protocol: u8,
+) -> ClassifierResult {
+    metadata.set_protocol(protocol);
+    match protocol {
```
cpacejo, line 168: It might be possible to use an `open_enum` here to represent all these constants.

adapter/ph/src/classifier.rs
```diff
@@ -0,0 +1,676 @@
...
+    }
+}
+
+fn is_option_length_error(
+    metadata: &packet::PacketMetadata,
+    offset: usize,
+    next_header: u8,
+    option_length: usize,
+) -> bool {
+    return option_length + offset > metadata.get_length()
```
cpacejo: Rust style nit: the value of the final statement is implicitly returned, if the statement does not end with a semicolon. Same is true of block expressions, like the match above. So you can elide all the returns that do not alter control flow.

adapter/ph/src/classifier.rs
```diff
@@ -0,0 +1,676 @@
...
+    let udp_header = UDPHeader::ref_from(header_bytes).unwrap();
+
+    metadata.set_src_port(udp_header.src_port);
+    metadata.set_dst_port(udp_header.dst_port);
+
+    ClassifierResult::OK
+}
+
+#[cfg(test)]
+mod tests {
```
cpacejo, line 285: Nice!!

cpacejo (approved):
Nice work!! I think the only major things I'd change in the classifier is to use the `Result` type to indicate success or failure, and to use plain slices rather than slice+offset if possible.

adapter/ph/src/classifier.rs
svjaffer: I was trying to do that, but casting to and from the enum ended up being really ugly.  Maybe when you're back you can show me a better way to go about it.

adapter/ph/src/classifier.rs
svjaffer: That was ending up super ugly (I spent a few hours on Friday trying to make that work properly and it got hairy).  Maybe once you're back we can talk about how to make that work better.

adapter/ph/src/packet.rs
svjaffer: Heh, OK.  I ran it to format my own code and it just happened to catch what was already there.  This would be an argument for doing the one format PR and then making sure the hooks are in, because otherwise we'll keep running into this case when we run the formatter on our own changes.

adapter/ph/src/packet.rs
svjaffer: I've changed this out so that it works better and avoids the assert.

adapter/ph/src/config.rs
svjaffer: I just moved it to packet.rs with the struct it accompanies.  We've already got a bunch of other constants there

adapter/ph/src/packet.rs
svjaffer: My best practices instinct is to make most things private and have accessors, which the Rust documentation I read seemed to encourage as well.  But we can talk about that as a principle another time.

adapter/ph/src/packet.rs
nonrenewable: (i concur with Chris in general but after reviewing much denser PRs recently it really doesn't bother me here as much as i'd thought 😂 )

adapter/ph/src/classifier.rs
```diff
@@ -0,0 +1,883 @@
+use crate::config;
```
nonrenewable, line 1: (fun fact, github collapsed this file at first and i entirely missed it, wonder where the classification logic lived :lolsob:)

adapter/ph/src/classifier.rs
```diff
@@ -0,0 +1,883 @@
...
+
+    // Begin IPv4 tests
+
+    #[test]
+    fn test_v4_tcp_success() {
+        let mut buf: [u8; config::PACKET_BUFFER_SIZE] = [0; config::PACKET_BUFFER_SIZE];
+        let mut packet = packet::Packet::new(&mut buf, packet::PACKET_BUFFER_MIN_BODY_OFFSET + 64);
+        #[rustfmt::skip]
+        let packet_data =
+            [
```
nonrenewable, line 314: this is a very VERY petty request and it's only because i don't work in this code that i might be asking it - would be nice to leave comments (if the linter doesn't get upset about it) alongside the lines representing the different subsections of the data. I can pattern match it, i think, with the results but if i were changing code and needed to change these an extra vote of confidence would be nice. that being said as i typed this i realized it might just be me and that's ok 😁

nonrenewable (commented):
I think this is awesome and i love all the tests!! Chris covered way more than i would have since i thought everything followed nicely and was clean enough but i'm glad you'll both find ways to make this even better.

I realized as i read through some of the classification logic that I really haven't read the doc (i imagine rfc doc?) describing this - did you work off of one? if so, would you mind sending it to me? if not, i'll bug you next week with my slight shortcomings of understanding!

Issue #120, opened 2024-06-19
rework dtls-worker to wait on UDP, not DTLS


cpacejo:
Because of the way UDP sockets work, we need to demultiplex tether traffic ourselves (based on source, i.e. client adapter, network address).  But this must happen _before_ the packets hit the DTLS layer, since each DTLS session has its own state.  Likely this entails dropping `tokio_openssl` as a dependency and using the non-blocking SSL interface.

In order to avoid an extra copy (since OpenSSL provides its own buffer), we'll want to use [`peek_sender()`](https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.peek_sender) to determine the source before actually dequeuing each packet.

Note to self: this will require a mini state machine per SSL connection to track whether we're in the middle of a write or not.

PR #128, opened 2024-06-20
implement GatedUdpStream

cpacejo:
This is needed for org-zpr/zpr-core#120… this allows us both to restrict OpenSSL to ingest a single UDP packet from the socket, and also to set the destination of outbound UDP packets.

### DIFF
--- a/adapter/ph/src/udp_stream.rs
+++ b/adapter/ph/src/udp_stream.rs
@@ -1,3 +1,5 @@
+use std::io::{Error, Read, Write};
+use std::net::SocketAddr;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 use tokio::io::{AsyncRead, AsyncWrite};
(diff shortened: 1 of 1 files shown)


adapter/ph/src/udp_stream.rs
```diff
@@ -43,3 +45,86 @@ impl<'a> AsyncWrite for UdpStream<'a> {
...
+            a.port() == b.port() &&
+            a.scope_id() == b.scope_id(),
+
+        _ => false
+    }
+}
+
+impl Read for GatedUdpStream<'_> {
+    fn read(self: &mut Self, buf: &mut [u8]) -> std::io::Result<usize> {
+        if self.read_gate > 0 {
```
nonrenewable, line 102: only question i can come up with here is the access to the read/threadedness of it - if it's single threaded, this is fine - is it fine if there are multiple readers that race into a read?  (as i look closer at the code details, i realize there might be syntax in the definition of all this that makes things safe and i just don't quite understand it!)

adapter/ph/src/udp_stream.rs
```diff
@@ -43,3 +45,86 @@ impl<'a> AsyncWrite for UdpStream<'a> {
     // NOTE: no point to implement the _vectored methods;
     // SSL BIO provides no means to make use of them
 }
+
+// A `GatedUdpStream` provides a means to communicate with a single remote address over
+// an unbound/unconnected `UdpSocket`.  Writes are sent to the configured remote address;
+// and reads can be gated to only succeed when it's known that an appropriate packet
+// is at the head of the queue.  (If there isn't one, it is dropped instead of being returned.)
+// When the gate is closed, a `WouldBlock` error is returned.
```
nonrenewable, line 53: questions about this description after reading some of the below (this may be well understood/implied via the naming and i'm just out of context and missing it)
- gated - i imagine the gate will only be opened if the connection is "established" (i know it's udp vs tcp so it's not exactly reliably established but packets should be expected to flow)? or is that a misconception and there's something else that word stands for?
- unexpected source address error - there's a comment before that error below saying "shouldn't have been gated" but i agree that the packet is dropped as it says here - is this due to a mismatch in packet streams processing somehow? or something else that i'm missing?

nonrenewable (approved):
I think this is cool even if i'm not 100% sure how it slots in in the bigger picture. Asked some questions, per usual but really not sure they're show stopping!

adapter/ph/src/udp_stream.rs
cpacejo: The "gatedness" is more granular than that – specifically, the gate is only opened when the next packet on the UDP socket is actually intended for the "connection" that this stream represents.  That is, it's on a packet-by-packet basis.

The "unexpected source address error" would occur if code using this opens the gate when the next packet on the UDP socket is _not_ intended for this "connection".  We can't undo reading it, and we can't do anything else with it, so… on the floor it goes!

nonrenewable: Perfect, that all makes sense! i was missing that this is on a per logical connection basis, but def got the "we can't put this genie back in the box and we messed up reading it here" so it smells like a bug elsewhere 😁 

Thank you!

adapter/ph/src/udp_stream.rs
cpacejo: Good question!  The `&mut` in the type of `self` guarantees that this block of code has exclusive access to `self`, meaning it always executes single-threaded context.

Exception – the underlying `UdpSocket` could be accessed concurrently, as it's behind a shared reference `&`.  This is necessary because other code will need to write to the socket and peek at it.  So ensuring no races there is up to the user.  But the methods of `UdpSocket` are thread-safe, so at most racy usage will result in a logic error, which would lead to an unintended packet being read, which will be caught by the source address check below.

nonrenewable: Awesomeness! that tracks and I'll try to remember all this as more things pop up to not repeat 🤞

PR #130, opened 2024-06-20
Added enums, changed existing enqueue to enqueue_packets, implemented…

### LINKED ISSUE #124: wrap queue messages in enums
All the messages we send on queues right now are plain `Packet` structs.  We want to be able to send different types of messages on queues.  So for each queue type, let's wrap this in an `enum`.

We'll want to define the `enum`s in `queues.rs`, next to each of the queues.  They can each be named like `OutboundSendMessage` (e.g. for the `OutboundSend` queue).  For now they'll just have a single variant, `Packet(Packet<'pktbuf>)`.

Then, for all the `enqueue()` methods, let's rename them `enqueue_packet()` and wrap the provided `packet` in the `Packet(...)` enum constructor.

(The alternative here is for `enqueue()` to accept a `Message` directly, but some messages, such as `TestPacket` which we'll be adding, will have slightly more involved semantics than just "put on the queue", so I think giving each message its own corresponding method is the right thing to do.)

In turn, all the workers which read from the queues will need to destructure each received message as `match msg { Packet(pkt) => { ... } }`, where `...` is the existing packet-handling code.

The end result should be no change in functionality.

owenrhysthomas:
… InboundSendMessage

### DIFF
--- a/adapter/ph/src/inbound_processor_worker.rs
+++ b/adapter/ph/src/inbound_processor_worker.rs
@@ -29,7 +29,7 @@ async fn worker<'pktbuf>(
                     pkt.advance(std::mem::size_of::<ZdpHeader>());
 
                     // send out decapsulated packet
-                    asm.inbound_send.enqueue(pkt).await;
+                    asm.inbound_send.enqueue_packet(pkt).await;
                 },
 
                 packet_type =>
--- a/adapter/ph/src/inbound_processor_worker.rs
+++ b/adapter/ph/src/inbound_processor_worker.rs
@@ -3,7 +3,7 @@ use bytes::Buf;
 use tokio::sync::mpsc;
 use zerocopy::FromBytes;
 use crate::assembly::Assembly;
-// use crate::packet::Packet;
+use crate::packet::Packet;
 use crate::zdp::*;
 use crate::InboundProcessorMessage;
 #[derive(Copy, Clone)]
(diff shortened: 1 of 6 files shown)


adapter/ph/src/inbound_processor_worker.rs
```diff
@@ -3,45 +3,50 @@ use bytes::Buf;
...
-                    pkt.advance(std::mem::size_of::<ZdpHeader>());
-
-                    // send out decapsulated packet
-                    asm.inbound_send.enqueue(pkt).await;
-                },
-
-                packet_type =>
-                    panic!("unhandled inbound packet type {}", packet_type.0)
+        for pkt in pkts.drain(..) {
+            match pkt {
```
svjaffer, line 21: This is a lot to put in a match block.  Can we move this into a function?

adapter/ph/src/inbound_processor_worker.rs
```diff
@@ -3,45 +3,50 @@ use bytes::Buf;
 use tokio::sync::mpsc;
 use zerocopy::FromBytes;
 use crate::assembly::Assembly;
-use crate::packet::Packet;
+// use crate::packet::Packet;
```
svjaffer: Did you end up needing to put this crate back in?  If not, go ahead and remove it.  It's easy to add back if we need it later and likely to remain here as a comment forever otherwise :)

svjaffer (approved):

svjaffer (approved):

PR #133, opened 2024-06-21
126 blocked handle testpacket messages

### LINKED ISSUE #126: handle TestPacket messages
Now that each worker can handle multiple message types, and we have defined our `TestPacket` message, let's add support for it to each worker.

We'll want to add a `TestPacket` variant to each queue's message type (you can just do one at first to make sure everything works).  Also add an async `enqueue_test_packet()` method which returns `Result<TestPacketMetrics, ???>` (where `???` is some appropriate error type).  For the queues which are actually arrays of queues, this method will need to take a single `usize` argument which is the queue number to send on.

(For the queues which are arrays of queues, you'll probably also want to add a method `fanout()` which simply returns the size of the queue array, so users know what values make sense to pass as the queue number.)

The associated worker then will have to handle the `TestPacket` variant.  It should do so simply by calling the `acknowledge()` method with the current queue depth of its receive queue as reported by [`Receiver::len`](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.len).

Blocked on:
- org-zpr/zpr-core#124
- org-zpr/zpr-core#125

owenrhysthomas:


adapter/ph/src/test_packet.rs
```diff
@@ -1,6 +1,7 @@
 use std::time::{Duration, Instant};
 use tokio::sync::oneshot::{Sender, Receiver, channel};
 
+// #[derive(Copy)]
```
svjaffer: Remove :)

svjaffer (approved):

PR #145, opened 2024-06-28
add limited "request visa" function to the test cli for visa service

mkolehmainen:
Not part of a story :)

But this adds a limited "request visa" function to the toy cli for playing with the visa service.


mkolehmainen:
Not sure why there are all those old commits on here.  The "files changed" tab is accurate.

PR #149, opened 2024-07-01
Added checks for format and warnings in runners

### LINKED ISSUE #59: add precommit hook(?) to disallow code with warnings
... also follow up with changing the runner to error if warnings are found after the clean up!

owenrhysthomas:


owenrhysthomas:
The only thing I really changed was in rust-adapter.yml. I added in two #[allow(dead_code)] in packet.rs and two #[allow(unused_imports)] in classifier.rs. All other changes are a result of cargo fmt.

.github/workflows/rust-adapter.yml
```diff
@@ -23,3 +23,24 @@ jobs:
     - uses: actions/checkout@v4
     - name: Build + test ph
       run: cd adapter/ph && make build && make test
+
+  format:
```
nonrenewable: very petty note - i'd name this ph-format (same prefix for the below and maybe also add the `cd` versions of these if we feel it's time? (it's a bit cumbersome for adding new subparts of the projects but looks really nice at the top level when a PR is open since every single one of these's names, i.e. ph-format, is surfaced with pass or fail) Also maybe rename the above to ph-build + cd-build? 🤷 

None of this is show stopping and it's probably a matter of taste. cc @mkolehmainen @cpacejo @svjaffer since mine is one opinion but it's soft and if folks prefer differently, might be worth taking things in another direction!

cpacejo: This sounds like a good idea to me!

cpacejo (approved):
Fantastic thank you for doing this!

.github/workflows/rust-adapter.yml
owenrhysthomas: Just pushed these changes in both this branch and PR 150 (all changes except ph-build and cd-build)

nonrenewable: Thank you!! 🙇 (i'm not writing code in here but if you ping the others, i think merging this in with formatting is something i would be into and eat the cost of but checking with the team they feel the same is the right move before merging either this or #150 depending on which one is chosen!!)

nonrenewable (approved):

cpacejo (approved):

PR #150, opened 2024-07-01
Added new tasks to runner for format and warnings

owenrhysthomas:


nonrenewable (commented):
fwiw i like the overhauling PR you have in #149 but it would be good to let everyone know you're doing it so there's no surprises and people can pick up the changes/rebase with least amount of effort!

PR #152, opened 2024-07-02
Redo the admin api in HTTP style (was gRPC)

### LINKED ISSUE #138: HTTP based ADMIN API for Visa Service (install policy only)
The visa service, which is a port of the zpr-prorotype visa service, hosts a gRPC admin service which is used by network administrators to install a new policy.

This should be converted to an HTTP API.

For first version (this story) the only function we need to support is install-policy.

mkolehmainen:
This is a start on ad admin API to the visa service.  Really only supports getting the current policy and installing a new one.  Though actually installing a policy out to a node is not yet supported inside the visa service.


visaservice/README.md
```diff
@@ -13,6 +13,89 @@ After a successful build the `vservice` binary will be found in
 
 
 
+## Visa Service Admin API
+
+This is an HTTPS API for controlling the visa service designed for network administratorsq. Access is protected by policy.
```
owenrhysthomas: administratorsq, 'q' intentional?

visaservice/README.md
```diff
@@ -13,6 +13,89 @@ After a successful build the `vservice` binary will be found in
...
+    "config_id": "",
+    "version": "",
+    "format": "base64;zip;41",
+    "container": ".... (base 64, compressed, serialzed polio.PolicyBundle) ...",
+}
+```
+
+Note that the `41` in the `format` field should be the current serialization ID for the policy schema.  
+In th code this is `SerialVersion` which can be found in `mods/polio/const.go`.
+
```
owenrhysthomas, line 75: th[e]?

visaservice/README.md
```diff
@@ -23,3 +106,16 @@ go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
 go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
 ```
 
+
+## Thrift
+
+The compiled thrift files are included int he source, but if you need to
```
owenrhysthomas: in the

visaservice/README.md
```diff
@@ -23,3 +106,16 @@ go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
 go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
 ```
```
owenrhysthomas, line 107: is this still accurate? now that the grpc has been removed?

visaservice/core/pkg/vservice/admin.go
```diff
@@ -31,166 +43,180 @@ type VSApi interface {
...
 	service struct {
 		localAddr netip.Addr // local service address
-		gsrvWg    sync.WaitGroup
-		gsrv      *grpc.Server
+		srvWg     sync.WaitGroup
+		srv       *http.Server
 	}
 }
 
 // NewAdminService creates the gRPC admin service -- you must call blocking function StartGrpc to start the service.
```
owenrhysthomas: update comment

visaservice/core/pkg/vservice/admin.go
```diff
@@ -31,166 +43,180 @@ type VSApi interface {
...
+
+	err := json.NewDecoder(r.Body).Decode(&bundle)
+	if err != nil {
+		svc.log.WithError(err).Error("admin service: failed to unmarshal policy bundle, policy install aborted")
+		http.Error(w, "policy unmarshal failed", http.StatusBadRequest)
+		return
+	}
 
 	// TODO: Is our grpc server multi-threaded?  Not sure so we are locking here
 	//       to allow only one install to run at a time.
```
owenrhysthomas: this is likely no longer relevant

visaservice/README.md
```diff
@@ -13,6 +13,89 @@ After a successful build the `vservice` binary will be found in
 
 
 
+## Visa Service Admin API
```
nonrenewable, line 16: I have several naive questions here 
- this is the API that's going to be split out from Visa Service at some point? 
- It's expected to be use-accessed directly as opposed to the Visa service itself so that's why it's JSON-based?
- it's only in here because of simplicity of migration/having things that exist that existed before?

visaservice/README.md
nonrenewable: i don't know if i love including the compiled files in the checked in version - i guess if they're very clearly separated/denoted and don't confuse with written code it's not the worst 🤷

visaservice/core/pkg/vservice/admin.go
```diff
@@ -31,166 +43,180 @@ type VSApi interface {
...
+	svc.log.Debug("admin server: fetch request processed successfully", "config", configID, "version", pcy.Version())
+	bundle := &PolicyBundle{
+		ConfigID:  configID,
+		Version:   pcy.Version(),
+		Format:    fmt.Sprintf("base64;zip;%d", pcy.GetSerialVersion()),
+		Container: base64.StdEncoding.EncodeToString(zbuf),
 	}
-	return resp, nil
+	w.Header().Add("Content-Type", "application/json")
+	json.NewEncoder(w).Encode(bundle)
```
owenrhysthomas, line 138: no longer want to return anything here? I noticed you removed the return values from above, so probably you don't need to return, just wanted to bring your attention just in case.

nonrenewable (approved):
Does what it says on the tin! Just checking on some things i don't exactly myself!

visaservice/core/pkg/vservice/constants.go
```diff
@@ -36,9 +33,6 @@ const (
 	// before the old configuration expires.
 	NetConfigSettleTime = 10 * time.Second
 
-	// Visa service signs the agents it authenticates under this key ID.
-	SigningKeyID = "vs"
-
```
owenrhysthomas, line 41: was this only used for the gRPC?

visaservice/README.md
mkolehmainen: Yes, there are still protocol buffer things in here.

visaservice/core/pkg/vservice/admin.go
mkolehmainen: Writing to that `w` thing does return bytes to the caller.

owenrhysthomas: Yes, that's what I figured. Makes sense!

visaservice/README.md
mkolehmainen: This API is "human" facing so it is HTTP/json for simplicity.  Our current plan is to use THRIFT for internal APIs.  This particular API is designed for an administrator to manipulate the visa service.  Probably will be through a command line or GUI tool.

visaservice/core/pkg/vservice/constants.go
mkolehmainen: Yes. This signing stuff is used in the prototype because it has a convoluted way of creating authenticated agents, and then handing them back to the node (which stores them in RAFT), and then later requesting them back and at that stage the signature was checked.  We don't need that any more since we are now keeping the agent database on the visa service.

visaservice/README.md
nonrenewable: gotcha, that all makes sense. I guess i was also trying to figure out if this is part of the admin service but based on what you're saying it sounds like it's visa-service specific? I thought it ran another listener rather than the original visa service why i'm a bit confused. Maybe it doesn't matter that i understand this all now or it makes more sense to talk about 🤷

mkolehmainen: You are correct that this part of, and at the moment probably _all of_ the admin service.  Minimal admin service function is install policy, update policy, & revoke something -- which as it happens are all visa service functions.

nonrenewable: i gotcha!! i'm following ✅

Issue #42, opened 2024-06-05
Figure out why PH tokio isn't using multiple OS threads


mkolehmainen:
Despite that we're using [`Builder::new_multi_thread()`](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.new_multi_thread), the Tokio runtime in the PH appears to be running only on one OS thread.  (This can be seen by running a workload through the PH, and observing in `top -H` that only one thread is using 100% of its core.)

Looking now this might be as simple as using the [`Builder::worker_threads()`](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.worker_threads) method, but it should be defaulting to the number of cores on the system, not 1.

The [`Runtime::metrics()`](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.metrics) method might be useful to figure out what's going on too.

It's also possible that, since the DTLS worker is the primary source of CPU-intensive work, Tokio realizes this and doesn't bother spinning up other OS threads for the other workers.  It would be nice to understand the mechanism behind this choice if this is true.

cpacejo:
I'm OK to call this done for now!  You found the answer to the original question (Tokio _is_ using multiple OS threads).  We do need a deeper performance dive, given the messy performance profile you found Tokio to have, but let's hold off on that for a bit.

PR #153, opened 2024-07-02
vm first two performance tests

### LINKED ISSUE #151: performance testing of PH
Let's do some basic performance testing and analysis.

You'll want to set up a pair of PHs connected to each other (i.e. in different virtual machines) and use `iperf` (`apt install iperf`) to generate some traffic between them.  (The basic configuration of running `iperf -s` on one machine and `iperf -c <server TUN IP>` on the other machine should be sufficient to start with.)  You'll want to set the TUN devices' MTUs to 1400 to ensure too-large packets aren't being sent.  You'll want to test in both directions (the `iperf` client sends data, while the server receives).

This will saturate the PH pipeline, meaning there's a bottleneck somewhere.  Use the performance-testing debug tool you made to measure the queue depths and latencies of the PH pipeline on both ends.  You should see full queues before the bottleneck point, and empty-ish queues after the bottleneck point.  (It's possible/likely the bottleneck is the first/final stage of the PH, the DTLS worker.)

The service time reported at the bottleneck point is now of interest – divide the average service time by the average queue depth for a rough indication of the average processing time per packet.  Invert this figure, it should correspond to the throughput in packets per second.  (`iperf` reports payload bandwidth – each packet will contain MTU-40 bytes of payload, or 1360 bytes per packet at MTU=1400.)

You'll also want to experiment with using UDP traffic (`-u` option to `iperf`, for which you'll also need to specify a bandwidth via `-b` – try higher and higher bandwidths until you find one that all traffic doesn't make it through – and also a packet size via `-l` – try packets of lengths 0, 24, 500, and 1200 as examples).  Notably, changing the packet size may change where the bottleneck lies.

This is mostly exploratory work – there's not a specific deliverable – but we do want to identify the bottleneck in each direction so we know what to focus our performance work on.

owenrhysthomas:


adapter/ph/performance_testing_udp.txt
```diff
@@ -0,0 +1,113 @@
+Devices: Laptop running Ubuntu 22.04 and VM running on laptop running Debian 12
+
+TEST 1: Server running on laptop, client on VM. PERF-SAMPLE running on server side 
+        for 30 seconds 4 samples/seconds. iperf -s -u -b 10000 on server side
+        iperf -s -u 192.168.99.1 -u -b 700000000 on client
+
+    Inbound Processor Duration values at - 10th Quantile: 3327ns, 25th Quantile: 3711ns,
+    50th Quantile: 5887ns, 75th Quantile: 32767ns, 90th Quantile: 55295ns
+
```
owenrhysthomas, line 9: I'm not sure why there's no mean here? I'll have to re-run these tomorrow.

adapter/ph/performance_testing.txt
```diff
@@ -0,0 +1,113 @@
+Testing performance for bottlenecks in system
+
+Devices: Laptop running Ubuntu 22.04 and VM running on laptop running Debian 12
+
+Test 1: Server running on laptop, client on VM. PERF-SAMPLE running on server side 
+        for 120 seconds 4 samples/seconds. iperf -s on server side.
```
cpacejo, line 6: One thing I'm curious is what `iperf` reported also, just to have these figures in context.

owenrhysthomas: I'll do the tests again and report the iperf as well.

cpacejo:
Another view of the data which might be worth reporting is the histogram itself, using e.g. [`iter_log`](https://docs.rs/hdrhistogram/latest/hdrhistogram/struct.Histogram.html#method.iter_log).  It's technically the same information as the quantiles, but because it's with respect to the time metric, rather the the sample count, I find it a bit easier to visualize.

adapter/ph/performance_testing.txt
```diff
@@ -0,0 +1,113 @@
...
+
+    Inbound Processor Duration values at - 10th Quantile: 4095ns, 25th Quantile: 6143ns,
+    50th Quantile: 36863ns, 75th Quantile: 77823ns, 90th Quantile: 139263ns
```
cpacejo, line 9: I'm surprised how high these figures are!  Inbound processor is doing almost nothing, and its queue is almost empty.  Service time for test packets placed in this queue I would expect to be in the sub-µs range… (or at least, the 50% quantile to be close to the 10%…)

adapter/ph/performance_testing.txt
```diff
@@ -0,0 +1,113 @@
...
+    50th Quantile: 425983ns, 75th Quantile: 524287ns, 90th Quantile: 589823ns
+
+    Outbound Processor Depth values at - 10th Quantile: 0 packets, 25th Quantile: 16 packets,
+    50th Quantile: 16 packets, 75th Quantile: 16 packets, 90th Quantile: 16 packets
+
+    Outbound Send Duration values at - 10th Quantile: 94207ns, 25th Quantile: 180223ns,
+    50th Quantile: 245759ns, 75th Quantile: 294911ns, 90th Quantile: 344063ns
+
+    Outbound Send Depth values at - 10th Quantile: 0 packets, 25th Quantile: 0 packets,
+    50th Quantile: 0 packets, 75th Quantile: 0 packets, 90th Quantile: 0 packets
```
cpacejo, line 57: Actually this is really weird!  The outbound processor isn't doing much of anything, the outbound send stage should be the bottleneck (since it's doing DTLS encryption).  But the outbound send queue is empty!.. not understanding why this would be, hmm…

adapter/ph/performance_testing.txt
cpacejo: Ah so I think the reason for this is actually #42 – everything is running on a single OS thread, which, since the PH as a whole is oversubscribed, results in all stages backing up behind one another in Tokio's scheduler.  So if any one stage takes a while processing something, every other stage which is preempted while processing (e.g. while enqueuing another packet from the batch onto another queue) also accounts that latency.  Fixing #42 I would expect should result in much cleaner metrics here.

cpacejo:
Definitely interesting results and I need to still think about the implications of some of them!  So excited to see this all working.

So I wonder where the best place to keep these results are.  I don't think they should be checked in to the main repo, but also I can't think of a better place to put them right now.  (Probably eventually, we'll have a runner which does something like this automatically, and inserts the results into an S3 bucket or some sort of database.)  So feel free to check them in for now.

cpacejo (approved):

PR #154, opened 2024-07-02
add "request visa" function to visa service api test client

### LINKED ISSUE #148: (backfill) support the visa request call in the visaservice test client
This was not a real story, but I did it anyway.  The visa service test client `visaservice/cli` can now be used to call the visa-request function.

mkolehmainen:
(take 2)


mkolehmainen:
This is new version of old PR after frightful rebase.

visaservice/cli/src/traffic_parser.rs
```diff
@@ -0,0 +1,261 @@
...
+            let res = parse_traffic(input, Protocol::TCP);
+            if res.is_err() {
+                println!("failed to parse valid input '{}', Error: {:?}", input, res);
+                assert!(false)
+            }
+        }
+    }
+
+    #[test]
+    fn test_parse_questionable() {
```
svjaffer, line 234: What's the plan for this?  And what does the production code do with these invalid/malformed flags?

mkolehmainen: The parser is basically a best-effort regex.  All the parsing is in the test client, so it is not in a production path.  Of course, if this client becomes useful then we can improve.  Another question might be, what happens when we pass a bunch of bogus crap to the visa service traffic matcher?  I'm not sure.  There are some tests left over from prototype but probably not enough.

visaservice/cli/src/traffic_parser.rs
```diff
@@ -0,0 +1,261 @@
...
+            None => {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    "Invalid input",
+                ))
+            }
+        }
+    } else {
+        // IPv4
+        let re = Regex::new(r"^(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})(?::(\d+))?>(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}):(\d+)(?:\[([SA]+)\])?").unwrap();
```
cpacejo: [Verbose mode](https://docs.rs/regex/latest/regex/index.html#example-verbose-mode) allows whitespace and comments which may make these easier to read.

mkolehmainen: Ah, that's interesting.  Let me see.

visaservice/cli/src/traffic_parser.rs
svjaffer: Can you make it more clear in the naming that this is test functionality instead of production functionality?

mkolehmainen: yes let me try that.

svjaffer (approved):

PR #156, opened 2024-07-03
155 report also histogram buckets

### LINKED ISSUE #155: report also histogram buckets
In the perf testing code, let's report the histogram bucket counts also, in addition to the quantiles.  So we might see something like:

queue depth | samples
0-1 | 5
2-3 | 8
4-7 | 15
8-15 | 5
16-31 | 1

(sorry about the formatting – just trying to convey the idea with some numbers!)

Using e.g. [`iter_log`](https://docs.rs/hdrhistogram/latest/hdrhistogram/struct.Histogram.html#method.iter_log) should get these values.

owenrhysthomas:


cpacejo (approved):
Looks good, thanks!

PR #158, opened 2024-07-04
visa service process for the node

mkolehmainen:
I am opening this PR for some early comments. This is _not_ ready to be merged -- for one thing there are no unit tests.


node/src/vsapi/mod.rs
```diff
@@ -0,0 +1,2608 @@
+// Autogenerated by Thrift Compiler (0.20.0)
```
cpacejo, line 1: I would advise we do not check in generated files.  Let's always generate them as part of the build process.

Actually… where is the source for this?  Is it already checked in?

visaservice/thrift/Makefile
```diff
@@ -21,8 +28,11 @@ install-go: go
 	test -d $(DEST_GO) || mkdir $(DEST_GO)
 	cp $(GGO)/vsapi/*.go $(DEST_GO)
 
+
 install-rs: rs
```
cpacejo, line 32: Should be mentioned in .PHONY?

node/src/vs/vsconn.rs
```diff
@@ -0,0 +1,387 @@
...
+//   At this point the node is "registered" with the visa service.
+//   TODO: The node should call authroize_connect on itself once registered.
+//
+//   Now the node can request visas or call for connect authorizations.
+//
+//   Also the node should set up a polling loop to respond to pushed visas and revocations
+//   from the visa service.
+//
+
+const POLL_INTERVAL_MS: u64 = 5000;
```
cpacejo: Maybe use a [`Duration`](https://doc.rust-lang.org/std/time/struct.Duration.html) here?

node/src/config.rs
```diff
@@ -0,0 +1,119 @@
...
+struct Creds {
+    certificate: String, // this nodes signed certificate file
+    private_key: String, // this nodes private key file
+}
+
+impl Configuration {
+    pub fn get_cert_path(&self) -> String {
+        let base = Path::new(&self.base_path);
+        let cert_pb = base.join(&self.creds.certificate);
+        return cert_pb.to_str().unwrap().to_string();
```
cpacejo, line 32: I think you can return a `PathBuf` here, or `Box<Path>`.

node/src/config.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub fn get_claims(&self) -> HashMap<String, String> {
+        let mut hmap: HashMap<String, String> = HashMap::new();
+        for (k, v) in self.claims.iter() {
+            hmap.insert(k.to_string(), v.to_string());
+        }
+        hmap
+    }
+
+    pub fn get_claim(&self, key: &str) -> Option<String> {
+        match self.claims.get(key) {
```
cpacejo, line 51: Can be written also as `Some(self.claims.get(key)?.as_str()?.to_string())`

node/src/config.rs
```diff
@@ -0,0 +1,119 @@
...
+        let base = Path::new(&self.base_path);
+        let cert_pb = base.join(&self.creds.private_key);
+        return cert_pb.to_str().unwrap().to_string();
+    }
+
+    // Gets a copy of the claims
+    pub fn get_claims(&self) -> HashMap<String, String> {
+        let mut hmap: HashMap<String, String> = HashMap::new();
+        for (k, v) in self.claims.iter() {
+            hmap.insert(k.to_string(), v.to_string());
```
cpacejo, line 45: `to_string()` is part of the `Display` mechanism: https://doc.rust-lang.org/std/string/trait.ToString.html#tymethod.to_string  My understanding is it's intended for user-facing display.  I think `String::from(...)` is what we want here.

(Same comment for the other occurrences in this file…)

node/src/config.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub fn get_claim(&self, key: &str) -> Option<String> {
+        match self.claims.get(key) {
+            Some(v) => v.as_str().map(|s| s.to_string()), // Double decode since v.as_str() from Table includes quotes.
+            None => None,
+        }
+    }
+
+    pub fn get_node_addr(&self) -> IpAddr {
+        if let Some(a) = &self.node_addr {
+            *a // somehow this de-ref returns a copy of the addr ???? XXX
```
cpacejo: `a` here is a reference, since you're matching against a reference (`&self.node_addr`).  So the `*` is needed to dereference that (which then results in a copy being made to return it).  I would expect you could drop both the `&` and `*` and things should work.

node/src/config.rs
```diff
@@ -0,0 +1,119 @@
...
+        } else {
+            panic!("node address not set in Configuration"); // This fair since the address is required in load_configuration
+        }
+    }
+}
+
+pub fn load_configuration(path: &str) -> Result<Configuration, std::io::Error> {
```
cpacejo, line 66: `Path` instead of `&str`?

node/src/config.rs
```diff
@@ -0,0 +1,119 @@
...
+    let file = std::fs::File::open(path)?;
+    let mut reader = std::io::BufReader::new(file);
```
cpacejo: I think `File` directly implements `read_to_string()`, which is already reading in large chunks, so the `BufReader` I think is unnecessary.

node/src/config.rs
```diff
@@ -0,0 +1,119 @@
...
+use serde::Deserialize;
+use std::collections::HashMap;
+use std::io::Read;
+use std::net::IpAddr;
+use std::path::Path;
+
+#[derive(Debug, Clone, Deserialize)]
+pub struct Configuration {
+    #[serde(skip)]
+    base_path: String,
```
cpacejo, line 12: Store these as `PathBuf`s instead maybe?

cpacejo (changes requested):
Still have some files to review, checkpointing my comments for now…

visaservice/thrift/Makefile
mkolehmainen: y

node/src/config.rs
mkolehmainen: yes thanks

node/src/vsapi/mod.rs
mkolehmainen: Eh... I don't want to require the thrift build chain in order to build the project.  These files should change very infrequently (once we get to a certain point...).

mkolehmainen: I have created an issue org-zpr/zpr-core#196 for our new "manifest" idea.

PR #162, opened 2024-07-05
Added batch sizes into PERF-SAMPLE test

### LINKED ISSUE #157: measure also batch sizes
In addition to service time and queue depths, a useful metric to measure during the performance testing might be batch size.  This can be measured the same way as queue size – when the test packet arrives, record the size of the batch it arrived in.  The batch size is just the `count` value returned by the `recv_many()` method which reads messages from the queues.  This value won't immediately tell us anything about the health of the system, but can help us diagnose whether batching is working effectively.

owenrhysthomas:


cpacejo (approved):
Fantastic thank you!!

PR #167, opened 2024-07-08
packet cloning

cpacejo:
We need this for packet capture…


owenrhysthomas (approved):
Cool! Looks good.

owenrhysthomas:
assert! panics if it fails, right? So clone_into_with_headrooom will panic if the user supplies too small of a headroom?

PR #168, opened 2024-07-08
Implemented capture queue, still needs testing

### LINKED ISSUE #164: Implement Capture queue
Let's add a new queue type, `Capture`, which represents a packet capture worker.  It should have one message type, `Packet`, which has both a `packet` field (like the other messages), and _also_ a timestamp field (a `std::time::Instant`).  It should have both an async `enqueue_packet()` method, and a non-async `try_enqueue_packet()` method, which uses `try_send()` to enqueue the packet.  Both of these should take the packet as an argument, and fill the timestamp similarly to how we do for test packets.

Unlike `enqueue_packet()` (which returns `()`), the `try_enqueue_packet()` method should return a `Result<(), TryEnqueueError<T>>`.  `TryEnqueueError<T>` can be an enum of one value, `Full(T)`, for now.  If `try_send()` fails with `Full(x)`, return `Full(x)` also.  If `try_send()` fails otherwise, simply panic.

Unlike the other queues, no need for test packets here.

This queue should be added to the assembly like the other queues.

owenrhysthomas:


adapter/ph/src/main.rs
```diff
@@ -140,6 +140,7 @@ fn main() -> ExitCode {
     let outbound_processor_batch_size = 16;
     let outbound_send_queue_size = 16;
     let outbound_send_batch_size = 8;
+    let capture_size = 8;
```
cpacejo: Nit – I'd call this `capture_queue_size` specifically.  (`capture_size` I feel could equally mean the maximum size of packet to capture, which is a setting we might want to add later.)

adapter/ph/src/packet.rs
```diff
@@ -7,6 +7,7 @@ use zerocopy_derive::{AsBytes, FromBytes, FromZeroes, KnownLayout, Unaligned};
 // This contains all state of a packet which is moving through the system.
 // TODO: possible we want to keep this stuff on the heap
 
+#[derive(Debug)]
```
cpacejo: Hm, this feels off to me – I think any debug representation of a packet we'd want to implement manually.  (As-is, I think this will dump out a hex representation of the entire packet buffer which isn't the greatest and maybe even has security implications.)  I'm guessing this is needed because of the panic message below (by way of the `Debug` trait on `CapPacket`)?

adapter/ph/src/queues.rs
```diff
@@ -163,3 +164,46 @@ impl<'pktbuf> OutboundSend<'pktbuf> {
...
+    pub fn try_enqueue_packet(
+        &self,
+        packet: Packet<'pktbuf>,
+        timestamp: std::time::Instant,
+    ) -> Result<(), TryEnqueueError<CapPacket<'pktbuf>>> {
+        let cap_pack: CapPacket = CapPacket { packet, timestamp };
+        let _result = match self.sender.try_send(cap_pack) {
+            Ok(()) => return Ok(()),
+            Err(TrySendError::Full(x)) => return Err(TryEnqueueError::Full(x)),
+            Err(TrySendError::Closed(x)) => panic!("panicking! {:?}", x),
```
cpacejo: I'm guessing this is the reason that the `Debug` trait was needed on `Packet` above – since `x` here is the packet we tried to send (it's returned so we have ownership again).  I think it's fine not to log `x` here – the real error is that the queue was closed.

Aside – I know I said to panic here but I'm having second thoughts, since that's adding a failure mode to the main data path.  Maybe it's actually better to return the `Full` error rather than panicking.  (I wouldn't bother adding a `Closed` value to the enum above, simply because there's nothing different the caller can/should do.)

(The implicit panic (via unwrap) in `enqueue_packet()` I think is fine… the alternative being blocking forever, which is worse.)

cpacejo (approved):
Yes this is what I had in mind, thanks!  Just a couple comments but feel free to address those in a separate PR if you just need to get this checked in.

cpacejo (approved):
Changes look good thank you!!

Issue #166, opened 2024-07-08
~~[BLOCKED]~~ Implement capture_worker


cpacejo:
The other half of `capture_worker` is the worker itself.  We can define it similarly to `inbound_send_worker`, except instead of writing packets to the TUN interface, it writes them to the currently configured `Savefile`.  After receiving a batch of packets from the `Capture` queue, it will have to take the `Savefile` mutex from the `capture_worker` state struct, and write them to it via [`write()`](https://docs.rs/pcap/latest/pcap/struct.Savefile.html#method.write).  You'll have to create the pcap `PacketHeader` yourself from the timestamp from the message, and the length of the packet's `body()` (which you'll also use for the packet data itself).  If there's no `Savefile`, don't do anything.  Either way, then return the packets to the buffer stack.

Blocked on:
- org-zpr/zpr-core#164
- org-zpr/zpr-core#165

PR #169, opened 2024-07-09
165 implement capture worker state

### LINKED ISSUE #165: Implement capture_worker state
Let's make a new worker, `capture_worker`, which writes packets to a capture file.

`capture_worker` will consist of two parts, the capture state, and the worker itself.  This task captures defining the state.

See `buffer_stack.rs` for an example of how to define the capture state.  It will be a struct which will have to track the [`Capture`](https://docs.rs/pcap/latest/pcap/struct.Capture.html) object, which is initialized on creation of the struct, and the current [`Savefile`](https://docs.rs/pcap/latest/pcap/struct.Savefile.html) object, which can be created and destroyed.  (So it should be wrapped in an `Option`.)  Additionally – the `Savefile` will need to be wrapped in a Tokio `Mutex`.  (So you'll have a `Mutex<Option<Savefile>>`.)

This struct should have methods to open, flush, and close a capture file.  (Names like `open_capture_file()` are probably fine.)  `open()` of course will need a `Path` argument, the others no arguments.  These take the mutex as appropriate, and use [`Capture::savefile`](https://docs.rs/pcap/latest/pcap/struct.Capture.html#method.savefile), [`Savefile::flush`](https://docs.rs/pcap/latest/pcap/struct.Savefile.html#method.flush), and destroying the `Savefile` object (which causes a `drop()`… but you might need to explicitly call [`std::mem::drop()`](https://doc.rust-lang.org/std/mem/fn.drop.html) to placate the borrow checker).

The struct should be added to the assembly.

owenrhysthomas:


owenrhysthomas:
~~We need to install libpcap-dev on GitHub, that's the reason that the PH runner can't compile the code and it won't link.~~ Done

adapter/ph/src/capture_worker.rs
```diff
@@ -0,0 +1,44 @@
+use pcap;
+use std::mem::drop;
+use tokio::sync::Mutex;
+
+#[allow(dead_code)]
```
cpacejo: Since these are referenced in the assembly probably the `dead_code` annotations are no longer necessary?

.github/workflows/rust-adapter.yml
```diff
@@ -43,7 +43,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Build + test ph
-      run: cd adapter/ph && make build && make test
+      run: sudo apt update && sudo apt install libpcap-dev && cd adapter/ph && make build && make test
```
cpacejo, line 46: Ah nice good catch!

adapter/ph/src/capture_worker.rs
```diff
@@ -0,0 +1,44 @@
...
+    capture: pcap::Capture<pcap::Dead>,
+    savefile: Option<pcap::Savefile>,
+}
+
+#[allow(dead_code)]
+impl CaptureWorker {
+    pub fn new() -> Self {
+        Self {
+            inner_cap: InnerCap {
+                capture: pcap::Capture::dead(pcap::Linktype(0)).unwrap(), // not sure what Linktype this should be
```
cpacejo: We will need to register our own! :)  But for now, use `USER0`.

adapter/ph/src/capture_worker.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    pub fn flush_savefile(&mut self) -> Result<(), pcap::Error> {
+        // self.savefile.get_mut().as_mut().unwrap().flush()
+        self.inner_cap.get_mut().savefile.as_mut().unwrap().flush()
+    }
+
+    // Have to use .into_inner() here to get ownership of the savefile in order to drop it,
+    // can't use get_mut() because calling drop() on a non-owned value does nothing.
+    // Means can't use entire CaptureWorker struct after calling this function.
+    pub fn destroy_savefile(self) {
```
cpacejo: I think just setting `self.savefile = None` should do the trick here (and thus you can use `&mut self`).  That will cause the savefile's drop method to be executed automatically.  (Note that we're implicitly relying on this behavior in `open_capture_file` as well, if a capture file is already open.)

If you did want to explicitly call drop I think you can do `self.savefile.take().map(drop)`.  (`take()` moves the item out of the option.)  I don't think that's necessary though.

Aside – I'd use the term "close" rather than "destroy"; "destroy" to me sounds like the underlying file might be deleted.

adapter/ph/src/capture_worker.rs
```diff
@@ -0,0 +1,44 @@
...
+            }
+            .into(),
+        }
+    }
+    pub fn open_capture_file(&mut self, path: String) {
+        self.inner_cap.get_mut().savefile =
+            Some(self.inner_cap.get_mut().capture.savefile(path).unwrap())
+    }
+
+    pub fn flush_savefile(&mut self) -> Result<(), pcap::Error> {
```
cpacejo: Let's use `capture_file` instead of `savefile` for all three of these to keep them consistent.

adapter/ph/src/capture_worker.rs
```diff
@@ -0,0 +1,44 @@
...
+    pub fn new() -> Self {
+        Self {
+            inner_cap: InnerCap {
+                capture: pcap::Capture::dead(pcap::Linktype(0)).unwrap(), // not sure what Linktype this should be
+                savefile: None,
+            }
+            .into(),
+        }
+    }
+    pub fn open_capture_file(&mut self, path: String) {
```
cpacejo: Let's accept a [`Path`](https://doc.rust-lang.org/nightly/std/path/struct.Path.html) here instead of a `String`.

cpacejo (changes requested):
Just a couple changes in the capture file API (specifically there's a way we can do closing without consuming the entire mutex).

adapter/ph/src/capture_worker.rs
```diff
@@ -0,0 +1,44 @@
...
+        self.inner_cap.get_mut().savefile =
```
cpacejo: Ah, yes, here you will want to use `.lock().await`, not `.get_mut()`.  Then, all these methods can accept `&self` instead of `&mut self`.

adapter/ph/src/capture_worker.rs
owenrhysthomas: ~~I changed the way I do these methods (I had to make them async so I could use the combination of .lock().await to access the object in the Mutex instead of .get_mut(), so I'll make this change in issue https://github.com/org-zpr/zpr-core/issues/166 (PR https://github.com/org-zpr/zpr-core/pull/180)~~ Never mind, I will just apply the changes in this PR

adapter/ph/src/capture_worker.rs
```diff
@@ -0,0 +1,44 @@
...
+        Self {
+            inner_cap: InnerCap {
+                capture: pcap::Capture::dead(pcap::Linktype(USER0)).unwrap(), // not sure what Linktype this should be
+                savefile: None,
+            }
+            .into(),
+        }
+    }
+    pub async fn open_capture_file(&self, path: &Path) {
+        self.inner_cap.lock().await.savefile =
```
cpacejo: You'll want to just take the lock once here.  E.g.:
```
let inner_cap = self.inner_cap.lock().await;
inner_cap.savefile = inner_cap.capture.savefile(path).unwrap();
```

adapter/ph/src/capture_worker.rs
```diff
@@ -0,0 +1,44 @@
...
+            Some(self.inner_cap.lock().await.capture.savefile(path).unwrap())
+    }
+
+    pub async fn flush_capture_file(&self) -> Result<(), pcap::Error> {
+        self.inner_cap
+            .lock()
+            .await
+            .savefile
+            .as_mut()
+            .unwrap()
```
cpacejo: Rather than `unwrap()` here, let's simply return success if there's no active savefile.

cpacejo (changes requested):

cpacejo (approved):
Perfect thanks!

PR #173, opened 2024-07-09
add try_get_buffers

cpacejo:
We need a way to attempt to get buffers without blocking, for packet capture (which should never block the datapath).


owenrhysthomas (approved):
Nice!

PR #180, opened 2024-07-10
166 blocked implement capture worker

### LINKED ISSUE #166: ~~[BLOCKED]~~ Implement capture_worker

owenrhysthomas:


adapter/ph/src/capture_worker.rs
```diff
@@ -9,36 +15,115 @@ pub struct CaptureWorker {
...
+pub struct Config {
+    pub batch_size: usize,
+}
+
+#[allow(dead_code)]
+async fn worker<'pktbuf>(
+    config: &Config,
+    asm: &Assembly<'pktbuf>,
+    queue: &mut mpsc::Receiver<CapPacket<'pktbuf>>,
+    path: String,
```
cpacejo: The capture path we'll be setting via an RPC call, no need to accept it as an argument here.

adapter/ph/src/capture_worker.rs
```diff
@@ -9,36 +15,115 @@ pub struct CaptureWorker {
...
+    let mut messages = Vec::new();
+    let mut savefile_exists = false;
+
+    if asm.capture_worker.inner_cap.lock().await.savefile.is_some() {
+        asm.capture_worker.open_capture_file(path).await;
+        savefile_exists = true;
+    }
+
+    while let _count @ 1.. = queue.recv_many(&mut messages, config.batch_size).await {
+        for cap_pack in &messages {
```
cpacejo: Right here (inside the `while`, outside the `for`) you'll want to grab the mutex of `inner_cap`, and while holding it, check whether the savefile exists (since concurrently, the RPC worker can open and close the savefile), i.e. with a `match`.  Then, inside the `for`, loop through and write everything to the savefile you found.  (Or, if there was no savefile, just skip the `for` entirely.)

(For reference – you can write `let lock = foo.lock().await;` to grab and hold the mutex for the rest of the lexical code block.)

(The idea being to only grab the mutex once for each batch of packets, rather than once for every packet.)

adapter/ph/src/queues.rs
```diff
@@ -207,3 +211,13 @@ impl<'pktbuf> Capture<'pktbuf> {
         };
     }
 }
+
+impl<'pktbuf> CapPacket<'pktbuf> {
```
cpacejo, line 215: I think public members like you have above is fine – so this `impl` block can be safely deleted.

cpacejo (changes requested):
Just needs a change in the logic to check for an active savefile – my description is probably not very clear so we can chat if you have questions :)  Otherwise looks good!

adapter/ph/src/capture_worker.rs
owenrhysthomas: Got it. I was originally going to use a match, but I wasn't sure if calling `open_capture_file` inside the `while` would cause problems because it's repeatedly opening the file.

adapter/ph/src/capture_worker.rs
```diff
@@ -38,7 +45,73 @@ impl CaptureWorker {
...
+) {
+    let mut messages = Vec::new();
+    let path = Path::new("temp"); // temporary until RPC is implemented
+
+    while let _count @ 1.. = queue.recv_many(&mut messages, config.batch_size).await {
+        let mut locked_mutex = asm.capture_worker.inner_cap.lock().await;
+        for cap_pack in &messages {
+            match &mut locked_mutex.savefile {
+                Some(s_file) => {
+                    asm.capture_worker.open_capture_file(&path).await;
```
cpacejo: The RPC worker will be opening and closing the file.  So here you'll just want to write to it.

adapter/ph/src/capture_worker.rs
```diff
@@ -38,7 +45,73 @@ impl CaptureWorker {
...
+fn savefile_write(cap_pack: &CapPacket, savefile: &mut Savefile) {
+    let creation_time: Duration = cap_pack.timestamp.duration_since(UNIX_EPOCH).unwrap();
+    let ts: timeval = timeval {
+        tv_sec: creation_time.as_secs() as i64,
+        tv_usec: creation_time.subsec_micros() as i64,
+    };
+
+    let header: PacketHeader = PacketHeader {
+        ts,
+        caplen: cap_pack.packet.body().len() as u32, // not sure if this is the right value, perhaps should be cap_pack.packet.metadata().len
```
cpacejo, line 102: This is correct!  (`caplen` is how much we are recording to the capture file, which we are recording the full packet body, so it's equal to `len`.)

adapter/ph/src/flow_control.rs
```diff
@@ -0,0 +1,78 @@
+use std::sync::atomic::{AtomicBool, Ordering};
+
+pub struct FlowControl {
+    copy_inbound_to_capture: AtomicBool,
+    copy_outbound_to_capture: AtomicBool,
+}
+
+#[allow(dead_code)]
+impl FlowControl {
+    pub(crate) fn new(inbound: bool, outbound: bool) -> Self {
```
cpacejo: I would just initialize these both to false, no need to accept as parameters here.

cpacejo (changes requested):

adapter/ph/src/capture_worker.rs
```diff
@@ -38,7 +45,70 @@ impl CaptureWorker {
...
+    config: &Config,
+    asm: &Assembly<'pktbuf>,
+    queue: &mut mpsc::Receiver<CapPacket<'pktbuf>>,
+) {
+    let mut messages = Vec::new();
+
+    while let _count @ 1.. = queue.recv_many(&mut messages, config.batch_size).await {
+        let mut locked_mutex = asm.capture_worker.inner_cap.lock().await;
+        for cap_pack in &messages {
+            match &mut locked_mutex.savefile {
```
cpacejo, line 66: Not a bug, but you can hoist this out of the for loop (i.e., just perform the `match` once, and then either loop or don't).  It won't be changing out from under you because you are holding the lock.  (Feel free to change this in a later PR.)

cpacejo (approved):
Looks good!  One nit but feel free to commit as-is.

adapter/ph/src/capture_worker.rs
owenrhysthomas: This makes sense, I realized I misread your original comment about the match, I just fixed it, pushing right now.

PR #181, opened 2024-07-10
Added flow control

### LINKED ISSUE #170: Add flow control struct
We'll need to add a new shared struct, part of the assembly, let's call it `FlowControl`.

It'll look similar to the counters struct, but its use is dynamically configuring the flow of packets through the system.

For now it will have two members, `copy_inbound_to_capture`, and `copy_outbound_to_capture`.  These will both be `AtomicBool` with appropriate set/get accessors.  They can be accessed with "relaxed" ordering since they're not part of any synchronization scheme.

(I don't love my naming of this or my proposed implementation, as I feel these are the first two example of a larger class of settings which I can't think of other examples of yet.  But these settings need to go somewhere and this works for now.)

owenrhysthomas:


cpacejo (approved):
Looks perfect thanks!!

(Meanwhile… I am researching RCU implementations for Rust so we can have a better system for managing these runtime configuration settings…)

PR #182, opened 2024-07-11
171 blocked copy packets to capture queue

### LINKED ISSUE #171: [BLOCKED] Copy packets to capture queue
The `inbound_processor` and `outbound_processor` both need to have the ability to copy packets to the capture queue.

For each batch of packets they process, you'll want to check the respective `copy` flag in the `flow_control` struct.  If true, then try to grab a number of buffers from the buffer stack equal to the number of packets in the batch.  (You may get back fewer buffers than requested, including 0 – if so, we'll make do with that amount.)  Copy each packet's data into one of these buffers using `Packet::clone_into`.  Then try to enqueue the packet in the capture queue – if that fails, delete the packet and return the buffer to the buffer stack.

Test packets, just ignore.  This means you might have grabbed too many buffers; that's OK, just return the leftovers to the buffer stack.

Where in the processing flow this goes isn't critical right now; easiest to appease the borrow checker is probably to do this as the first thing after receiving a batch of packets.

Blocked on:
- org-zpr/zpr-core#167
- org-zpr/zpr-core#170
- org-zpr/zpr-core#173

owenrhysthomas:


adapter/ph/src/capture_worker.rs
```diff
@@ -3,12 +3,12 @@ use crate::CapPacket;
 use core::future::Future;
 use libc::timeval;
 use pcap::{Capture, Dead, Error, Linktype, Packet, PacketHeader, Savefile};
+use std::path::Path;
 use std::time::{Duration, UNIX_EPOCH};
 use tokio::sync::mpsc;
-
 use tokio::sync::Mutex;
+
 pub const USER0: i32 = 147;
```
cpacejo: I think I missed this in an earlier review – why not use `https://docs.rs/pcap/latest/pcap/struct.Linktype.html#associatedconstant.USER0` directly?

adapter/ph/src/inbound_processor_worker.rs
```diff
@@ -23,6 +25,29 @@ async fn worker<'pktbuf>(
...
+                        {
+                            Ok(()) => (),
+                            Err(TryEnqueueError::Full(ret_capture_packet)) => {
+                                let ret_buf = ret_capture_packet.packet.destroy();
+                                asm.buffer_stack.put_buffer(ret_buf);
+                            }
+                        }
+                    }
+                    InboundProcessorMessage::TestPacket(_pkt) => asm.buffer_stack.put_buffer(buf),
+                    // TODO potentially restrustructure, this wastes a buffer if there are lots of test packets
```
cpacejo: I think if you write the loop in terms of `pkts` instead, you should be able to `pop()` buffers off of `bufs`.  And then just pass the remaining `bufs` to `put_buffers()`.

adapter/ph/src/inbound_processor_worker.rs
```diff
@@ -23,6 +25,29 @@ async fn worker<'pktbuf>(
...
+                    InboundProcessorMessage::Packet(pkt) => {
+                        let pkt_clone: Packet = pkt.clone_into(buf);
+                        match asm
+                            .capture_queue
+                            .try_enqueue_packet(pkt_clone, SystemTime::now())
+                        {
+                            Ok(()) => (),
+                            Err(TryEnqueueError::Full(ret_capture_packet)) => {
+                                let ret_buf = ret_capture_packet.packet.destroy();
+                                asm.buffer_stack.put_buffer(ret_buf);
```
cpacejo: Here, I would exit the loop also.  Subsequent enqueues are almost certain to fail.

Also let's count failed captures here as well.  (A separate PR is fine for this to avoid PR scope creep.)

adapter/ph/src/main.rs
```diff
@@ -294,6 +294,14 @@ fn main() -> ExitCode {
 
             js.spawn(rpc_worker::launch(&*asm, &*unix_socket));
 
+            js.spawn(capture_worker::launch(
+                &capture_worker::Config {
+                    batch_size: capture_queue_size,
```
cpacejo: Like the other queues, let's configure this in terms of desired batch size instead of queue size, and make the queue size 2× the batch size.

adapter/ph/src/queues.rs
```diff
@@ -203,21 +203,18 @@ impl<'pktbuf> Capture<'pktbuf> {
         timestamp: SystemTime,
     ) -> Result<(), TryEnqueueError<CapPacket<'pktbuf>>> {
         let cap_pack: CapPacket = CapPacket { packet, timestamp };
-        let _result = match self.sender.try_send(cap_pack) {
+        match self.sender.try_send(cap_pack) {
             Ok(()) => return Ok(()),
             Err(TrySendError::Full(x)) | Err(TrySendError::Closed(x)) => {
-                return Err(TryEnqueueError::Full(x))
+                // cap_pack.destroy();
+                return Err(TryEnqueueError::Full(x));
```
cpacejo: (Also I must have missed this in an earlier review, sorry!)  Instead of returning a `CapPacket` message here (which is otherwise not exposed to the user of this API) let's pull out the packet from the message (i.e. `x.packet`) and return that.  (Also let's rename `x` to e.g. `cap_pack` to make it clearer to a reader what's going on here.)

adapter/ph/src/queues.rs
```diff
@@ -203,21 +203,18 @@ impl<'pktbuf> Capture<'pktbuf> {
...
-impl<'pktbuf> CapPacket<'pktbuf> {
-    // pub (crate) fn new(packet: Packet<'pktbuf>, timestamp: Instant) -> Self{
-    //     Self { packet, timestamp }
-    // }
-
-    // pub fn get_packet(&self) -> Packet {
-    //     self.packet
-    // }
-}
+// impl<'pktbuf> CapPacket<'pktbuf> {
```
cpacejo: `destroy()` shouldn't be needed here.  I think it's correct to treat the `CapPacket` message as a plain structure and just let the receiver destructure it and `destroy()` the contained packet directly.

cpacejo (changes requested):
Looks good but I agree I would restructure those loops to be in terms of the packets to send, not the buffers we're copying them to.  I suspect it will simplify that code a bit.

adapter/ph/src/inbound_processor_worker.rs
owenrhysthomas: https://github.com/org-zpr/zpr-core/issues/192

adapter/ph/src/inbound_processor_worker.rs
```diff
@@ -75,3 +80,36 @@ async fn handle_packets<'pktbuf>(
...
+) {
+    let mut bufs = Vec::new();
+    let mut num_bufs = asm.buffer_stack.try_get_buffers(count, &mut bufs);
+    // now that this loop is based on packets, it could be moved into the loop
+    // below for performance
+    for pkt in pkts {
+        if num_bufs > 0 {
+            match pkt {
+                InboundProcessorMessage::Packet(pkt) => {
+                    let pkt_clone: Packet = pkt.clone_into(bufs.pop().unwrap());
```
cpacejo: Instead of tracking `num_bufs` separately, you can just rely on the return value of `pop()` here (and break from the loop early too).  i.e. `match bufs.pop() { None => break, Some(buf) => ... }`

adapter/ph/src/queues.rs
```diff
@@ -201,23 +201,14 @@ impl<'pktbuf> Capture<'pktbuf> {
...
-    ) -> Result<(), TryEnqueueError<CapPacket<'pktbuf>>> {
+    ) -> Result<(), TryEnqueueError<Packet<'pktbuf>>> {
         let cap_pack: CapPacket = CapPacket { packet, timestamp };
-        let _result = match self.sender.try_send(cap_pack) {
+        match self.sender.try_send(cap_pack) {
             Ok(()) => return Ok(()),
-            Err(TrySendError::Full(x)) | Err(TrySendError::Closed(x)) => {
-                return Err(TryEnqueueError::Full(x))
+            Err(TrySendError::Full(cap_pack)) | Err(TrySendError::Closed(cap_pack)) => {
+                // cap_pack.destroy();
```
cpacejo: This commented code I think can go?

cpacejo (approved):
Perfect thanks!!  I think can even get rid of the `num_bufs` counter in the capture loop but otherwise looks perfect.

PR #187, opened 2024-07-12
switch to using tokio-tun to eliminate dependency on launcher

cpacejo:


adapter/ph/src/main.rs
```diff
@@ -292,9 +258,8 @@ fn main() -> ExitCode {
                 op_outq,
             ));
 
-            js.spawn(rpc_worker::launch(&*asm, &*unix_socket));
```
cpacejo, line 295: Drive-by fix: this is a duplicate of line 243/202.

owenrhysthomas (approved):
Cool! Just make sure to format before merging.

PR #188, opened 2024-07-12
Implemented capture RPC commands except query

### LINKED ISSUE #174: Create RPCs to enable packet capture
We'll want to add an RPC interface to manipulate the packet capture state.  I think 1-to-1 mapping of the state controls is fine.  So we'll have RPC functions to:
- set the capture file
- flush the capture file
- close the capture file
- enable/disable inbound capture
- enable/disable outbound capture
- query capture state (this can just return as a response both the capture file path, and the capture enable flags)

I think also, "close the capture file" should additionally disable both inbound and outbound capture.  And, the "enable capture" methods should refuse to set the flags if there is no capture file set.  (Else we're wasting CPU cycles doing nothing.)  I wouldn't worry about race conditions involving the flags and capture file for now, though it might warrant a TODO in the code.  (That is, it's possible for a "close capture file" to race "enable capture flag" in a way that leaves the capture file closed and the capture flag set.  Solving this is way more trouble than it's worth.)

Blocked on:
- org-zpr/zpr-core#165
- org-zpr/zpr-core#170

owenrhysthomas:


owenrhysthomas:
I'm not sure how to implement the query capture state command other than by adding another element to the `InnerCap` struct because I'm not sure how to get the path otherwise. We never store the path anywhere and I can't find a way to get the path from a `Savefile` using methods from `pcap::Savefile` or `pcap::Capture`

owenrhysthomas:
Still needs more testing before merge

adapter/ph/src/rpc_worker.rs
```diff
@@ -404,3 +448,79 @@ fn values_from_hist(hist_name: &str, units: &str, hist: Histogram<u64>) -> Strin
 
     values
 }
+
+async fn set_capture(asm: &Assembly<'_>, path_str: &str) -> String {
+    let path = Path::new(path_str);
+    asm.capture_worker.open_capture_file(path).await;
+    let mut message: String = "".to_string();
```
cpacejo: `String::new()` is the canonical way to do this.

cpacejo (approved):
Perfect thanks!

Thinking now let's skip the query part for the time being.  I have a future task lined up to have the RPC just pass a file descriptor, not a filename, in order to avoid privilege escalation.  In which case we won't be able to get at the filename anyway (I think).

Issue #190, opened 2024-07-12
TUN packet info


cpacejo:
Let's capture TUN packet info.  This tells us whether the packet was truncated, and what type the packet is.  If it is truncated, and if it's not IPv4 or 6, we should drop it.

Also, let's generate it on the way out.

PR #193, opened 2024-07-12
add and act on TUN packet info

cpacejo:
Read & write packet info from/to TUN device.

Also added an IP version field to the packet metadata since it isn't otherwise readily discernible.

Drive-by modification of `IpAddress` to enforce that IPv4 addresses are always correctly translated into IPv6 addresses.  (i.e. no possibility to stomp an already-written IPv6 address resulting in a mishmash.)

Addresses #190.

### DIFF
--- a/adapter/ph/src/ext/tokio_tun.rs
+++ b/adapter/ph/src/ext/tokio_tun.rs
@@ -40,10 +40,11 @@ impl TunExt for Tun {
 pub mod tun_pi {
     use bytes::buf;
 
+    // per-packet packet info
     #[derive(Clone, Copy)]
     pub struct TunPi {
-        pub strip: bool,
-        pub proto: u16,
+        pub strip: bool,  // the inbound packet was truncated (ignored outbound)
+        pub proto: u16,  // Ethertype of packet
     }
 
     #[cfg(target_os = "linux")]
@@ -89,6 +90,8 @@ pub mod tun_pi {
         }
         .into()
     }
+
+    pub const PI_SIZE: usize = std::mem::size_of::<os::TunPi>();
 
     pub fn write_pi<B: buf::BufMut>(buf: &mut B, pi: TunPi) {
         let os_pi: os::TunPi = pi.into();
--- a/adapter/ph/src/ext/tokio_tun.rs
+++ b/adapter/ph/src/ext/tokio_tun.rs
@@ -36,3 +36,68 @@ impl TunExt for Tun {
         Ok(size)
     }
 }
+
+pub mod tun_pi {
+    use bytes::buf;
+
+    #[derive(Clone, Copy)]
+    pub struct TunPi {
+        pub strip: bool,
+        pub proto: u16,
+    }
+
+    #[cfg(target_os = "linux")]
+    mod os {
+        const TUN_PKT_STRIP: u16 = 0x0001;
+
+        #[repr(C)]
+        #[derive(Clone, Copy)]
+        pub struct TunPi {
+            flags: u16,
+            proto: [u8; 2],
+        }
+
+        impl From<TunPi> for super::TunPi {
+            fn from(pi: TunPi) -> super::TunPi {
+                super::TunPi {
+                    strip: pi.flags & TUN_PKT_STRIP != 0,
+                    proto: u16::from_be_bytes(pi.proto),
+                }
+            }
+        }
+
+        impl From<super::TunPi> for TunPi {
+            fn from(pi: super::TunPi) -> TunPi {
+                TunPi {
+                    flags: 0,
+                    proto: pi.proto.to_be_bytes(),
+                }
+            }
+        }
+    }
+
+    pub fn read_pi<B: buf::Buf>(buf: &mut B) -> TunPi {
+        let mut os_pi = std::mem::MaybeUninit::<os::TunPi>::uninit();
+        let slice = os_pi.as_mut_ptr();
+        buf.copy_to_slice(unsafe {
+            /* SAFETY: we immediately initialize */
+            std::slice::from_raw_parts_mut(slice as *mut u8, std::mem::size_of::<os::TunPi>())
+        });
+        unsafe {
+            /* SAFETY: was just initialized */
+            os_pi.assume_init()
+        }
+        .into()
+    }
+
+    pub fn write_pi<B: buf::BufMut>(buf: &mut B, pi: TunPi) {
+        let os_pi: os::TunPi = pi.into();
+        buf.put(unsafe {
+            /* SAFETY: we are reading exactly the structure */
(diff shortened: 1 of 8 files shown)


adapter/ph/src/inbound_send_worker.rs
```diff
@@ -19,23 +29,30 @@ async fn worker<'pktbuf>(
...
+        for msg in &mut messages {
             match msg {
-                InboundSendMessage::Packet(msg) => {
-                    tun.send_vectored(&[IoSlice::new(msg.body())])
+                InboundSendMessage::Packet(pkt) => {
+                    tun_pi::write_pi(
+                        pkt,
+                        tun_pi::TunPi {
+                            strip: false,
+                            proto: ip_ethertype(pkt.metadata().get_ip_version()),
```
mkolehmainen: Isn't the IP version just sitting in the first four bits of the IP packet?

adapter/ph/src/main.rs
```diff
@@ -215,7 +216,6 @@ fn main() -> ExitCode {
 
             let tun_devs = TunBuilder::new()
                 .name(cmd_line.tun_if.unwrap_or(String::new()).as_str())
-                .packet_info(false)
```
mkolehmainen, line 218: If packet info is set false, then the kernel will look at the packet for version, right? So why then do we need to be sending version in the write calls?

adapter/ph/src/net_defs.rs
```diff
@@ -0,0 +1,4 @@
+// Standard network constants.
+
+pub const ETHERTYPE_IP: u16 = 0x0800;
+pub const ETHERTYPE_IPV6: u16 = 0x86dd;
```
mkolehmainen, line 4: Why are there ETHERTYPE values here.  The tun doesn't care or use this, right?  Wouldn't it make more sense for us to be using IP_TYPE ?  Or is this a generality so that we can use tap or some sort of raw socket later?

adapter/ph/src/ext/tokio_tun.rs
```diff
@@ -36,3 +36,68 @@ impl TunExt for Tun {
         Ok(size)
     }
 }
+
+pub mod tun_pi {
+    use bytes::buf;
+
+    #[derive(Clone, Copy)]
+    pub struct TunPi {
```
mkolehmainen, line 45: What does "Pi" mean?  At first I thought it was "packet interface", but now maybe I think it is "packet info".

adapter/ph/src/outbound_recv_worker.rs
```diff
@@ -25,7 +31,18 @@ async fn worker(config: &Config, asm: &Assembly<'_>, tun: &Tun) {
         // since neither `read_buf()` nor `enqueue()` support it
         for buf in bufs.drain(..) {
             let mut pkt = Packet::new(buf, OUTBOUND_PACKET_HEADROOM);
+
             tun_recv_buf(tun, &mut pkt).await.unwrap();
+
+            let pi = tun_pi::read_pi(&mut pkt);
+            if pi.strip || is_ip(pi) {
```
mkolehmainen: Should this be `!is_ip(pi)`.  (Haven't gotten to what the strip bool is yet -- maybe a comment in that tokio_tun file would help?)

mkolehmainen (changes requested):
Looks good.  I think I spotted one bug?  Made some comments.  Still not sure what the STRIP flag is.

adapter/ph/src/ext/tokio_tun.rs
cpacejo: Yes "Packet Info".  I don't love the acronym but I'm matching what's used in the Linux TUN API.

adapter/ph/src/inbound_send_worker.rs
cpacejo: That's a good point actually, why I shouldn't write code when I'm sleepy :)  You're right, no need to store it separately.

adapter/ph/src/main.rs
cpacejo: Yes, but we want it on the way in, to distinguish IPv4/6 from random other packets someone might throw at us (which might happen to have a 4 or 6 in the first 4 bits).  Also to catch truncated packets.  And unfortunately the flag applies to both directions.  (And e.g. leaving the proto set at 0 doesn't cause it to be filled in.)  [reference](https://github.com/torvalds/linux/blob/master/drivers/net/tun.c#L1874-L1888)

adapter/ph/src/net_defs.rs
cpacejo: They are what the TUN interface uses to distinguish the kind of packet that is transiting it.  See [here](https://github.com/torvalds/linux/blob/master/drivers/net/tun.c#L1874-L1888) for the default automatic behavior.

mkolehmainen: Ah, I thought it was just using the IP type values.

adapter/ph/src/outbound_recv_worker.rs
mkolehmainen: Oh I see now, TUN_PKT_STRIP.  Indicates that the packet was truncated.

cpacejo: Oooooooops!!!  I was playing around with this while testing and forgot to set it back....... if only we had an [end-to-end test](https://github.com/org-zpr/zpr-core/pull/198) (which does catch this, yay!)

adapter/ph/src/inbound_send_worker.rs
cpacejo: I think my thinking was keeping all the packet classification results separate from the packet itself.  But, I think that's not worth it for this case; we can just assume an IP packet in the inbound send worker.

adapter/ph/src/inbound_send_worker.rs
```diff
@@ -19,23 +34,37 @@ async fn worker<'pktbuf>(
...
 
     while let count @ 1.. = queue.recv_many(&mut messages, config.batch_size).await {
-        for msg in &messages {
+        for msg in &mut messages {
             match msg {
-                InboundSendMessage::Packet(msg) => {
-                    tun.send_vectored(&[IoSlice::new(msg.body())])
+                InboundSendMessage::Packet(pkt) => {
+                    let proto = ip_ethertype(ip_version(pkt.body()));
+                    let mut hdr = pkt.alloc_zeroed_header::<[u8; tun_pi::PI_SIZE]>() as &mut [u8];
```
cpacejo, line 41: Also, this was going at the wrong end before.  Correct now and working (per the not-yet-checked-in end-to-end test.)

mkolehmainen (approved):
sweet!

Issue #191, opened 2024-07-12
TUN carrier state


cpacejo:
Looks like there is an ioctl to set the TUN carrier state.  We should set this appropriately based on the state of the ZPR tether.

PR #194, opened 2024-07-13
add means to control TUN carrier flag

cpacejo:
Addresses #191.  Reflects the state of the ZPR tether on the TUN device itself, so the kernel can respond correctly to a down link.


adapter/ph/src/ext/tokio_tun.rs
```diff
@@ -2,39 +2,51 @@
...
-        // SAFETY: we've now initialized this much of the slize
+        // SAFETY: we've now initialized this much of the slice
         unsafe {
             buf.advance_mut(size);
         }
         Ok(size)
     }
+
+    fn set_carrier(&self, carrier: bool) -> Result<()> {
+        unsafe { tun_set_carrier(self.as_raw_fd(), &carrier.into()) }?;
```
owenrhysthomas, line 48: Perhaps leave a comment regarding the safety justification here as you do in the two functions above.

owenrhysthomas (approved):

PR #197, opened 2024-07-15
175 add packet capture calls to debug tool

### LINKED ISSUE #175: Add packet capture calls to debug tool
The debug tool should be able to issue RPCs to perform a packet capture.  Beside a 1-to-1 mapping of each of the RPC calls, it might be nice also to have a single subcommand which performs the full sequence of:
- set capture file
- enable packet capture
- sleep (or wait for Ctrl+C)
- close capture file

Blocked on:
- org-zpr/zpr-core#174

owenrhysthomas:


adapter/ph-debug/src/main.rs
```diff
@@ -118,3 +128,23 @@ fn handle_perf_sample(duration: u64, frequency: u64, port: String) -> std::io::R
...
+
+fn handle_set_capture(file_path: String, port: String) -> std::io::Result<()> {
+    let command = "SET-CAPTURE".to_string() + " " + file_path.as_str() + "\n";
+    println!("{}", command);
+    basic_call_response(command, port)?;
+
+    Ok(())
+}
+
+fn capture_sequence(file_path: String, time: u64, port: String) -> std::io::Result<()> {
```
cpacejo, line 140: I think if you use `&str` (a reference) here rather than `String` (an owned buffer) you won't have to repeatedly clone it.  (I see the pattern elsewhere, so fine to tackle in a separate review to keep things consistent.)  Also will get rid of the calls of `to_string()`.

cpacejo (approved):
Nice thanks!!  I think a lot of the `String` stuff can be cleaned up but that's fine for a separate PR.  `String` is best used only when you want to own a copy of the string which you can modify.  `&str` is better if you just need a reference to a string and you're not storing it anywhere.

adapter/ph-debug/src/main.rs
owenrhysthomas: Just made a new issue to do this (#207 )

PR #198, opened 2024-07-15
add basic end-to-end connectivity test

cpacejo:
This just configures and launches a basic PH pair and sends pings through to check connectivity.  I'll figure out how to link this up to a GitHub runner in a separate PR…

### DIFF
--- a/adapter/ph/test/basic-test.sh
+++ b/adapter/ph/test/basic-test.sh
@@ -179,10 +179,10 @@ CHILDREN=(${CHILDREN[@]} "$!")
 # Wait for connectivity
 #
 
-echo -n "Wait for TUN carriers... "
-wait_for 5 check_carrier zpr-a tun0
-wait_for 5 check_carrier zpr-b tun0
-echo "done."
+echo "Wait for TUN carrier..."
+wait_for 5 check_carrier zpr-a tun0 || { echo "FAILURE"; exit 1; }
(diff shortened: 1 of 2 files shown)


adapter/ph/test/basic-test.sh
```diff
@@ -0,0 +1,216 @@
+#!/usr/bin/bash
+set -euo pipefail
+
+PH_BIN=$(realpath "$(dirname $0)/../target/debug/ph-test")
+
+ZPR_USER=$USER
+
+A_HOST_ADDR=10.0.0.1
```
mkolehmainen, line 8: I assume that the namespace stuff mens that these addresses will never conflict with actual addressed used on the host, right?

mkolehmainen (commented):
Sucks you have to be root, but this looks nice and clean to me.

mkolehmainen (approved):
approve!

adapter/ph/test/basic-test.sh
cpacejo: Correct!

cpacejo:
> Sucks you have to be root, but this looks nice and clean to me.

Yes, unfortunately it's needed to get access to network namespaces..  I'll make a task to try to see if there's a way to give a normal user the necessary privileges (and for the test to detect that and then not try to get root).

Thanks!

PR #201, opened 2024-07-15
node starts polling loop for visa service

### LINKED ISSUE #147: node starts polling loop for visa service
Post registration with visa service, the node should set up a polling loop, monitoring the visa service for messages (visas and revocations).

A full implementation is blocked by
- org-zpr/zpr-core#146

However, this story is now really about building the visa service client component that lives in the node, and is kicked off once the node has accepted a certificate from the alleged visa service.

mkolehmainen:
This adds a visa service "handler" to the node main loop.  There is support for registering with the visa service, polling, and de-registering.  No support for the request-visa call or the authenticate-connect call.

This is implemented as a placeholder since normally this would be kicked off until such a time as the visa service adapter connects.  So for now, you can pass a flag to the node when you launch it to force it to connect to a listening visa service.  Hitting CTRL-C will stop the node, and deregister from the visa service.


node/src/config.rs
```diff
@@ -0,0 +1,194 @@
...
+
+#[derive(Debug, Clone, Deserialize)]
+struct Creds {
+    certificate: String, // this nodes signed certificate file
+    private_key: String, // this nodes private key file
+}
+
+impl Configuration {
+    pub fn get_cert_path(&self) -> PathBuf {
+        let base = Path::new(&self.base_path);
```
cpacejo, line 30: I think you can just `self.base_path.join()` (since `PathBuf` implements `Deref<Target = Path>`)

(No difference in behavior or performance, just fewer characters…)

node/src/config.rs
```diff
@@ -0,0 +1,194 @@
...
+
+    pub fn get_claim(&self, key: &str) -> Option<String> {
+        match self.claims.get(key) {
+            Some(v) => v.as_str().map(|s| String::from(s)), // Double decode since v.as_str() from Table includes quotes.
+            None => None,
+        }
+    }
+
+    pub fn get_node_addr(&self) -> IpAddr {
+        if let Some(a) = self.node_addr {
```
cpacejo: Can also be written `self.node_addr.expect("node address not set...")` (no difference, just less typing)

node/src/config.rs
```diff
@@ -0,0 +1,194 @@
...
+                Some(s) => s,
+                None => String::from("")
+            };
+            hmap.insert(k.clone(), vstr);
+        }
+        hmap
+    }
+
+    pub fn get_claim(&self, key: &str) -> Option<String> {
+        match self.claims.get(key) {
```
cpacejo: Can be written `self.claims.get(key)?.as_str().map(...)`

node/src/vs/vsconn.rs
```diff
@@ -0,0 +1,721 @@
...
+
+#[derive(Debug)]
+struct Shared {
+    state: Mutex<State>,
+}
+
+#[derive(Debug)]
+struct State {
+    service_addr: String, // visa service listen address, format "HOST:PORT"
+    claims: BTreeMap<String, String>,
```
cpacejo, line 60: It's a `BTreeMap` here but a `HashMap` in `config`.  Should these be the same type?  (And if so, which?  I'm guessing we don't need the ordering that `BTreeMap` provides so maybe `HashMap` is better?)

node/src/vs/vsconn.rs
```diff
@@ -0,0 +1,721 @@
...
+            tokio::select! {
+                _ = interval.tick() => {
+                    match self.do_poll(&vsc) {
+                        Ok((visas, revokes, more)) => {
+                            poll_errors = 0;
+                            if more {
+                                info!("VSConn::run poll signals there is more data ... ignoring for now");
+                            }
+                            for v in visas {
+                                let _ = output_tx.send(VSOutput::PushedVisa(v)).await;
```
cpacejo: Should `.unwrap()` these, no?  (i.e. if the channel is inadvertently closed I assume we'd want to crash and restart instead of dropping these requests?)

node/src/vs/vsconn.rs
```diff
@@ -0,0 +1,721 @@
...
+        }
+
+        if let Err(e) = tx_chan.send(cmd).await {
+            error!("VSConn::send_command failed: {}", e);
+            return Err(Error::new(ErrorKind::Other, "VSConn::send_command failed"));
+        }
+        Ok(())
+    }
+
+    // Blocking network call -- holds the state lock too.
```
cpacejo: Is there a way we can not hold the lock across I/O?  (Doing that means the rest of the system can get stuck on an internal mutex waiting on some external system which might be dead.)

Also, possible to make this `async`?  Calling a non-async blocking function from within an async function isn't great as it can cause other unrelated async threads to get stuck behind it.

(I'd give more specific recommendations but I can't figure out where the blocking call actually is… I'd expect it's `poll_vs()` but I don't see that blocking anywhere.)

cpacejo (approved):
Thanks for making the changes!  I have only one main concern about `do_poll()` blocking while holding a mutex and on an async thread.  Otherwise just some suggestions to save typing.

node/src/vs/vsconn.rs
mkolehmainen: This is because the the thrift generator use BTreeMap for map types and this value here in State gets cloned into a thrift object later.  So I could make config a BTreeMap too to reduce cognitive load.

node/src/vs/vsconn.rs
mkolehmainen: yes, will patch this.

PR #204, opened 2024-07-15
Added counters for capture packets written and dropped

### LINKED ISSUE #192: Add counters for failed captures
https://github.com/org-zpr/zpr-core/pull/182/files#r1676340896

owenrhysthomas:


adapter/ph/src/inbound_processor_worker.rs
```diff
@@ -114,4 +115,8 @@ fn clone_cap_packs<'pktbuf>(
         }
     }
     asm.buffer_stack.put_buffers(bufs.into_iter());
+    asm.counters[CounterType::InCapPacksDrop].set(
+        asm.counters[CounterType::InPacksRec].get_count()
+            - asm.counters[CounterType::InCapPacksWrite].get_count(),
+    );
```
cpacejo: Because these could be modified from a concurrent context this is racy and can result in incorrect counts.  (Two threads could both read the counts, and then write then, rather than read/write/read/write.)

I agree we should "batch" the counter modifications, but if we do that we have to do the batching "locally" (i.e., have two `mut` variables which count the packets received and written) and then write out the changes using `.increment()`.  (This minimizes memory contention on the counters.)

I opened a separate task to address the possibility to batch counter modifications.  For now though I'd just perform the `increment()` for dropped packets one at a time (like how you do for written packets).

cpacejo (changes requested):
Data race!  Otherwise looks good!

adapter/ph/src/inbound_processor_worker.rs
owenrhysthomas: ~~The problem is that once we drop one packet, we just break out of the loop and stop attempting to enqueue packets. I could change the way we restore the buffers to the buffer_stack and do it one by one and count there, but that wouldn't account for which packets were TestPackets, unless we also count the number of TestPackets and subtract those. I could always just remove the break; and continue trying to enqueue even if the first one fails.~~ Or have a separate loop at the bottom for the packets that didn't get enqueued, that's probably the best way to do it. Typing all this out helped, so I'll leave it here for my own reference, debug duck moment.

owenrhysthomas:
I changed the way failed captures are counted to loop through however many packets were not dropped and increment that way. It is also possible to increment the packets whenever a capture clone does not happen i.e. where the break statements are in the code. The breaks could be replaced increments. I feel like that may be a more elegant solution, although probably somewhat less efficient.

adapter/ph/src/inbound_processor_worker.rs
```diff
@@ -100,18 +102,24 @@ fn clone_cap_packs<'pktbuf>(
...
-                None => break,
+                None => break, //asm.counters[CounterType::InCapPacksDrop].increment(),
             },
             InboundProcessorMessage::TestPacket(_) => (),
         }
     }
     asm.buffer_stack.put_buffers(bufs.into_iter());
+    for _ in 0..(pkts.len() - num_enqueued) {
+        asm.counters[CounterType::InCapPacksDrop].increment()
+    }
```
cpacejo: This is correct but better would be if we added a `.increment_by(x)` method to the `Counter` type, so we can just perform a single call.

cpacejo (approved):
Looks race-free now, thanks!  I think we can add an `.increment_by()` method to avoid having to loop to increment the counters but feel free to do that as a separate PR as it's an optimization.

adapter/ph/src/inbound_processor_worker.rs
owenrhysthomas: We have this already! Silly I didn't use it. I'll make this change now.

PR #206, opened 2024-07-15
178 add capture link layer header

### LINKED ISSUE #178: Add capture link-layer header
I forgot about link-layer framing.  This is necessary to track capture direction.  We'll want to add a one-byte header to all captured packets which indicates capture direction.  We can use `Packet::clone_into_with_headroom()` and `Packet::alloc_zeroed_header()` to allocate a single `u8` at the beginning of the packet, which we set to 0 for inbound packets, and 1 for outbound packets.

The `Capture::enqueue_packet()` method should take an additional parameter, `direction`, which let's define as an enum `PacketDirection` of two values, `Inbound` and `Outbound`.  (This parameter you'll also have to add to the message type of course.)

Also, we should use [`Capture::set_datalink`](https://docs.rs/pcap/latest/pcap/struct.Capture.html#method.set_datalink) to set the datalink type to `Linktype::USER0`.  (Which we'll later actually register a new datalink type with the pcap folk.)

owenrhysthomas:


adapter/ph/src/outbound_processor_worker.rs
```diff
@@ -71,11 +71,14 @@ fn clone_cap_packs<'pktbuf>(
...
                 // Ensures there's at least one buffer
                 Some(buf) => {
-                    let pkt_clone: Packet = pkt.clone_into(buf);
-                    match asm
-                        .capture_queue
-                        .try_enqueue_packet(pkt_clone, SystemTime::now())
-                    {
+                    let mut pkt_clone: Packet = pkt.clone_into_with_headroom(buf, 1);
+                    let dir: &mut u8 = pkt_clone.alloc_zeroed_header();
+                    *dir = 1;
```
cpacejo, line 76: ~~These two lines which prepend the direction to the packet we'll want to do in the capture worker.  (And set `dir` based on the `direction` field of the `CapPacket`.)  That way it gets set both on the inbound and outbound directions.~~

~~Here it's sufficient to ensure there's enough headroom with the `clone_into_with_headroom()` call as you're doing.  Though I would specify the headroom amount using a `const` defined in `queues` (rather than a hardcoded `1`).  Call it let's say `CAPTURE_PACKET_MIN_HEADROOM`.~~

cpacejo (changes requested):

cpacejo (approved):

PR #210, opened 2024-07-16
add documentation and tests to packet module

cpacejo:
This changeset only moves stuff, documents stuff, and adds tests.  Also, I deleted two unnecessary methods from `Packet`, and changed the representation of `PacketMetadata` to fix some access alignment issue.


owenrhysthomas (commented):
Looks good!

PR #211, opened 2024-07-16
packet filtering with bpfprogram

### LINKED ISSUE #183: Add capture packet filtering
We want to add packet filtering capability to the PH, based on BPF.  This involves several steps, feel free to make separate tasks as appropriate:

- Replace the boolean flags in `flow_control` with an [`Option<BpfProgram>`](https://doc.rust-lang.org/std/sync/struct.RwLock.html) representing the capture filter.  (If `None`, no packets are to be captured – same meaning as both flow control flags being false.)  This value will need to be protected by some sort of synchronization construct.  For now, let's use a [`RwLock`](https://doc.rust-lang.org/std/sync/struct.RwLock.html), but ideally we will use an [RCU](https://en.wikipedia.org/wiki/Read-copy-update) mechanism to minimize read overhead.
- We'll want to check the filter in the inbound and outbound packet processors, _before_ copying it to the capture queue.  (This minimizes the work we do when a narrow filter is enabled.)  But since the filter should run against the captured data, and the pcap library doesn't support buffer chaining, we'll need to temporarily add the link-layer header in the inbound and outbound processors, instead of in the capture worker.  (So we copy the packet to the capture queue with the link-layer header already in place, and then remove it from the original packet, via `Packet::advance()`.)
- In order to leave room for the temporary link-layer header, we'll need to ensure in the inbound receive worker that we allocate sufficient headroom.  (1 byte is enough for now, but when we implement multi-flow capture, we'll need more.)
- [`BpfProgram::filter()`](https://docs.rs/pcap/latest/pcap/struct.BpfProgram.html#method.filter) can then be used by the inbound and outbount packet processors to check whether the packet should be captured.  If so, then we can copy the packet into the capture queue (same as we do today if the appropriate `flow_control` flag is set).  Note that the BPF filter itself will be able to check packet direction; hence we no longer need separate flags.
- We'll need to teach the RPC interface about BPF filters, so they can be set (instead of the direction flags).  We'll want to accept a string which we then compile (with optimization set) via [`Capture::compile()`](https://docs.rs/pcap/latest/pcap/struct.Capture.html#method.compile).  (Ideally we would do this in the RPC client, but the `pcap` crate doesn't support serializing `BpfProgram`s.  So we'll have to do it in the RPC worker for now.)

Problem is, this will all be awkward to test, since libpcap doesn't know about the ZPR link-layer type, so we can't use structured BPF expressions.  But we can test for now with expressions that inspect the raw link data e.g. `link[5] == 123`.

Blocked on:
- org-zpr/zpr-core#166
- org-zpr/zpr-core#175

owenrhysthomas:
…processor workers, haven't tested yet


owenrhysthomas:
This needs much more testing. I am still figuring out the format of the BpfProgram and how I have to input information in order to get the packets to filter in the desired way, but I just wanted to request the review so you could see the current state. I put it in the "In Review" bucket for now but will move it back to "In Progress" when I am actively doing more intense testing, including with wireshark.

Also it seems I may have to do some rebasing to resolve conflicts with main.

adapter/ph-debug/src/main.rs
```diff
@@ -49,11 +51,11 @@ fn main() -> std::io::Result<()> {
         "SET-CAPTURE" => handle_set_capture(args.file_path, port)?,
         "FLUSH-CAPTURE" => basic_call_response("FLUSH-CAPTURE\n".to_string(), port)?,
         "CLOSE-CAPTURE" => basic_call_response("CLOSE-CAPTURE\n".to_string(), port)?,
-        "ENABLE-IN-CAPTURE" => basic_call_response("ENABLE-IN-CAPTURE\n".to_string(), port)?,
-        "DISABLE-IN-CAPTURE" => basic_call_response("DISABLE-IN-CAPTURE\n".to_string(), port)?,
-        "ENABLE-OUT-CAPTURE" => basic_call_response("ENABLE-OUT-CAPTURE\n".to_string(), port)?,
-        "DISABLE-OUT-CAPTURE" => basic_call_response("DISABLE-OUT-CAPTURE\n".to_string(), port)?,
-        "CAPTURE-SEQUENCE" => capture_sequence(args.file_path, args.time, port)?,
+        "SET-PROGRAM" => handle_set_program(args.program, port)?,
```
cpacejo: I would call this `SET-CAPTURE-PROGRAM`.

adapter/ph/src/flow_control.rs
```diff
@@ -1,79 +1,116 @@
...
-    pub fn get_inbound(&self) -> bool {
-        self.copy_inbound_to_capture.load(Ordering::Relaxed)
+    pub async fn delete_program(&self) {
+        self.inner_control.lock().await.flow = None;
     }
 
-    pub fn set_outbound(&self, val: bool) {
-        self.copy_outbound_to_capture.store(val, Ordering::Relaxed);
+    pub async fn check_packet(&self, packet: &[u8]) -> bool {
+        let inner_flow = &mut self.inner_control.lock().await.flow;
```
cpacejo, line 32: Let's see if we can try to get `RwLock` working here, as holding the mutex while we run the filter means filtering becomes single-threaded even if we have a parallelized processing path.

adapter/ph/src/flow_control.rs
```diff
@@ -1,79 +1,116 @@
...
     }
 
-    pub fn get_outbound(&self) -> bool {
-        self.copy_outbound_to_capture.load(Ordering::Relaxed)
+    pub async fn program_exists(&self) -> bool {
+        let inner_flow = &mut self.inner_control.lock().await.flow;
+        match inner_flow {
+            Some(_) => true,
+            None => false,
+        }
```
cpacejo: You can use `inner_flow.is_some()` here.

adapter/ph/src/rpc_worker.rs
```diff
@@ -106,27 +107,15 @@ async fn handle_connection(
                     .await?;
                 buf_writer.write_all("OK\n".as_bytes()).await?
             }
-            "ENABLE-IN-CAPTURE" => {
+            "SET-PROGRAM" => {
```
cpacejo: Same here, `SET-CAPTURE-PROGRAM` and `DELETE-CAPTURE-PROGRAM`.  (Since the debug interface is more general than just packet capture.)

cpacejo (approved):
Looks good!  Just some minor comments.

I'll send you a Slack about writing filter programs to test with.

adapter/ph-debug/src/main.rs
```diff
@@ -140,3 +141,15 @@ fn capture_sequence(file_path: String, time: u64, port: &str) -> std::io::Result
 
     Ok(())
 }
+
+fn handle_set_capture_program(program: Option<String>, port: &str) -> std::io::Result<()> {
+    match program {
+        Some(program) => {
+            let command = "SET-CAPTURE-PROGRAM".to_string() + " " + program.as_str() + "\n";
```
owenrhysthomas: changed to use format!(), will be shown in next commit

owenrhysthomas:
Just merged main onto this branch, going to work on RwLock now

adapter/ph/src/flow_control.rs
owenrhysthomas: I initially used Mutex over RwLock always with the intention of going back and fixing it because RwLock was giving way more errors and I just wanted to get things working. I should've noted that in a comment.

cpacejo (approved):
OK I think it's fine to ignore RwLock for now!  Switching to `cbpf-rs` might fix all that.

PR #213, opened 2024-07-17
Remove DTLS

cpacejo:
Remove DTLS and everything needed to support it.

Split DTLS worker into inbound + outbound stages.

Comment out X.509 cert stuff for now.  We will want to refactor that to load the same information but not tie it to an `Ssl` object.

Also a drive-by enhancement to the integ test to pass env vars to the PH.

Addresses #195.


adapter/ph/src/inbound_recv_worker.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    loop {
+        // grab some buffers from the pool
+        asm.buffer_stack
+            .get_buffers(config.batch_size, &mut bufs)
+            .await;
+
+        // TODO: batch receive
+        for buf in bufs.drain(..) {
+            let mut pkt = Packet::new(buf, 0);
```
owenrhysthomas, line 24: Will have to add headroom here for link-layer header. Not relevant to this PR just wanted to remind myself.

owenrhysthomas (approved):
Exciting! Looks good to me.

adapter/ph/src/main.rs
```diff
@@ -287,28 +275,23 @@ fn main() -> ExitCode {
...
+            ));
+
+            js.spawn(outbound_send_worker::launch(
+                &outbound_send_worker::Config {
+                    batch_size: outbound_send_batch_size,
                 },
                 &*asm,
-                ssl_stream,
+                &*socket,
                 os_outq,
```
owenrhysthomas, line 295: is there a reasoning for inbound recv before outbound send? Just for my own curiosity. I would think that if we receive packets before we can send them we might end up with backed-up queues. But I suppose it's asynchronous so it probably doesn't really make a difference. I guess I'm wondering if there is a specific reason or if it's arbitrary.

cpacejo: Correct, the order in which we launch the tasks from `main` doesn't matter because they are all running as separate Tokio tasks.  So each of them proceeds concurrently.

(When these two workers lived in the same thread because the DTLS interface forced them to, I did have to spend a while making sure that exactly the problem you call out didn't happen.)

owenrhysthomas: That makes sense. Thanks!

PR #214, opened 2024-07-17
cleaned up ph-debug, function parameters from String to &str

### LINKED ISSUE #207: Change String to &str/clean up ph-debug
https://github.com/org-zpr/zpr-core/pull/197#pullrequestreview-2178630712

owenrhysthomas:
…&str and changed where \n is added to streamline match

### DIFF
--- a/adapter/ph-debug/src/main.rs
+++ b/adapter/ph-debug/src/main.rs
@@ -4,6 +4,7 @@
 // with a '-' on the command line
 
 use clap::Parser;
+use std::fmt::Write as _;
 use std::io::prelude::*;
 use std::net::Shutdown;
 use std::os::unix::net::UnixStream;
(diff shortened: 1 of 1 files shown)


adapter/ph-debug/src/main.rs
```diff
@@ -116,33 +114,32 @@ fn handle_watch(time: u64, port: String) -> std::io::Result<()> {
...
 
     Ok(())
 }
 
-fn handle_set_capture(file_path: String, port: String) -> std::io::Result<()> {
-    let command = "SET-CAPTURE".to_string() + " " + file_path.as_str() + "\n";
-    basic_call_response(command, port)?;
+fn handle_set_capture(file_path: String, port: &str) -> std::io::Result<()> {
+    let mut command: String = String::new();
+    let _ = write!(&mut command, "SET-CAPTURE {}\n", file_path);
```
cpacejo: I think you can instead use `format!()` to do the same thing without having to explicitly create a `String` and deal with error returns.  (i.e. `let command = format!("SET-CAPTURE {}\n", file_path)`)

cpacejo (approved):
Nice!  I like the use of `write!()`, I think you can do one better with `format!()` though.

Issue #184, opened 2024-07-11
Find a better BPF engine than pcap


cpacejo:
The `pcap` crate's BPF engine has several shortcomings:

- It can't operate on chained buffers.  So e.g. we're forced to write the link-layer header into the packet buffer.  This isn't an issue right now (and actually may be the most performant option) but might cause issues later.
- It can't serialize the compiled BPF program.  This means we're forced to include the compiler in the PH itself and perform the compilation there.
- It returns a boolean rather than an integer indicating the desired amount of packet to capture.
- It is part of the `pcap` crate which is rather large.

BPF is rather simple.  Let's try to find a replacement implementation which addresses the above issues.  Or potentially write our own (as a standalone library).

cpacejo:
https://github.com/polachok/bpfjit a possibility, but x86 only

cpacejo:
https://github.com/qmonnet/rbpf is for eBPF which is unnecessary, but it seems better maintained.

PR #215, opened 2024-07-17
initial implementation of simple BPF engine

cpacejo:
Addresses #184.

Needs more testing.  But the implementation closely follows the [libpcap implementation](https://github.com/the-tcpdump-group/libpcap/blob/master/bpf_filter.c).

Also needs the validator written.  Again that will follow the libpcap implementation.

The compiled code is very tight.  Check with `cargo rustc -r -- --emit=obj` and then `objdump -dCr target/release/deps/cbpf_rs-*.o`.


cbpf-rs/src/lib.rs
```diff
@@ -0,0 +1,390 @@
...
+    }
+
+    fn filter_slice(&self, packet: &[u8]) -> u32 {
+        let mut a = 0u32;
+        let mut x = 0u32;
+        let mut pc = 0;
+        let mut mem = [0u32; 16];
+
+        loop {
+            let insn = unsafe { self.program.get_unchecked(pc) };
```
cpacejo, line 161: Oh, I should add a note.  All the `unsafe` calls in this function rely on the program having been validated for their safety.  Of course the validator doesn't exist yet but that is forthcoming.  (It's also simple thankfully.)

cbpf-rs/src/lib.rs
```diff
@@ -0,0 +1,390 @@
...
+    use super::*;
+    use pcap::{Capture, Linktype};
+
+    #[test]
+    fn basic_test() {
+        let capture = Capture::dead(Linktype::ETHERNET).unwrap();
+        let prog = BpfProgram::from(capture.compile("ip src 1.2.3.4", true).unwrap());
+
+        #[rustfmt::skip]
+        let mut packet: Vec<u8> = vec![
```
cpacejo, line 377: I should comment this, but I also want to abstract this out into a simple "packet-maker" function to support easily writing more tests.  So consider this string of magic numbers temporary.

mkolehmainen (approved):
This is supremely geeky and totally awesome.  Maybe once this is more complete, and has a test suite you should make it it's own crate that people can use outside of ZPR.

PR #216, opened 2024-07-17
Broke up perf-sample into multiple functions/added helper functions

### LINKED ISSUE #177: Clean up RPC worker
`async fn perf_sample` is 238 lines long since `cargo fmt` was run on the file. There is lots of repeated code that could be broken into functions for ease of understanding and readability.

owenrhysthomas:


owenrhysthomas:
There are definitely more ways to break down the code into more functions, even some of the helper functions could be broken down, but I feel like the changes made here strike a good balance between making the code more readable while not becoming hard to follow because there are functions inside of functions inside of functions, although that's just an opinion, I would be happy to further break down the function.

Based on comments made on the ph-debug cleanup I am going to replace some instances of write! with format!

cpacejo (approved):
Nice!  I'm glad you're taking the time to refactor/clean things up as you see fit.  It's easy to get caught up in the flow of producing new code and forget about quality-of-life stuff like this, I know!

PR #217, opened 2024-07-17
new md5 files so build can notice thift source changes

### LINKED ISSUE #196: create manifest system for tracking changes to sources for auto-generated files
General idea is that we leave the auto-generated files (thrift, etc) in the source, but add a manifest to the directory where the files are.  THEN, we have our build system (ie, make) check that the sources for the generated files match the details in the manifest.

The manifest could be simply a list of SOURCE_FILENAME, MD5 values.

mkolehmainen:
A bunch of hackery to ensure that the built thrift files are based on current thrift reality.


cpacejo (approved):
Looks good!  Thanks for doing this!

PR #224, opened 2024-07-19
[cbpf-rs] add validation

cpacejo:
The validation intends to ensure that when filtering, we do not violate any of the safety assumptions used in the `filter()`. method.

I want to do fuzz testing (in lieu of unit testing this) but it's a little tricky as most fuzz testers only work with one variable-length input whereas we have two.  So that will come later.

### DIFF
--- a/cbpf-rs/src/lib.rs
+++ b/cbpf-rs/src/lib.rs
@@ -1,8 +1,10 @@
 use std::borrow::Borrow;
 use std::hint::unreachable_unchecked;
 
+/// BPF compatibility version
 pub const VERSION: i32 = 199606;
 
+/// a single BPF instruction
 #[repr(C)]
 #[derive(Copy, Clone)]
 pub struct BpfInsn {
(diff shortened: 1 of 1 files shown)


cbpf-rs/src/lib.rs
```diff
@@ -84,12 +101,7 @@ fn code_miscop(code: u16) -> u16 {
...
 
-pub enum BpfError {
-    DecodeError,
-    JumpTargetError,
-    MemRangeError,
-    PacketRangeError,
-}
+const MEMWORDS: usize = 16;
 
 struct CodeHelper<const A: u16, const B: u16, const C: u16> {}
```
mkolehmainen: It would be nice to have some documentation here about the CodeHelper construct works with the macro below.

mkolehmainen (approved):
I don't see anything wrong at a superficial layer.  Agree tests would be good.  A few more comments wouldn't hurt!

PR #232, opened 2024-07-22
Changed packet filtering to use cbpf-rs instead of pcap

### LINKED ISSUE #219: switch to using cbpf-rs
Let's switch to using our internal `cbpf-rs` crate to execute the BPF filters.  You should be able to `.into()` or `from()` to convert from the `pcap::BpfProgram` into a `cbpf_rs::BpfProgram` and keep that in the capture state.  (It should be `Send`!  If not let me know…)  Then you can call `.filter()` on the `cbpf_rs::BpfProgram`… 0 means don't capture, > 0 means capture that many bytes.  So you can set that value (capped to the actual packet length) as `caplen` in the packet header; I think pcap will correctly heed that and only write that number of bytes  from the packet.  Other than that it "should" just work??

owenrhysthomas:


owenrhysthomas:
As you can see on line 103 of capture_worker.rs I tried to change the caplen to be the value returned from cbpf-rs::BpfProgram::filter, but when I did this it caused tons of packets to be dropped. I printed out the value returned from flow_control::check_packet() and it was always 65535. I wanted to figure out where this number was coming from. I printed out the value returned from cpbf::BpfProgram::filter() and it was the same value. I also printed out the value returned from filter_slice inside of cbpf-rs, and it was also 65535. I'm not sure why that function is always returning the same value, but because it was causing so many packets to be dropped I didn't set caplen in the header to be the value returned from filter.

cpacejo:
> As you can see on line 103 of capture_worker.rs I tried to change the caplen to be the value returned from cbpf-rs::BpfProgram::filter, but when I did this it caused tons of packets to be dropped. I printed out the value returned from flow_control::check_packet() and it was always 65535. I wanted to figure out where this number was coming from. I printed out the value returned from cpbf::BpfProgram::filter() and it was the same value. I also printed out the value returned from filter_slice inside of cbpf-rs, and it was also 65535. I'm not sure why that function is always returning the same value, but because it was causing so many packets to be dropped I didn't set caplen in the header to be the value returned from filter.

Yes the compiled BPF filters, to accept "all of" a packet, return a value large enough to capture "any" packet.  libpcap appears to use 65535; `tcpdump` uses 262144 (= 256 KiB).  You'll want to use `std::cmp::min()` to limit the value returned to the packet's _actual_ length before passing it to the pcap file writer.  My guess is probably the writer isn't validating the `caplen` value and so a too-large value makes it "look like" the next few packets you write are part of the 1st packet.

cpacejo (approved):
Looks good!  See my comment in the overview how I think you should be able to get `caplen` working.

PR #233, opened 2024-07-23
220 blocked move bpf compilation out of ph

### LINKED ISSUE #220: [BLOCKED] move BPF compilation out of PH
We'll want to move BPF compilation out of the PH.  The debug tool should compile the BPF programs (still using the `pcap` crate), extract the instructions using `.get_instructions()`, interpret those as `cbpf_rs::BpfInsn`s (their references should be interchangeable), serialize them and write them over the RPC link, then deserialize them again as `cbpf_rs::BpfInsn`s in the PH, and build them into a program using `cbpf_rs::BpfProgram::validate()`.

For the serialization format, let's do what `bpf_asm` does (which is the same as `tcpdump -ddd` does, except with commas instead of newlines).  The format is simply:

```
<number of instructions>,<insn 0>,<insn 1>,<insn 2>,...
```

where each `<insn>` is:
```
<code> <jt> <jf> <k>
```
(i.e. the four fields from the `BpfInsn` struct, printed as base-10 numbers)

So for example:
```
11,48 0 0 0,21 0 8 1,48 0 0 1,21 0 6 1,48 0 0 25,116 0 0 4,21 0 3 4,32 0 0 41,21 0 1 84281096,6 0 0 262144,6 0 0 0
```
is a program.

(Beside that this is an easy format to work with, it means we can also easily write a debug tool subcommand to send raw BPF bytecode from `bpf_asm` or `tcpdump -ddd`.)

Blocked on:
- org-zpr/zpr-core#219

owenrhysthomas:


cpacejo (approved):
Perfect!!  Just the one TODO on error checking which you noted, fine to get that in a separate commit though.

I think I've figured out my Borrow issues, I'll wait for you to check this in though before making the change (so then I'll just modify your use site also -- and make sure I did actually fix it how i want).

PR #235, opened 2024-07-23
Gets direction data only as 0 or 1

### LINKED ISSUE #218: Wireshark ZDP P2P link header dissector
Start with writing a dissector for the link header (i.e. the one byte which indicates direction – "zdp-link-p2p").  The example [here](https://wiki.wireshark.org/Lua/Dissectors) shows how to make a dissector for just a few fields.  Just ignore everything about TCP and UDP (i.e. no need for reassembly, nor to automatically register the dissector with any tables).  Then you can manually register it as the header dissector for DLT_USER0 and it should show the captured direction of packets.

owenrhysthomas:


owenrhysthomas:
This dissector only identifies the direction by the number, not a word. Future improvements could include figuring out how to get wireshark to interpret that as a word instead.

image

owenrhysthomas:
~~Currently making improvements on this to make it more modular - practice for making the main ZDP one, which will be more complex.~~ Done

owenrhysthomas:
image
image

owenrhysthomas:
This currently uses tcp_port at the end because the wireshark shows TCP because I'm doing TCP iperf, but because ZPR is UDP we will eventually switch it to udp_port. I'm also not sure what port number to add it to.

tools/link_header_dissector.lua
```diff
@@ -0,0 +1,32 @@
+-- Dissector for link header to determine direction
+
+zdp_link_p2p_proto = Proto("Direction", "Inbound/Outbound Dissector")
+
+in_out = ProtoField.uint8("direction.direction", "Direction", base.DEC)
```
cpacejo: Instead of "direction." here, let's use "zdp_link_p2p.".

tools/link_header_dissector.lua
```diff
@@ -0,0 +1,32 @@
+-- Dissector for link header to determine direction
+
+zdp_link_p2p_proto = Proto("Direction", "Inbound/Outbound Dissector")
```
cpacejo: Instead of "Direction" let's call this "ZDP Link P2P", and for the description, "ZDP point-to-point message data".

tools/link_header_dissector.lua
```diff
@@ -0,0 +1,32 @@
...
+    local direction_name = "Unknown"
+
+    if direction == 0 then direction_name = "Inbound"
+    elseif direction == 1 then direction_name = "Outbound" end
+
+    return direction_name
+end
+
+local tcp_port = DissectorTable.get("tcp.port")
+tcp_port:add(59274, zdp_link_p2p_proto)
```
cpacejo: So this I don't think we need… this would be to signal to Wireshark when it would expect to see this link-layer protocol encapsulated in another protocol.  But because it's a link-layer format, which we're only producing explicitly as packet capture headers, this will never appear anywhere else.  (We'll always just need to register directly with the link-layer dissector table as DLT_USER0.)

(For the ZDP message dissector, you will want to do this!)

cpacejo (approved):
Fantastic!  I would just change the naming to reflect that this is the ZDP link-layer header, and also no need to register with the dissector tables.

PR #240, opened 2024-07-24
Created ZDP Dissector

### LINKED ISSUE #238: Wireshark ZDP payload dissector
See the packet structure [here](https://docs.google.com/document/d/1W0NDiHSKVvnVx5QytaxD00uUaoA6t9YU/edit#heading=h.3s49zyc).  That is the general ZPR packet format; we'll use the Baseline Configuration for sizes.  So in general, a packet will start with:
* ZPI (1 byte)
* Type (1 byte)
* Excess-Length (1 byte)
* Sequence Number (2 bytes)

Then, for packets of type == 0 (Transit Packet), the rest of the packet looks like:
* Stream ID (4 bytes)
* Pad (assume 8 bytes for now, this isn't decided yet)
* MAC (assume 4 bytes)
* D2D SAID (1 byte)
* compressed agent packet (N bytes)
* D2D MAC (assume 4 bytes)

For packets of type == 1 through 127 (stream-oriented management message), the rest of the packet looks like:
* Stream ID (4 bytes)
* Management Data (N bytes)
* Pad (assume 8 bytes)
* MAC (assume 4 bytes)

For packets of type == 128 through 255 (other management message), the rest of the packet looks like:
* Management Data (N bytes)
* Pad (assume 8 bytes)
* MAC (assume 4 bytes)

("N bytes" here means, the rest of the packet, except for any following fields.  Note also that, you'll have to subtract the "Excess-Length" field from the full packet length to get the actual length of the ZDP packet.  So if the packet's length is 60 and Excess-Length is 12, the packet is actually only 48 bytes long, the last 12 are garbage.)

For now – I would just decode these three general packet types, and decode each field simply as a big-endian integer.  (Except Pad, decode as a sequence of bytes.)  Just treat the agent packet and management data as an uninterpreted sequence of bytes.  Feel free to break this work into pieces as you find appropriate!

Note this won't decode what we're sending today correctly, as we're using the prototype's packet format, which is different.  Sarah's working on fixing this as we speak.

Follow on steps (we'll open separate tasks) include:
* decoding the compressed agent packet header (section 5.26) (#246)
* passing the "rest" of the compressed agent packet (i.e., IP options and payload) to be dissected by Wireshark according to their type (#244)
* decoding the various management messages… but let's just do a couple of these as examples, let's say echo request/response, bind agent address request/response… there are many but we aren't going to be using them all immediately (#245)
* decoding Type as a named value instead of a plain integer (according to table 3 in section 6.2.4.4), at least just for the management messages we know how to decode (see #242)

owenrhysthomas:


owenrhysthomas:
Edit: Protocol should probably be ZPR not ZDP

image
image

tools/zdp_dissector.lua
```diff
@@ -0,0 +1,59 @@
...
+zpi_val = ProtoField.uint8("zdp.zpi", "ZPI", base.DEC)
+zdp_type = ProtoField.uint8("zdp.type", "Type", base.DEC)
+excess_len = ProtoField.uint8("zdp.excess_len", "Excess Length", base.DEC)
+seq_num = ProtoField.uint16("zdp.seq_num", "Sequence Number", base.DEC)
+stream_id = ProtoField.uint64("zdp.streamid", "Stream ID", base.DEC)
+pad = ProtoField.bytes("zdp.pad", "Pad")
+mac_addr = ProtoField.uint64("zdp.mac", "MAC", base.DEC)
+d2d_said = ProtoField.uint8("zdp.d2d_said", "D2D SAID", base.DEC)
+agent_packet = ProtoField.bytes("zdp.agent_packet", "Agent Packet")
+d2d_mac = ProtoField.uint64("zdp.d2d_mac", "D2D MAC", base.DEC)
```
cpacejo: `uint32` (4 bytes) for the stream ID and the two MACs I think?

tools/zdp_dissector.lua
```diff
@@ -0,0 +1,59 @@
...
+    else 
+        subtree:add(management_packet, buffer(5, real_len - 17))
+        subtree:add(pad, buffer(real_len - 12, 8))
+        subtree:add(mac, buffer(real_len - 4, 4))
+    end
+
+end
+
+local tcp_port = DissectorTable.get("tcp.port")
+tcp_port:add(59274, zdp_proto)
```
cpacejo: This we'll want to be UDP, and… let's pick port 1021 for now.  (It's reserved for experimentation along with 1022.)

Let's also add to the "ip.proto" table, let's use protocol # 253 for now.  (Also reserved for experimentation, along with 254.)

And finally to the "eth.type" table, add as protocol # 0x88B5.  (Also reserved for experimentation, along with 0x88B6.)

cpacejo (approved):
This looks great!!

cpacejo:
> Edit: Protocol should probably be ZPR not ZDP

ZDP is correct!  That is what we call these packets.

tools/zdp_dissector.lua
```diff
@@ -0,0 +1,65 @@
...
+    else 
+        subtree:add(management_packet, buffer(5, real_len - 17))
+        subtree:add(pad, buffer(real_len - 12, 8))
+        subtree:add(mac, buffer(real_len - 4, 4))
+    end
+
+end
+
+local udp_port = DissectorTable.get("udp.port")
+udp_port:add(1201, zdp_proto)
```
cpacejo: typo: 1021 :)

cpacejo (approved):
Just needs a typo fixed, otherwise perfect, feel free to merge after fixing!

PR #241, opened 2024-07-24
node thrift endpoint for policy install

### LINKED ISSUE #137: node admin (THRIFT) endpoint can accept a new policy
In the early prototype, the visa service was part of a node and when an administrator wanted to install a new policy, the administrator would connect to the admin service running on the Node+VisaService and send an Install-Request.  The visa service would then use the RAFT database (each node was a member) to push out the new policy details.

The final version of the prototype split off the visa service into its own thing, so to install a new polity, the administrator would connect to the admin service running on the visa service, which would then use the Visa-Support-Service (running on one node) to send an Install-Request, the node would then use RAFT as usual.

In the Reference Implementation we think we can stop using RAFT so the question is how will we deliver a new policy to the system.  It turns out that nodes do not need to know much about the policy details, but they do need to know the current network configuration and the current topology.

One idea is that each node could maintain a connection to ETCD service and look there for its configuration details.

The idea that this story is proposing is that each node will implement a service interface (THRIFT, on some well known port) which will allow an admin service (could start off as the visa service) to send in messages related to policy updates.

Every node connected to the visa service through ZPRnet, so it is not a stretch to have the visa service tell all nodes what to do.  

A subtask here is to figure out exactly what details we need to tell a node.

- A new policy can invalidate visas that a node is holding.  The visa service should be able to figure that out and send visa-revocation messages in the usual way.
-  A new policy can change the links a node has (topology)
-  A new policy may update the network configuaration value.

mkolehmainen:
This is a rather large changeset which adds the visa-support-service API into the node code.  This re-implements some operations from the prototype that occur when a new policy is installed in the visa service: tell node about the configuration and version, and update relevant visas.  Importantly also sends the node visas for accessing the visa service and for visa service access to the visa-support-service.

This gets rid of the POLL function on the visa service API since the VSS api has a push function which is more efficient.  Instead of poll I renamed it PING and we can use it or not for connectivity checking.

There is just demo/debug code in the node right now that can fire up the VSS and also force the node to connect immediately to a non-zpr visa service.  Normally, visa service is only reachable via ZPR public address through adapter.


visaservice/thrift/vs.thrift
```diff
@@ -119,15 +120,9 @@ struct VisaHop {
...
-struct VisaRevocation {
-  1: i32 issuer_id,
-  2: i64 configuration,
-}
-
-struct PollResponse {
-  1: list<VisaHop> visas,
-  2: list<VisaRevocation> revocations,
-  3: i32 more, // >0 if there are more visas or revocations available.
+struct Pong {
```
svjaffer, line 124: Why "Pong"?

mkolehmainen: "pong" is the response to "ping".  Get it? "ping! pong!" heh.

svjaffer: Is this the response to the ping?  I kept seeing `Pong ping` and other indications that this is the ping and not the response.

svjaffer (approved):

visaservice/thrift/vs.thrift
mkolehmainen: Yes that's right.  It's unfortunate that the syntax for thrift has the return value ahead of the function name, so "Pong Ping()" reads backwards from what I would like!

PR #243, opened 2024-07-25
242 decode type as name not just number

### LINKED ISSUE #242: decode type as name, not just number
decoding Type as a named value instead of a plain integer (according to table 3 in section 6.2.4.4), at least just for the management messages we know how to decode

owenrhysthomas:


tools/zdp_dissector.lua
```diff
@@ -55,5 +59,54 @@ function zdp_proto.dissector(buffer, pinfo, tree)
 
 end
 
+function get_type_name(type)
+    local type_name = "Unknown"
+
+    if type == 0 then type_name = "Transit Packet"
```
cpacejo: I feel this might slow down packet dissection to have linear lookup like this.  Let's use a table lookup like [this](https://stackoverflow.com/a/37493047) (but with strings instead of functions of course).  And if the packet type isn't found in the table, then fall back to a short if/elseif sequence to classify for the various reserved ranges.

cpacejo (changes requested):

tools/zdp_dissector.lua
```diff
@@ -55,11 +59,66 @@ function zdp_proto.dissector(buffer, pinfo, tree)
 
 end
 
+function get_type_name(type)
+    local type_name_table =
```
cpacejo: I think by being defined inside the function body this table will get recreated on every function call (very expensive).  Is it possible to lift `type_name_table` outside the function?

(I know it's a little weird for me to be concerned about performance here especially without measurement – I'm just looking for easy wins now.  Opening a large PCAP in Wireshark can take a loooong time to dissect all the packets…)

cpacejo (approved):
Looks good but if possible let's lift the table definition outside the function.  Feel free to push after making that change!

tools/zdp_dissector.lua
owenrhysthomas: Totally makes sense, I'm on it now.

PR #247, opened 2024-07-25
cbpf: rework Borrow impl

cpacejo:
Usage should be less surprising now.  I realized my issue was actually how I was defining the iterator type argument for `validate()`, which I mistakenly thought earlier was an issue with my `Borrow` definition, and tried to fix it there.  So this fix actually fixes `validate()`, which in turn lets me put `Borrow` back to how it should be.


owenrhysthomas (approved):
Looks good to me!

PR #249, opened 2024-07-26
Change ZDP header definition to more closely resemble the spec

svjaffer:
This specific chunk of code is a bit wonky because it's just casting everything to the PerFlow base ZDP header, which is not correct.

### DIFF
--- a/adapter/ph/src/zdp.rs
+++ b/adapter/ph/src/zdp.rs
@@ -48,7 +48,7 @@ pub enum ZdpPacketType {
 
 #[derive(FromZeroes, FromBytes, AsBytes, Unaligned)]
 #[repr(packed)]
-pub struct ZdpAbbreviatedHeader {
+pub struct ZdpBaseHeader {
     pub packet_type: ZdpPacketType,
     pub excess_length: u8,
     pub sequence_number: u16,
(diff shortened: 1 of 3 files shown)


adapter/ph/src/zdp.rs
```diff
@@ -7,32 +7,64 @@ use zerocopy_derive::{AsBytes, FromBytes, FromZeroes, Unaligned};
...
+    RegisterAgentAddressRequest = 140,
+    RegisterAgentAddressResponse = 142,
+    UnregisterAgentAddressRequest = 143,
+    UnregisterAgentAddressResponse = 144,
+    Report = 145,
 }
 
 #[derive(FromZeroes, FromBytes, AsBytes, Unaligned)]
 #[repr(packed)]
 pub struct ZdpAbbreviatedHeader {
```
cpacejo: This struct name is an artifact of the Prototype.  I am happy for you to pick a better name like `ZdpCommonHeader` or such :)

adapter/ph/src/zdp.rs
```diff
@@ -7,32 +7,64 @@ use zerocopy_derive::{AsBytes, FromBytes, FromZeroes, Unaligned};
...
 
 #[derive(FromZeroes, FromBytes, AsBytes, Unaligned)]
 #[repr(packed)]
-pub struct ZdpHeader {
+pub struct ZdpPerFlowHeader {
+    pub abbreviated_header: ZdpAbbreviatedHeader,
+    pub stream_id: u32,
+}
+
+pub struct ZdpEchoHeader {
```
cpacejo, line 66: We probably want the zerocopy derives here too, no?

adapter/ph/src/zdp.rs
```diff
@@ -7,32 +7,64 @@ use zerocopy_derive::{AsBytes, FromBytes, FromZeroes, Unaligned};
...
     pub abbreviated_header: ZdpAbbreviatedHeader,
```
cpacejo: I'm torn but leaning toward _not_ including the common header in each message type.  That I think will make it easier to separate the code which performs the request/response dance (which only cares about the header and the _bytes_ of the payload) from the code which generates the message content (which only cares about the payload, and specifically the structure of it).

That said – Echo seems like Frank intended it as a bit of a special case which can be very quickly handled early in the processing chain (since he defined the response and request to be explicitly identical, counter to the structure of all other messages).  So maybe it's appropriate for Echo to include the common header here.  So feel free to ignore this comment for now.

cpacejo (approved):

cpacejo:
Just a naming nit and some missing `derive`s… and a philosophical tangent you can ignore for now :) looks good otherwise thanks!!

cpacejo (approved):

PR #250, opened 2024-07-29
244 passing the compressed payload to other wireshark dissectors

### LINKED ISSUE #244: passing the compressed payload to other wireshark dissectors
passing the "rest" of the compressed agent packet (i.e., IP options and payload) to be dissected by Wireshark according to their type

It turns out this is not completely possible because the protocol as well as next header data is compressed out of the IP header (v4 and v6 respectively), meaning as of now the agent packet cannot completely be dissected. In the future, a table can be created and the information can be filled in using information from the initial request and response messages that set up the flow, which do contain such information.

owenrhysthomas:


tools/zdp_dissector.lua
```diff
@@ -13,48 +13,87 @@ d2d_said = ProtoField.uint8("zdp.d2d_said", "D2D SAID", base.DEC)
...
+            agent_header_subtree:add(ihl, ihl_val)
+            agent_header_subtree:add(dscp, buffer(23, 1))
+            agent_header_subtree:add(frag_id, buffer(24, 2))
+            agent_header_subtree:add(frag_offset, buffer(26, 2))
+            agent_header_subtree:add(ttl, buffer(28, 1))
+            if ihl_val > 5 then
+                local options_len = ihl_val - ((ihl_val - 5) * 4)
+                agent_header_subtree:add(ip_options, buffer(29, options_len))
+                -- pass ip options to an options dissector here (I could not find an existing IP options dissector)
+                Dissector.get("tcp"):call(buffer(29 + options_len, real_len - (options_len + 33)):tvb(), pinfo, tree)
```
cpacejo: In this case I wouldn't dissect it at all – just leave as raw bytes.  (It's important to skip still though!)

cpacejo (approved):
I would just not try to dissect the IP options since there is no dissector available.  Otherwise looks good!

PR #255, opened 2024-07-29
Slightly restructure inbound processor flow

cpacejo:
I was writing this as a task and realized it was both a tangent, and also more complex to explain than just to do so I just did it!  This just splits inbound processing to have two branches, one for processing per-flow mgmt packets, the other for non-flow mgmt packets.


owenrhysthomas (approved):
All looks good.

PR #257, opened 2024-07-29
Added dissection of management packets

### LINKED ISSUE #245: decode management messages
decoding the various management messages… but let's just do a couple of these as examples, let's say echo request/response, bind agent address request/response… there are many but we aren't going to be using them all immediately

owenrhysthomas:


tools/zdp_dissector.lua
```diff
@@ -78,14 +102,98 @@ function zdp_proto.dissector(buffer, pinfo, tree)
...
+        management_subtree:add(dest_info, buffer(bytes_used, 2))
+    end
+end
+
+function handle_bind_agent_addr_response(buffer, management_subtree)
+    management_subtree:add(req_seq_num, buffer(0, 2))
+    management_subtree:add(status_code, buffer(2, 1))
+    management_subtree:add(info_len, buffer(3, 1))
+    local add_info_len = buffer(3, 1):uint()
+    management_subtree:add(status_info, buffer(4, add_info_len)) -- do we want if statement here, or just let it show up
```
cpacejo: I haven't seen `<MISSING>` before in other protocol dissections with optional fields so I think the `if` statement is the right way to go.

cpacejo (approved):
Nice!  I'm hoping within the week we'll actually have management traffic to dissect with this :)

PR #259, opened 2024-07-30
254 teach outbound processor worker about management messages

### LINKED ISSUE #254: Teach `outbound_processor_worker` about management messages
`outbound_processor_worker` knows about `Packet`s and `TestPacket`s.  We'll also need to give it "non-flow management" and "per-flow management" message types.

The "non-flow management" message type looks like a Packet, but has an additional Type argument, which is a zdp::ZdpPacketType value.  So e.g. `NonFlowMgmt(zdp::ZdpPacketType, Packet<...>)` as a new message type enum value, and a corresponding new async method `OutboundProcessor::enqueue_non_flow_mgmt(&Self, zdp::ZdpPacketType, Packet<...>)`.

The "per-flow management" message type is similar, but additionally takes a `stream_id: u32` parameter.

`outbound_processor_worker` will then need to learn to handle messages of these types.  We'll still want to perform packet capture and send same as we do for transit packets.  But instead of putting a `ZdpPerFlowHeader` with type `TransitPacket` on the front, for non-flow mgmt we'll want to put a `ZdpBaseHeader` with type as specified in the `NonFlowMgmt` message; for per-flow mgmt, we'll use a `ZdpPerFlowHeader` with type and stream ID filled in.

(Ignore sequence numbers for now!)

owenrhysthomas:


adapter/ph/src/outbound_processor_worker.rs
```diff
@@ -25,10 +25,25 @@ async fn worker<'pktbuf>(
     while let count @ 1.. = queue.recv_many(&mut pkts, config.batch_size).await {
         for pkt in pkts.drain(..) {
             match pkt {
-                OutboundProcessorMessage::Packet(pkt) => {
+                // TOOD not a very elegant solution, consider restructuring
```
cpacejo: This looks fine to me!

adapter/ph/src/queues.rs
```diff
@@ -130,6 +133,38 @@ impl<'pktbuf> OutboundProcessor<'pktbuf> {
...
+            .await
+            .unwrap();
+    }
+
+    #[allow(dead_code)]
+    pub async fn enqueue_per_flow_mgmt(
+        &self,
+        zdp_packet_type: zdp::ZdpPacketType,
+        packet: Packet<'pktbuf>,
+        stream_id: u32,
```
cpacejo: Nit: let's switch the order of these arguments, to match the order that the fields appear in the headers.  (Easier to remember that way.)

cpacejo (approved):
Just an argument ordering nit, feel free to push after changing!

PR #260, opened 2024-07-30
256 add report messages  handle in inbound processor worker

### LINKED ISSUE #256: add Report messages & handle in `inbound_processor_worker`
We'll want to define a struct in `zdp.rs` for the `Report` message type ([RFC 6.5 § 6.3.13](https://docs.google.com/document/d/1W0NDiHSKVvnVx5QytaxD00uUaoA6t9YU/edit#heading=h.356xmb2)).  Call it `ZdpReportHeader`.  It will likely be the first message type defined.  (Excepting `ZdpEchoHeader`, but that is a special case and you should ignore it!)  It will need to have the same `derive` and `repr` annotations as the other structs in that file so it can be decoded directly from packet data.  Only include the fields which are part of the `Report` message itself; don't include the ZDP header as a field.  And, of the two fields – you can only actually include the first field (Report Data Length), since the 2nd field is variable-length.

In `inbound_processor_worker`, we'll now need to process this message.  Again, this will likely be the first management-type message we're processing there, but you can follow the example of `TransitPacket`.  (Difference: this is a non-flow packet!  So it should go in the other branch of the `if`.)  To process this message, we'll want to do three things – (1) validate the Report Data Length field against the actual remaining packet data, (2) print out the report contents (i.e., the contents of the packet after the Report Data Length field), and (3) return the packet buffer to the buffer pool.

When printing the contents – let's treat that data as a UTF-8 string.  (i.e. `std::str::from_utf8()`), and print to `stderr`.  Note!  Writing to stderr (or any pipe that can block indefinitely) is actually in general a very bad idea from the middle of the data path like this, since if it gets stuck or even delays a little bit everything else gets stuck behind it.  This is just temporary for now to get a message to do something.  (Actually let's leave a comment to that effect.)

owenrhysthomas:


adapter/ph/src/inbound_processor_worker.rs
```diff
@@ -98,6 +98,20 @@ async fn handle_packet<'pktbuf>(
         pkt.advance(std::mem::size_of::<ZdpBaseHeader>());
 
         match packet_type {
+            ZdpPacketType::Report => {
+                let hdr =
+                    ZdpReportHeader::ref_from_prefix(pkt.body()).expect("too-short inbound packet");
+                // Not sure if this if statement is what you mean by "validate", or do you want an assert or panic
+                // or something similar?
+                if pkt.body().len() as u16 - 2 == hdr.report_data_length {
+                    pkt.advance(2);
```
cpacejo: Instead of `2`, use `std::mem::size_of::<ZdpReportHeader>()`.

adapter/ph/src/inbound_processor_worker.rs
```diff
@@ -98,6 +98,20 @@ async fn handle_packet<'pktbuf>(
         pkt.advance(std::mem::size_of::<ZdpBaseHeader>());
 
         match packet_type {
+            ZdpPacketType::Report => {
+                let hdr =
+                    ZdpReportHeader::ref_from_prefix(pkt.body()).expect("too-short inbound packet");
+                // Not sure if this if statement is what you mean by "validate", or do you want an assert or panic
+                // or something similar?
```
cpacejo: Specifically check that `report_data_len <= pkt.body().len()` after advancing past the ZdpReportHeader.  (Equals is too strict here – we just need to make sure we don't attempt to read past the end of the packet.)

But yes, if it doesn't validate, ignore it for now.  Actually let's also leave a TODO to revisit – Sarah & I were discussing this morning how best to handle protocol errors since we don't have a system in place yet.

adapter/ph/src/inbound_processor_worker.rs
```diff
@@ -98,6 +98,20 @@ async fn handle_packet<'pktbuf>(
...
+                if pkt.body().len() as u16 - 2 == hdr.report_data_length {
+                    pkt.advance(2);
+                    // TODO printing to stderr blocks indefinitely, this is just temporary
+                    eprintln!("{}", std::str::from_utf8(pkt.body()).unwrap());
```
cpacejo: Here use `pkt.body()[..report_data_length]`, since the report may shorter than the remainder of the packet.  (Though, I wouldn't expect these to ever differ… I'm just going based off the spec here.)

(Note, to satisfy the borrow checker, you'll need to read `report_data_length` into a local variable before calling `advance()` above.)

cpacejo (changes requested):

adapter/ph/src/inbound_processor_worker.rs
```diff
@@ -98,6 +98,24 @@ async fn handle_packet<'pktbuf>(
         pkt.advance(std::mem::size_of::<ZdpBaseHeader>());
 
         match packet_type {
+            ZdpPacketType::Report => {
+                let hdr =
+                    ZdpReportHeader::ref_from_prefix(pkt.body()).expect("too-short inbound packet");
+                // TODO handle protocol errors i.e. if the body is shorter
+                let report_data_length: usize = hdr.report_data_length.into();
+                if pkt.body().len() - std::mem::size_of::<ZdpReportHeader>() >= report_data_length {
+                    let report_data_length: usize = hdr.report_data_length.into();
```
cpacejo: No need for the 2nd `let report_data_length` here, the scope of the 1st one covers this block.

Also if you move the call to `advance` outside the `if` check it will become simpler (no need to subtract).

cpacejo (approved):
Looks good!  Just a suggestion to clean up some redundant code around the message length check.  Also don't forget to replace the `2` in the call to `.advance()` with a call to `size_of()`.  Feel free to push after making that change!

adapter/ph/src/inbound_processor_worker.rs
owenrhysthomas: Yes! I forgot to delete the one inside. I moved it out to streamline and then forgot to remove it inside :sweat_smile:

PR #261, opened 2024-07-30
258 add internal api to generate report messages

### LINKED ISSUE #258: add internal API to generate Report messages
Now let's add an internal API that the adapter can use to generate Report messages.  Let's add an async method on the Assembly directly (I think first one of those!) which accepts the string to send as its sole argument.  It then will have to grab a buffer, wrap it in a Packet, format the Report management message in there (just the part in § 6.3.13), and enqueue it with the outbound processor as a Report message.

Tips: You'll want to allocate the Packet with enough headroom for all the headers we stick on.  (256 bytes should be sufficient… feel free to define this somewhere centrally.)  You can use `.as_bytes()` to get the `str` as UTF-8 bytes, and `Packet::put()` to write these in.  (Make sure to check that the string is not too long to fit in the remaining space!)  And you can use `Packet::alloc_zeroed_header()` to allocate space on the front of the packet to write the Report header into.

And to test this all – once a PH comes up, let's call the above function to just send a message "Reporting for duty!" or such.  (It's OK to leave this test in for now, it'll go away at some point when we clean up startup.)

Blocked on:
- org-zpr/zpr-core#254
- org-zpr/zpr-core#256

owenrhysthomas:


adapter/ph/src/config.rs
```diff
@@ -2,3 +2,6 @@
 
 // Size of a packet buffer.
 pub const PACKET_BUFFER_SIZE: usize = 4096 * 3;
+
+// Size of headroom necessary for Report messages
+pub const REPORT_HEADROOM: usize = 256; // Not sure if config is best place, perhaps in zdp.rs?
```
cpacejo: I think here is the right place, but I think we can be more general with this.  Let's call it `DEFAULT_MESSAGE_HEADROOM`.

adapter/ph/src/assembly.rs
```diff
@@ -46,3 +49,29 @@ pub struct Assembly<'pktbuf> {
...
+        self.outbound_processor
+            .enqueue_non_flow_mgmt(ZdpPacketType::Report, pkt)
+            .await;
+    }
+
+    #[allow(dead_code)]
+    pub async fn send_discard(&self) {
+        let buf = self.buffer_stack.get_buffer().await;
+        // I know this likely doesn't need as much headroom as report becuase it doesn't have a unique header,
+        // but I wanted to make sure I left enough room for the ZDP header
```
cpacejo: Yes it's fine they both use the same headroom (hence my suggestion to just call it `DEFAULT_MESSAGE_HEADROOM`).  The packet buffer has plenty of space (12 KiB right now and we'll probably make it larger) so it's OK to err on the side of too much space.

adapter/ph/src/assembly.rs
```diff
@@ -46,3 +49,29 @@ pub struct Assembly<'pktbuf> {
...
 }
+
+impl<'pktbuf> Assembly<'pktbuf> {
+    pub async fn send_report(&self, to_send: &str) {
+        let buf = self.buffer_stack.get_buffer().await;
+        let mut pkt = Packet::new(buf, config::REPORT_HEADROOM);
+        let hdr = pkt.alloc_zeroed_header::<ZdpReportHeader>();
+        hdr.report_data_length = to_send.len() as u16;
+        if pkt.remaining_mut() >= to_send.len() {
+            pkt.put(to_send.as_bytes())
```
cpacejo: If the message doesn't fit, let's just exit early (so we don't send a malformed packet).  In fact you can check whether the message will fit even before getting a buffer (so you don't have to return it in the error case), by checking against `packet::PACKET_BUFFER_MAX_BODY_SIZE - config::DEFAULT_MESSAGE_HEADROOM`.

(Normally I'd say to return an error via a `Result<>` but once we figure out what Reports actually are – not just random strings – it's possible they'll just be a fixed size anyway, so simply exiting early is fine for now.)

cpacejo (changes requested):

adapter/ph/src/assembly.rs
```diff
@@ -46,3 +49,29 @@ pub struct Assembly<'pktbuf> {
 
     pub tun_ctl: TunCtl<'pktbuf>,
 }
+
+impl<'pktbuf> Assembly<'pktbuf> {
+    pub async fn send_report(&self, to_send: &str) {
+        // this condition will need to be adjusted when we have complete ZPR packets
+        // with the information at the end of the packet at well
```
cpacejo, line 56: Ah good call!

cpacejo (approved):
Looks good thanks for the changes!

PR #263, opened 2024-07-31
fix broken visaservice-go test

mkolehmainen:


visaservice/core/pkg/vservice/vsinst-thrift.go
```diff
@@ -244,13 +244,13 @@ func vsapiTrafficDescToIpTraffic(pd *vsapi.TrafficDesc) *snip.Traffic {
 //
 
 // returns API key
-func (vs *VSInst) BackDoorInstallAPIKeyForUnitTest(node_addr netip.Addr, node_name string) (string, error) {
-	return vs.BackDoorInstallAPIKeyForUnitTestExp(node_addr, node_name, time.Now().Add(5*time.Minute))
+func (vs *VSInst) BackDoorInstallAPIKeyForUnitTest(node_addr netip.Addr, node_name, vssAddr string) (string, error) {
```
owenrhysthomas, line 230: Do you still want `node_name string`? I'm not very familiar with Go syntax, I just noticed that node_name is now the only parameter to a function that appears alone in the function header.

owenrhysthomas (approved):
Look good, assuming removing `string` after that parameter was correct.

visaservice/core/pkg/vservice/vsinst-thrift.go
mkolehmainen: ah in go you can put the type at the end and use commas if the type is the same.  So foo(a, b string) means same as foo(a string, b string).

owenrhysthomas: Got it, that makes sense. I figured it was an optional type thing, I didn't realize that you could use commas if they were the same! That's cool.

PR #266, opened 2024-08-01
Added SyncReqState and send funcs on Assembly

### LINKED ISSUE #264: request/reply handler -- no retries
To support synchronous management messages we'll need a request/reply handler.  [§ 6.4](https://docs.google.com/document/d/1W0NDiHSKVvnVx5QytaxD00uUaoA6t9YU/edit#heading=h.3abhhcj) of the RFC defines how this functionality should work, but there are some open questions which need resolving so let's just implement what I've got below for now.

First we'll need some state to track outstanding synchronous requests.  Let's call this `SyncReqState<'pktbuf>`.  It should live in the `Assembly`.  It should have two members, one an inner struct (call it `SyncReqInnerState<'pktbuf>`) protected by an std `Mutex`, and the other, a [`tokio::sync::Semaphore`](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html).

The `SyncReqInnerState` for now will just have a [`Option<tokio::sync::oneshot::Sender<Packet<'pktbuf>>>`](https://docs.rs/tokio/latest/tokio/sync/oneshot/struct.Sender.html) as a member, call it `reply_channel`.  It will grow other members later once we (Sarah Frank & I) sort out how sequence numbers work.

`SyncReqState` should have a `new()` method to initialize it – initialize `reply_channel` as `None`, and the semaphore with `Semaphore::new(1)` (i.e. it has one permit).  Going forward, a node will create a new `SyncReqState` per link, but for now we'll just have the one which is a member of the `Assembly`.

The purpose of `SyncReqState` is to track outstanding synchronous requests which other parts of the PH make.  When making a request, the requestor acquires a permit from the semaphore (to make sure that at most we have 1 oustanding request) and registers a new `OneShot` channel in the `reply_channel` field, and wait on that for a response.  Then, `inbound_processor`, upon receipt of a reply, will check `reply_channel`, if there's something there, send the reply there.  That will wake up the requestor, which then releases the permit and continues on its way.

So – for the requestor, let's add a pair of async methods on the `Assembly`, call them `send_sync_non_flow_req()` and `send_sync_per_flow_req()`.  They'll have the same arguments for now as the corresponding `enqueue_*_mgmt()` methods on the outbound processor, except they should have two `ZdpPacketType` arguments (one for request type one for response type), and should return a `std::result::Result<Packet<'pktbuf>, SyncReqError>` (for non-flow) or `std::result::Result<(u32, Packet<'pktbuf>), SyncReqError>` (for per-flow; the `u32` is the returned flow ID).  (Define `SyncReqError` as an enum with two values for now, `LinkClosed` and `ProtocolError`.)  These method should do a few things:

1. Acquire a permit from the `SyncReqState`.
2. [Create](https://docs.rs/tokio/latest/tokio/sync/oneshot/fn.channel.html) a `OneShot` receiver/sender pair.
3. Acquire the mutex in `SyncReqState`.
4. Assign the `OneShot` sender to `reply_channel`.
5. Release the mutex.  (This you can do by placing steps 3 & 4 in braces, so the mutex is released when the scope exits.)
6. Enqueue a mgmt message with the outbound processor using the appropriate method.
7. `.await` the `OneShot` receiver.  If you get an error due to the channel being closed, return a `SyncReqError::LinkClosed` error.
8. Return the semaphore permit.  You can do this with `std::mem::drop()`.
9. Confirm that the reply message type corresponds to the request message type.  If they don't match, destroy the packet, return the buffer to the buffer pool, increment a counter (you'll have to make a new one, call it `BadMgmtResponse` or such), and return `ProtocolError`.
10. For the per-flow case, copy out the stream ID.
11. Remove the base message header or flow message header using `.advance()`.
12. Return the `Packet`, along with the stream ID in the per-flow case.

On the flip side of things, we'll need to teach `inbound_processor` about replies.  Here, you'll want to check if a received message is a response using `ZdpMessageType::is_response()`.  If it is, the inbound processor should grab the `SyncReqState` mutex, check `reply_channel` for a channel, if there is one, send the packet on it.  (You'll need to use `Option::take()` to move the sender out, since sending consumes the sender.  & you can drop the mutex immediately after calling `Option::take()`.)  If there's no reply channel (it's already `None`) or the send fails – drop the packet and count it.  (We'll need another new counter, call it `UnexpectedMgmtResponse` or such.)

This should do it for now!  Later we'll add the capability to timeout and retry, and to check sequence numbers.

owenrhysthomas:


adapter/ph/src/assembly.rs
```diff
@@ -48,9 +56,112 @@ pub struct Assembly<'pktbuf> {
...
 
     pub tun_ctl: TunCtl<'pktbuf>,
+
+    pub sync_req_state: SyncReqState<'pktbuf>,
+}
+
+#[allow(dead_code)]
+pub struct SyncReqState<'pktbuf> {
+    pub inner_req: Mutex<SyncReqInnerState<'pktbuf>>,
+    pub semaphore: Semaphore,
```
cpacejo: Let's make these two members, and `SyncReqInnerState` itself, all private.

adapter/ph/src/assembly.rs
```diff
@@ -48,9 +56,112 @@ pub struct Assembly<'pktbuf> {
     pub counters: EnumMap<CounterType, Counter>,
 
     pub tun_ctl: TunCtl<'pktbuf>,
+
+    pub sync_req_state: SyncReqState<'pktbuf>,
+}
+
+#[allow(dead_code)]
+pub struct SyncReqState<'pktbuf> {
+    pub inner_req: Mutex<SyncReqInnerState<'pktbuf>>,
```
cpacejo: Let's use a `std::sync::Mutex` here, not a Tokio Mutex.  (Reason: we are only ever holding the mutex briefly, never across an `.await`.  The `std` Mutex will be more performant in this case.)

adapter/ph/src/assembly.rs
```diff
@@ -48,9 +56,112 @@ pub struct Assembly<'pktbuf> {
...
+                }
+                rec_pkt.advance(std::mem::size_of::<ZdpBaseHeader>());
+                return Ok(rec_pkt);
+            }
+            Err(_) => return Err(SyncReqError::LinkClosed), // drop permit here as well? or will it automatically happen when the function gets popped off the call stack?
+        }
+    }
+
+    #[allow(dead_code)]
+    pub async fn send_sync_per_flow_req(
```
cpacejo, line 184: Possible to factor out the shared code between these two methods?  E.g. have a 3rd private method which accepts a `stream_id: Option<u32>`, which these two methods simply forward to.

adapter/ph/src/assembly.rs
```diff
@@ -48,9 +56,112 @@ pub struct Assembly<'pktbuf> {
...
 
 impl<'pktbuf> Assembly<'pktbuf> {
+    #[allow(dead_code)]
+    pub async fn send_sync_non_flow_req(
+        &self,
+        zdp_request_type: zdp::ZdpPacketType,
+        zdp_response_type: zdp::ZdpPacketType,
+        packet: Packet<'pktbuf>,
+    ) -> Result<Packet<'pktbuf>, SyncReqError> {
+        let permit: SemaphorePermit = self.sync_req_state.semaphore.acquire().await.unwrap(); // TODO error handling in case we don't get permit
```
cpacejo, line 113: I think `unwrap()` here is fine.  We should never be called if the semaphore gets closed (which is the only case in which this would fail).

adapter/ph/src/assembly.rs
```diff
@@ -48,9 +56,112 @@ pub struct Assembly<'pktbuf> {
...
+                if zdp_response_type != rec_hdr.packet_type {
+                    let ret_buf = rec_pkt.destroy();
+                    self.buffer_stack.put_buffer(ret_buf);
+                    self.counters[CounterType::BadMgmtResponse].increment();
+                    return Err(SyncReqError::ProtocolError);
+                }
+                rec_pkt.advance(std::mem::size_of::<ZdpBaseHeader>());
+                return Ok(rec_pkt);
+            }
+            Err(_) => return Err(SyncReqError::LinkClosed), // drop permit here as well? or will it automatically happen when the function gets popped off the call stack?
```
cpacejo: Correct!  The permit will be dropped as it goes out of scope.  I just had you do it early above since there's more "stuff" happening that we don't need it held for.  (Not that that really matters there as the permit is held potentially for hundreds of milliseconds waiting for a response; the few extra non-blocking instructions we do after that don't make a difference.  But writing it explicitly in that case means we don't even have think about whether the following instructions block or not.)

adapter/ph/src/inbound_processor_worker.rs
```diff
@@ -65,7 +65,23 @@ async fn handle_packet<'pktbuf>(
 
     let base_hdr = ZdpBaseHeader::ref_from_prefix(pkt.body()).expect("too-short inbound packet");
 
-    if base_hdr.packet_type.is_per_flow() {
+    if base_hdr.packet_type.is_response() {
+        let channel;
```
cpacejo: I didn't know you could do this!!!!! 🤯  (In OCaml you can't and in F# it's awkward.)

adapter/ph/src/inbound_processor_worker.rs
```diff
@@ -65,7 +65,23 @@ async fn handle_packet<'pktbuf>(
...
+        {
+            let mut inner_req = asm.sync_req_state.inner_req.lock().await;
+            channel = inner_req.reply_channel.take();
+        }
+        match channel {
+            Some(channel) => {
+                let _ = channel.send(pkt); // TODO error handling
```
cpacejo: In the error case here you can also just destroy the packet (which is returned in the `Err()`), return it to the buffer and count it as `UnexpectedMgmtResponse`; same as you do in the `None` case.

adapter/ph/src/assembly.rs
cpacejo, line 113: Also!  There is a "fun" bug here :)  (One which will be easiest to address after modifying this code to handle retries.)

In general, an async function has to always be prepared to be "cancelled" at any `.await` (termed a "cancellation point").  This would occur if the caller _started_ waiting on the future we return, but doesn't _finish_ waiting.  (This can happen e.g. by the caller passing the our future to a `select!()`, but something else finishes first, and then the caller just drops our future.)  We'd stop executing at the `.await` as if we returned there instead.

So specifically here, this leaves the issue that, we have a `Packet` which we then… don't do anything with, and it gets dropped on the floor.  Resource leak!

So we have to have a mechanism to destroy the packet and return it to the buffer pool automatically if it goes out of scope.  I wrote something to do that here: https://github.com/org-zpr/zpr-core/pull/271

Again – no need to address now because when you add retry capability, where the packet comes from will change.  (And it's possible we won't have this problem then but I'll have to see the code to be sure!)

cpacejo (approved):
Looks great!  Just a request to factor out the common code between the two methods.  Feel free to open a separate PR for that though.  And a couple nits on mutex type and visibility – feel free to push after addressing those.

adapter/ph/src/inbound_processor_worker.rs
owenrhysthomas: Do you mean declaring the variable without assignment?

adapter/ph/src/assembly.rs
owenrhysthomas: Yes! I was planning on doing this in the future, so good to get it out of the way now!

adapter/ph/src/inbound_processor_worker.rs
cpacejo: Yes!  I think I've wanted to do this a couple times and worked around it 😅

adapter/ph/src/assembly.rs
cpacejo: I'm OK if it's in a separate PR :)

adapter/ph/src/inbound_processor_worker.rs
owenrhysthomas: That's interesting, I didn't know it wasn't allowed in OCaml. I originally initiated it to `None`, but then I got a warning from the Rust compiler saying that the value was never read because I re-assign the value immediately, which I thought was very cool. And then it doesn't need to be `mut`.

adapter/ph/src/assembly.rs
owenrhysthomas: I just finished it, so I'm just going to put it in this PR I think, if that's okay.

adapter/ph/src/assembly.rs
```diff
@@ -89,23 +90,47 @@ pub enum SyncReqError {
 }
 
 impl<'pktbuf> Assembly<'pktbuf> {
-    #[allow(dead_code)]
-    pub async fn send_sync_non_flow_req(
+    pub fn get_sender(&self) -> Option<Sender<Packet<'pktbuf>>> {
+        match self
```
cpacejo, line 93: The `match` here is redundant, you can just simply write `self.sync_req_state......take()` as the function body.

adapter/ph/src/assembly.rs
```diff
@@ -89,23 +90,47 @@ pub enum SyncReqError {
 }
 
 impl<'pktbuf> Assembly<'pktbuf> {
-    #[allow(dead_code)]
-    pub async fn send_sync_non_flow_req(
+    pub fn get_sender(&self) -> Option<Sender<Packet<'pktbuf>>> {
```
cpacejo, line 92: Because `Assembly` has a large scope let's make this a method of `SyncReqState` instead.

adapter/ph/src/assembly.rs
```diff
@@ -118,10 +143,41 @@ impl<'pktbuf> Assembly<'pktbuf> {
...
+                        let per_flow_hdr = ZdpPerFlowHeader::ref_from_prefix(rec_pkt.body())
+                            .expect("too-short inbound packet");
+                        rec_stream_id = Some(per_flow_hdr.stream_id);
+                        rec_pkt.advance(std::mem::size_of::<ZdpPerFlowHeader>());
+                    }
+                    None => {
+                        rec_pkt.advance(std::mem::size_of::<ZdpBaseHeader>());
+                        rec_stream_id = None;
+                    }
+                }
```
cpacejo, line 157: Instead of the `match` here you could put this processing step into each of the two public functions, since it's happening after the permit has been returned.  (And so this private function just returns the packet, with the header still in place.)

cpacejo (approved):
This looks good!  Just a few nits but feel free to address in a separate PR if that's easier.

PR #270, opened 2024-08-01
outbound network errors should not terminate PH (fixes integ test)

cpacejo:
The integration test was broken because the packets we're now sending at startup were erroring out before the network came up.  This fixes that.


owenrhysthomas (approved):
Great!

PR #271, opened 2024-08-02
add DropGuard, use for get_buffer

cpacejo:
There is a problem with using packet buffers in async functions: if they are held across a cancellation point (i.e. an `.await`), this causes the async function to no longer be cancel-safe, since the packet buffer will be dropped and not returned to the buffer pool.

We can't make buffers automagically return to the buffer pool in general, because doing so requires a reference to the buffer pool, which we don't want to make global or carry around with the packet.

So I invented `DropGuard`, which lets one _temporarily_ attach a destructor to a value, and added variants of `get_buffer()` and `Packet::new()` which utilize it.  (`get_buffers()` ideally would have this also but I can't find a satisfying way to do so, and it's not needed in the contexts where we need to hold buffers across cancellation points, which is specifically slow-path code.)

### DIFF
--- a/adapter/ph/src/buffer_stack.rs
+++ b/adapter/ph/src/buffer_stack.rs
@@ -1,3 +1,4 @@
+use crate::ext::std::mem::{drop_guard, DropGuard};
 use std::sync::Mutex;
 use tokio::sync::Notify;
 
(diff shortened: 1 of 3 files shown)


adapter/ph/src/buffer_stack.rs
```diff
@@ -23,23 +24,30 @@ impl<'buf, const BUFSIZ: usize> BufferStack<'buf, BUFSIZ> {
...
         loop {
-            {
-                let mut bufs = self.buffers.lock().unwrap();
-                match bufs.pop() {
-                    Some(buf) => return buf,
-                    None => (),
-                }
+            let mut bufs = self.buffers.lock().unwrap();
+            match bufs.pop() {
+                Some(buf) => break buf,
```
cpacejo, line 32: Drive-by cleanup.

mkolehmainen (approved):
Looks great.

PR #273, opened 2024-08-02
separate PerFlowHeader from BaseHeader

cpacejo:
This stems from a discussion with Sarah yesterday.  Basically the code becomes a bit cleaner if we treat the BaseHeader and PerFlowHeader as two separate things, which come after one another in a packet, rather than the PerFlowHeader "containing" the BaseHeader.


owenrhysthomas (approved):
This looks good, and will make the shared helper function between `send_sync_per_flow_req` and `send_sync_non_flow_req` more elegant as well. We'll just have to make sure either this PR or that PR (#266) gets updated depending on the order in which we commit them. I assume this one will get committed first, so I'll update that one once it does.

cpacejo:
Thanks!  OK yes I'll go ahead and push this now, since I think it'll still get in your way as you're working on the message retries anyway.

PR #274, opened 2024-08-02
add zerocopy-bytes crossover methods

cpacejo:
These methods let us read/write `zerocopy` structs to/from `bytes` buffers.  This is handy because `Packet` is a `bytes` buffer.  Particularly this is useful for simply reading short headers off the front of a packet.  (Rather than having to reference the header, copy out its fields, then advance the data pointer.)


owenrhysthomas (approved):
This looks good. We'll also want to use the read_from_buf in the new functions in the Assembly. I'm happy to do that next week.

PR #277, opened 2024-08-02
attempt to lib-ify PH

mkolehmainen:
My attempt to make ph also usable as a library.  Just created a lib.rs file based on main.rs.  Made everything pub, and also had to put some dead_code warning suppressors in there.

Good? Bad?


adapter/ph/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "ph-test"
+name = "ph"
```
cpacejo, line 2: This breaks the integration test in `test/basic-test.sh` (which calls the binary `ph-test`).

cpacejo (approved):
This is fine for now but I'm in the process of rearranging things in the PH so we can use it for a node, so who knows what the PH will look like after that.

cpacejo:
My guess the reason for the warnings is that something can't be both a library and an application.  You'd probably have to break `main` out into a separate package.

adapter/ph/Cargo.toml
mkolehmainen: good catch

.github/workflows/rust-adapter.yml
```diff
@@ -55,12 +55,15 @@ jobs:
...
+    - name: Check warnings 
+      run: sudo apt update && sudo apt install libpcap-dev && cd adapter/ph && cargo rustc --lib -- -D warnings && cargo rustc --bin ph -- -D warnings
 
-  ph-test-warnings:
-      runs-on: ubuntu-latest
-      steps: 
-      - uses: actions/checkout@v4
-      - name: Check test warnings
-        run: sudo apt update && sudo apt install libpcap-dev && cd adapter/ph && cargo rustc --tests -- -D warnings
+# *OFF* becase ph has multiple targets (lib & binary) the --tests function is ubable to pass
```
mkolehmainen, line 61: @owenrhysthomas - I commented this out for now since from what I can tell cargo cannot support --tests and extra args to rustc when there are multiple targets.

owenrhysthomas: Ok! Sounds good.

owenrhysthomas: @mkolehmainen see #280, the warnings for tests are now back, but only on the binary ph, not the library version.

PR #278, opened 2024-08-02
Changed return value from send req helper

owenrhysthomas:


owenrhysthomas:
I didn't realize if I made a new PR from the same branch all the previous commits would still show up, the only relevant ones are the last two.

cpacejo (approved):
Looks good!

PR #280, opened 2024-08-05
Commented out tests that don't currently pass because of ph lib-ifica…

owenrhysthomas:
…tion


owenrhysthomas:
The warnings test no longer works because there are multiple targets because of the lib-ification of the PH. Mathias commented out the test from the runners (https://github.com/org-zpr/zpr-core/pull/277#discussion_r1702314166) I figured it should also be removed from the pre-commit hooks for now.

svjaffer (approved):

PR #282, opened 2024-08-05
factor out capture stuff

cpacejo:
In preparation for the fast-path refactor, this factors out the packet capture code from the inbound and outbound processors into a single function in a new module `fastpath`.

I've made a couple modifications as part of this:
* added "packets filtered" counters – these are packets we could have captured but chose not to due to the filter
* added an explicit ZDP "link-layer" header structure (vs. just writing a raw u8 value)
* optimize the batch processing by exiting early when we're out of buffers/queue space
* switch to RwLock to guard the filter state, so the accessor methods can be non-async
* remove the unused `direction` argument from the capture messages

There is one new TODO I introduced; I wasn't able to figure out an efficient way to _use_ the batch functionality, so that is a minor regression.  I will try to revisit that after finishing the fastpath reorg.


adapter/ph/src/rpc_worker.rs
```diff
@@ -382,15 +382,15 @@ async fn set_capture_program(asm: &Assembly<'_>, str_message: String) -> String
 
     let mut return_message = format!("Program: {program} set\n");
     match cbpf_rs::BpfProgram::validate(&insn_vec) {
-        Ok(final_program) => asm.flow_control.set_program(final_program).await,
+        Ok(final_program) => asm.flow_control.set_program(final_program),
```
owenrhysthomas, line 385: With the removal of this `await`, I think this function no longer needs to be `async`.

adapter/ph/src/rpc_worker.rs
```diff
@@ -382,15 +382,15 @@ async fn set_capture_program(asm: &Assembly<'_>, str_message: String) -> String
...
         _ => return_message = format!("Invalid program recieved, program not set\n"),
     }
 
     return_message
 }
 
 async fn delete_capture_program(asm: &Assembly<'_>) -> String {
-    asm.flow_control.delete_program().await;
+    asm.flow_control.delete_program();
```
owenrhysthomas, line 393: No longer needs to be `async` either, I believe.

owenrhysthomas (approved):
Nice!

adapter/ph/src/rpc_worker.rs
cpacejo: Oh!  Good catches!

PR #284, opened 2024-08-05
move ext module into its own crate

cpacejo:
Easy thing to factor out into its own crate.  This way other crates of ours can share this stuff.

The only changes I made were to add feature flags for each library that is "extended" by `ext`.  So if another of our crates uses this, it won't get dependencies it doesn't care about.


mkolehmainen (approved):
Looks good to me.

Should there be a little readme that lists the features available?

It also reads funny that the zerocopy.rs file appears to reference the bytes feature when there is also a zerocopy feature.

cpacejo:
I'll add some documentation before pushing!

That specific functionality in the `zerocopy` crate is "crossover" functionality which links `zerocopy` stuff with `bytes` stuff.  So it's only enabled if _both_ those features are requested (i.e., the user is using both `zerocopy` and `bytes` and would like them both extended).

(In theory, there could be other, non-`bytes`-dependent, things in the `zerocopy` crate.)

PR #285, opened 2024-08-06
Factor out ZPI encap/decap and encryption into fastpath module

cpacejo:
Also add some new counters, per the ZPR spec.  Slightly simplified `name_counters()`.

Encryption and MAC are TODOs.

Switched some argument orders to follow the order most general → most specific.


adapter/ph/src/inbound_processor_worker.rs
```diff
@@ -57,10 +56,26 @@ where
...
+        }
+    }
+
+    fastpath::maybe_capture(asm, Direction::Inbound, &mut pkt);
+
+    let _zpi = match fastpath::decap_zpi(asm, 0, &mut pkt) {
+        Ok(zpi) => zpi,
+        Err(err) => {
+            fastpath::drop_and_count(asm, pkt, err);
+            return;
```
owenrhysthomas, line 76: There are lots of other places we can now use this `drop_and_count` function, especially in a bunch of the code I've written recently. (ex. line 92-94 & 98-100 in this file) I can open another PR and do that. (#287)

owenrhysthomas (approved):
Looks good!

adapter/ph/src/inbound_processor_worker.rs
cpacejo: Oh, that would be great!!

PR #286, opened 2024-08-06
support odd length checksums

cpacejo:
To support ZPI 0… also let's support 0-length checksums and checksums up to the actual maximum length the algorithm supports.


mkolehmainen:
I thought ZPI=0 uses two byte checksum.

cpacejo:
> I thought ZPI=0 uses two byte checksum.

By "odd-length" I mean the body of the thing being checksummed can be odd.  The checksum itself is the 2-byte internet checksum.

mkolehmainen (approved):
Looks good to me!

PR #289, opened 2024-08-06
Request/reply handler - implement retries

### LINKED ISSUE #267: request/reply handler -- retries
Now let's teach the request/reply handler how to retry sending requests.  It will need to do this if it does not receive a response.  Specifically the schedule suggested in the RFC is: 3 retries (beyond the initial attempt), spaced 1 second apart, before returning a timeout error.  (Let's specify these as constants in `config.rs`.)

The first tricky bit here is, we need to be able to recreate the packet if we need to resend it.  We could do this by making a copy of the packet body in a local buffer, but there's actually another problem which would be good to solve.  Specifically, because we allow only one outstanding request at a time, if a lot of callers try to send management messages at once, they'll all have written their messages into packet buffers and be hanging on to them for potentially several seconds, possibly starving the rest of the system of these buffers.

So we can solve both these problems this way.  Instead of taking a `pkt: Packet<'pktbuf>` argument, the sync-msg-send methods should take a `pkt_fn: impl Fn<&mut Packet<'_>> + Send + 'static` argument.  This function is a "message-construction" function: the caller should supply a function which builds the message in the given `Packet`.  Now, it will be the responsibility of the sync-msg-send methods to allocate the packet buffer, they can do so only when they have acquired the semaphore permit, and they can do so repeatedly if they need to re-send the request.  So you can make that change first: in the sync-msg-send methods, after acquiring the semaphore permit, grab a buffer from the buffer pool, create a `Packet` from it, and call `pkt_fn()` on a &mut reference to that packet.  Then proceed to send the packet as before.

(The `+ Send + 'static` I don't think is necessary here.  But it gives us space in the future to make the implementation of these functions pass the function to a worker, which would require these bounds.)

Now to actually implement the retries.  Instead of simply `.await`ing on the `OneShot` receiver, you'll want to use a [`tokio::select!`](https://docs.rs/tokio/latest/tokio/macro.select.html) in a loop.  The `select!` will have two branches, one for the receiver, and one for a [`tokio::time::sleep()`](https://docs.rs/tokio/latest/tokio/time/fn.sleep.html) of one second.  If the receiver branch matches, then you can return the received packet like normal.  If the sleep branch matches, then you'll want to loop and try again, a maximum of 3 times.  (Each time, you'll need to get a new packet buffer, and call `pkt_fn()` to fill it in.)  If that limit is reached, then do the following:

1. Take the inner mutex of the `SyncReqState` struct.  Take the sender out using [`Option::take()`](https://doc.rust-lang.org/std/option/enum.Option.html#method.take) (this lets you move out of an option).
2. If the sender was `None` – that means the inbound processor raced us and won, and there's a message waiting for us.  Drop the mutex, `.close()` the receiver (just in case somehow someone else stole the sender), read the message using [`.try_recv()`](https://docs.rs/tokio/latest/tokio/sync/oneshot/struct.Receiver.html#method.try_recv), process it and return it.  (If `.try_recv()` fails – which it only could if some other code stole the sender – then just return `Timeout` as in step 3.)
3. Otherwise (the sender was still there), we know there's no raced message (since `.send()` consumes the sender).  No need to explicitly close the receiver or check for a message on it.  Return a `Timeout` error to the caller.  (You'll need to add this to the `SyncReqError` enum.)

Ideally, we have some sort of unit/integration testing around this functionality.  You might be able to make a fake `Assembly` to do some basic unit testing with.  That might be a heavy enough lift though to warrant a separate task.  The easy way to test to start with though, is to increase the retry count, then you should see the `Hello` messages succeed in both directions regardless of the order you start the two PHs in.  (And you should be able to see the retries occurring over Wireshark as well…)

Blocked on:
- org-zpr/zpr-core#264

owenrhysthomas:


owenrhysthomas:
No retries yet, just closure

owenrhysthomas:
I haven't yet done the unit/integration tests, but I did verify that now the Hello request/reply now is printed by both sides when running basic-test.sh, even without increasing the number of re-requests.

adapter/ph/src/assembly.rs
```diff
@@ -98,51 +100,80 @@ impl<'pktbuf> Assembly<'pktbuf> {
...
             inner_req.reply_channel = Some(sender);
         }
 
-        // Determine if sending a non-flow or per-flow message
-        match stream_id {
-            Some(stream_id) => {
-                self.outbound_processor
-                    .enqueue_per_flow_mgmt(zdp_request_type, stream_id, packet)
-                    .await
+        for _i in 0..3 {
```
cpacejo: `0..=3` (i.e. a closed range), so we have an initial try, then three retries.

Also let's put that `3` in `config.rs`, call it `DEFAULT_REQUEST_RETRY_COUNT` or such.

adapter/ph/src/assembly.rs
```diff
@@ -98,51 +100,80 @@ impl<'pktbuf> Assembly<'pktbuf> {
...
+                        .enqueue_non_flow_mgmt(zdp_request_type, packet)
+                        .await
+                }
             }
+            tokio::select! {
+                received_val = &mut receiver => {
+                    drop(permit);
+                    return self.match_received(received_val, SyncReqError::ProtocolError, zdp_response_type);
+                }
+                _ = sleep(Duration::from_secs(1)) => ()
```
cpacejo: Same here – let's put that `Duration` in `config.rs`, call it `DEFAULT_REQUEST_RETRY_PERIOD` or such.

adapter/ph/src/assembly.rs
```diff
@@ -98,51 +100,80 @@ impl<'pktbuf> Assembly<'pktbuf> {
...
+                None => {
+                    self.outbound_processor
+                        .enqueue_non_flow_mgmt(zdp_request_type, packet)
+                        .await
+                }
             }
+            tokio::select! {
+                received_val = &mut receiver => {
+                    drop(permit);
+                    return self.match_received(received_val, SyncReqError::ProtocolError, zdp_response_type);
```
cpacejo: This should be `SyncReqError::LinkClosed` I think?  (Since an error will occur if the sender has gone away.)

adapter/ph/src/assembly.rs
```diff
@@ -98,51 +100,80 @@ impl<'pktbuf> Assembly<'pktbuf> {
...
+                }
+                _ = sleep(Duration::from_secs(1)) => ()
+            }
+        }
+        match self.sync_req_state.get_sender() {
             None => {
```
cpacejo: Looking at this now, I realize there's no need to break this into two separate cases.  Instead of `take()`ing the sender, we can just set it to `None`, and then follow the logic in this block (close the receiver, drop the permit, check for a remaining message).  Reason being – if there _was_ a sender there, after we `close()` the receiver, the receiver will read as `Closed` regardless and we'll correctly return `Timeout`.

(The advantage to simplifying is – we don't need to explain _why_ we're able to optimize away these steps in the case that the sender was still present.)

cpacejo (approved):
Nice!! Just a couple nits, feel free to push after addressing!

adapter/ph/src/assembly.rs
owenrhysthomas: I called it DEFAULT_REQUEST_RETRY_TIMER because that's what Frank uses in the RFC

PR #292, opened 2024-08-06
copy only caplen bytes during pcap

cpacejo:
This is a simple optimization I realized we could make: when cloning a packet for packet capture, we only need to copy the prefix of the packet which we're actually going to write.  Since that copy happens in the data path, this can prevent the data path from being unnecessarily bogged down if it's configured to capture only headers, but most packets are very large.

The bulk of the code is just the new functionality in `Packet` to make a clone of only part of a `Packet`.  Otherwise I just swapped the `caplen` variable out for `orig_len`, since the captured packet's length is now effectively `caplen`.


owenrhysthomas (approved):
Nice!

PR #293, opened 2024-08-06
Allowed install_hooks to be executed from any folder

### LINKED ISSUE #281: allow install hooks script to work independent of location
Currently `install_hooks.sh` only works if the person calling it is in `zpr-core/tools`. Fix this so a user can invoke the script from any directory.

owenrhysthomas:


tools/install_hooks.sh
```diff
@@ -2,6 +2,9 @@
 
 HOOKS="pre-commit"
 PATH_TO_HOOKS=../.git/hooks
+PATH_TO_SCRIPT=$(cd "$(dirname "${BASH_SOURCE[0]}")"; pwd -P)
```
svjaffer: So... I pulled this down and tried it out and got an error on this line.  You may need to take a second swing at this.

owenrhysthomas: Huh, worked on my end. I'll investigate.

svjaffer (approved):

Issue #176, opened 2024-07-09
pass FD instead of filename for packet capture


cpacejo:
Passing a path to the PH to use for packet capture presents the following security hole: if a lesser-privileged user is allowed to access the RPC endpoint (e.g. by changing its Unix permissions), that user may specify a path which only the PH user is allowed to access and thus overwrite that file.

The way to fix this is to pass a file descriptor, rather than a path, from the RPC client to the PH.  This is possible by sending ancillary data over the Unix Domain socket.  The RPC client should open the specified path using `open(O_WRONLY | O_CREAT | O_TRUNC)`, then sent as `SCM_RIGHTS` ancillary data as part of the "set capture file" request.  In the PH, a `File` to capture to can be created from this FD using [`File::from_raw_fd()`](https://docs.rs/tokio/latest/tokio/fs/struct.File.html#impl-FromRawFd-for-File).

This should replace, not augment, the existing path-based "set capture file" RPC call.

The file descriptor used in the RPC client can simply be `stdout`, after it's validated that it's not a terminal.

Blocked on:
- org-zpr/zpr-core#175

PR #294, opened 2024-08-06
replace pcap dependency with homegrown solution

cpacejo:
The major change here is adding the `pcap_writer` module, which just implements a trivially simple PCAP file writer.  The key differences from the `pcap` crate or `pcap_file` are:
1. It's asynchronous – this means if the PCAP destination is e.g. a network file system or an SSH tunnel, it won't block the Tokio runtime.
2. It's not `pcap` – meaning we now can eliminate that (and the big C library which it depends on) as a dependency from the PH.

The remainder of changes just switch to using this instead of `pcap`, which, beside type changes, mostly are (1) switching some methods to async, (2) handling some new errors, and (3) opening the `File` ourselves.

(Long-term plan is also to switch away from having a `File` as the backing store, and instead allowing an arbitrary file descriptor, which is passed in via the RPC mechanism (#176).)

### DIFF
--- a/adapter/ph/src/capture_worker.rs
+++ b/adapter/ph/src/capture_worker.rs
@@ -1,10 +1,10 @@
 use crate::assembly::Assembly;
+use crate::pcap_writer::*;
 use crate::queues::CapPacket;
 use core::future::Future;
-use libc::timeval;
-use pcap::{Capture, Dead, Error, Linktype, Packet, PacketHeader, Savefile};
+use std::io;
 use std::path::Path;
-use std::time::{Duration, UNIX_EPOCH};
+use tokio::fs::File;
 use tokio::sync::{mpsc, Mutex};
 
 pub struct CaptureWorker {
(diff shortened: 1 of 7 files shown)


adapter/ph/src/capture_worker.rs
```diff
@@ -1,46 +1,50 @@
...
+            Some(PcapWriter::open(File::open(path).await?, linktype::USER0).await?);
+        Ok(())
     }
 
-    pub async fn flush_capture_file(&self) -> Result<(), Error> {
+    pub async fn flush_capture_file(&self) -> Result<(), io::Error> {
         let sf = &mut self.inner_cap.lock().await.savefile;
         match sf {
-            Some(ref mut sf) => sf.flush(),
+            Some(ref mut sf) => sf.flush().await,
```
owenrhysthomas, line 35: do you want a `?` here too? Otherwise I don't see how this fn can return an `io::Error`

owenrhysthomas (approved):
Nice! Just the one question about the `?`

adapter/ph/src/capture_worker.rs
cpacejo: `sf.flush().await` itself returns a `Result<(), io::Error>` – so here I'm just returning directly whatever it returns.  (I could alternatively write `{ sf.flush().await?; Ok(()) }` though.)

owenrhysthomas: Oh, I see. That makes sense.

PR #295, opened 2024-08-07
recieved -> received & added handling for wrong # of args to rpc_work…

### LINKED ISSUE #186: Error handling in RPC commands
Right now if any of the RPC commands that expect multiple values, like PERF-SAMPLE or SET-CAPTURE, are not provided multiple arguments, the program simply fails with no explanation. Could send some kind of message letting them know

EDIT: Just realized this is only a problem if directly connecting to the RPC worker, if you use the debug tool it sets default values, rendering this problem mostly irrelevant.

owenrhysthomas:
…er comms

### DIFF
--- a/adapter/ph/src/main.rs
+++ b/adapter/ph/src/main.rs
@@ -202,9 +202,6 @@ fn main() -> ExitCode {
 
             let mut js = JoinSet::new();
 
-            // // Launches RPC worker program
-            // js.spawn(rpc_worker::launch(&*asm, &*unix_socket));
-
             let usr1_stream = Box::leak(Box::new(signal(SignalKind::user_defined1()).unwrap()));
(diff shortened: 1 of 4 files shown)


adapter/ph/src/main.rs
```diff
@@ -202,8 +202,8 @@ fn main() -> ExitCode {
 
             let mut js = JoinSet::new();
 
-            // Launches RPC worker program
-            js.spawn(rpc_worker::launch(&*asm, &*unix_socket));
+            // // Launches RPC worker program
+            // js.spawn(rpc_worker::launch(&*asm, &*unix_socket));
```
cpacejo: Oh, I see this was duplicated on line 250!  Feel free to just delete no need to leave as a comment.

cpacejo (approved):
Looks good!

PR #299, opened 2024-08-07
Added documentation

### LINKED ISSUE #296: Documenting!
I have not been very good about leaving comments/function headers. Here I am starting to go back through and document :)

owenrhysthomas:


adapter/ph/src/assembly.rs
```diff
@@ -100,6 +101,47 @@ pub enum SyncReqError {
 }
 
 impl<'pktbuf> Assembly<'pktbuf> {
+    // Sender functions for non-per flow request management packet.
+    // Requires the type of ZDP packet being sent as well as the type of the
+    // expected response packet.
+    // Returns the received packet or an error
```
cpacejo: I think we can be more specific – the packet has all ZDP headers removed; only the management response body remains.

adapter/ph/src/assembly.rs
```diff
@@ -100,6 +101,47 @@ pub enum SyncReqError {
...
+        pkt_fn: impl Fn(&mut Packet<'_>) + Send + 'static,
+    ) -> Result<Packet<'pktbuf>, SyncReqError> {
+        self.send_sync_req_helper(zdp_request_type, zdp_response_type, None, pkt_fn)
+            .await
+    }
+
+    // Sender functions for per flow request management packet.
+    // Requires the type of ZDP packet being sent as well as the type of the
+    // expected response packet.
+    // Returns the received packet or an error
```
cpacejo: Don't forget to mention what the `u32` is (stream ID).

adapter/ph/src/assembly.rs
```diff
@@ -100,6 +101,47 @@ pub enum SyncReqError {
 }
 
 impl<'pktbuf> Assembly<'pktbuf> {
+    // Sender functions for non-per flow request management packet.
```
cpacejo: If you use `///` instead of `//` these will show up as rustdoc documentation!  I'll drop a note in the channel how to view that.

cpacejo (approved):
Nice!  Thanks for doing this :)  Just a couple nits, feel free to push after addressing.

PR #300, opened 2024-08-07
Initial work on a key manager for setting up and maintaining an SA

### LINKED ISSUE #251: Adapter Sends key message to node, gets token reply
This is part of the EPIC story of org-zpr/zpr-core#18.

- Node is configured to listen on a "docK" UDP port.  
- Adapter is instructed to connect to the node "dock" port.
- Adapter sends a Key Management message to the node.
- Node logs the event and sends a Key Management message back.
- Adapter logs the event.

That's it.  No SA setup or anything. Just working on the pathway for the eventual key management exchange.

Once this is complete, we open up org-zpr/zpr-core#253 .

mkolehmainen:
* New km.rs module in ph/ which includes some code that is clearly temporary (eg, SillyKeyManager).
* The node binary will set up the silly key manager waiting for a message.
* The adapter (cd) binary will send the silly KM when told to connect.

A follow on story will get an actual key manager routine in here (eg, Noise).


adapter/cd/src/cd/mod.rs
```diff
@@ -7,29 +7,71 @@ pub use crate::cd::config::Config;
...
+            Err(e) => {
+                error!("cmonitor shut down with error: {}", e);
+            }
+        }
+    });
+
+    info!("waiting for cmonitor to start...");
+    let mut i = 0;
+    while i < 10 {
+        if monitor.get_command_channel().is_some() {
```
cpacejo, line 51: Could `CMonitor` instead return the channel from `new()`?

adapter/cd/src/zdp/client.rs
```diff
@@ -0,0 +1,103 @@
...
+
+#[derive(Debug, Clone)]
+pub struct ZDPClient {
+    addr: String,
+}
+
+impl ZDPClient {
+    pub fn new(addr_port: &str) -> ZDPClient {
+        ZDPClient {
+            addr: addr_port.to_string(),
```
cpacejo: nit: `.to_owned()` is a little stylistically better.  `.to_string()` depends on the `Display` trait which is meant for generating user-facing output.  Whereas `.to_owned()` is meant for exactly the purpose you are using it for here – making an owned copy of a referenced value.

(`.to_string()` is not incorrect _per se_ though as I believe for strings it is guaranteed to return the string unmodified.)

adapter/cd/src/zdp/client.rs
```diff
@@ -0,0 +1,103 @@
...
+
+use std::net::SocketAddr;
+
+use tokio::net::UdpSocket;
+
+use ph::km::{KeyManager, SillyKeyManager};
+
+#[derive(Debug, Clone)]
+pub struct ZDPClient {
+    addr: String,
```
cpacejo: Might make for easier/cleaner error handling to store a [`SocketAddr`](https://doc.rust-lang.org/std/net/enum.SocketAddr.html) here and require the caller of `new()` to parse/validate the address?

node/src/zdp/server.rs
```diff
@@ -0,0 +1,108 @@
...
+use tokio::sync::mpsc;
+use tokio_util::sync::CancellationToken;
+
+use tracing::info;
+
+use ph::km::{KeyManager, SillyKeyManager};
+
+#[derive(Debug, Clone)]
+pub struct ZDPServer {
+    addr: String, // listen address, "host:port"
```
cpacejo: Same here – maybe better to require a `SocketAddr` in `new()`?  Leave the parsing to `main()` or whomever.

cpacejo (commented):
I've reviewed everything _except_ the key manager.  There's a lot here all at once so I'm only able to give a superficial review to most of it.

adapter/ph/src/km.rs
```diff
@@ -0,0 +1,919 @@
...
+
+#[derive(Debug)]
+struct KMShared<'mgr> {
+    state: Mutex<KMState<'mgr>>,
+}
+
+struct KMState<'mgr> {
+    // Warning - have no idea what I'm doing with lifetimes.
+    // What I'm trying to assert here is that the impl of KeyManagerStateMachine must live as long as the KeyManager it is passed to.
+    statemachine: Box<dyn KeyManagerStateMachine + 'mgr>,
```
cpacejo, line 93: This is correct!

adapter/ph/src/km.rs
```diff
@@ -0,0 +1,919 @@
...
+    kmsettings: KMSettings,
+
+    sa_id: u8, // current SA identifier
```
cpacejo: I have a ph/zpr.rs now for "ZPR concepts", feel free to add an `SaId: u8` type there!  (Or I can do when I integrate.)

adapter/ph/src/km.rs
```diff
@@ -0,0 +1,919 @@
...
+
+    mgmt_tx: Option<mpsc::Sender<Bytes>>, // Internal queue for key management messages to be processed.
```
cpacejo, line 99: I wonder if we can do without the `Option` here, require this queue to be specified on creation of `KeyManager`.

Also then, I think `mgmt_tx` can live outside the mutex.  Then you don't have to perform the mutex dance every time you access it.

adapter/ph/src/km.rs
```diff
@@ -0,0 +1,919 @@
...
+
+/// Stateful key manager for ZDP.  Requires an instance of a [KeyManagerStateMachine] to do the actual work.
+#[derive(Debug, Clone)]
+pub struct KeyManager<'mgr> {
+    shared: Arc<KMShared<'mgr>>,
+}
+
+#[derive(Debug)]
+struct KMShared<'mgr> {
+    state: Mutex<KMState<'mgr>>,
```
cpacejo, line 87: Later we should be able to use `RcuBox` here to avoid having to grab a mutex for each packet.

adapter/ph/src/km.rs
```diff
@@ -0,0 +1,919 @@
...
+    /// For all packets, there must be enough padding included in the body length to
+    /// accomodate any expansion caused by encryption.
+    ///
+    /// For transit packets the padding space must be between the ZDP header and the
+    /// agent data bits.
+    ///
+    /// `message` is expected to be a ZDP message wihout a ZPI value.  We add a ZPI
+    /// value to the front of the message -- note that the value we add is just the
+    /// SA_ID.  It's up to caller to mix in the configuration ID value.
+    pub fn encrypt_transport(&self, message: &mut Packet) -> io::Result<()> {
```
cpacejo, line 154: Possible to define a custom error type?  `io::Error`s are difficult to map to anything actionable.

adapter/ph/src/km.rs
```diff
@@ -0,0 +1,919 @@
...
+                // programming error
+                panic!("SA_ID is zero");
+            }
+            encr = state.statemachine.get_transport_encryptor();
+            sa_id = state.sa_id;
+        }
+
+        // The assumption here is that caller has already built in space of any padding required by key manager protocol.
+        let encr_len: usize;
+        match base_hdr.packet_type {
```
cpacejo, line 184: You can use `.is_per_flow()` to determine whether the packet type indicates a per-flow packet.

adapter/ph/src/km.rs
```diff
@@ -0,0 +1,919 @@
...
+        match encr.encrypt_transport(sa_id, &message.body()[0..encr_len], &mut encr_buf) {
+            Ok(len) => {
+                // Copy the encrypted data back into the message -- there should be sufficient room for it since
+                // caller should know our required padding space and alignment.
+                message.body_mut()[0..len].copy_from_slice(&encr_buf[0..len]);
+
+                // Now we need to push a our SA id onto the front.
+                // (Note ZPI should be part of integrity protected data)
+                let zpi_buf = message.alloc_zeroed_headroom(1);
+                zpi_buf[0] = sa_id;
```
cpacejo, line 214: Ah here we have a conflict!  I have the fastpath adding the ZPI before encryption.  This I did because (a) the ZPI also (nominally) indicates the packet layout, (b) packet capture should therefore capture the ZPI so packets can be correctly dissected by Wireshark, and (c) packet capture necessarily must occur before encryption.

So I wonder what the right solution is here.  Maybe this method can assume the ZPI is present, and look up the SA to use from that?

(I can also sort this out after you check in & I link this into the fastpath.)

adapter/cactl/cactl/main.go
```diff
@@ -35,6 +35,14 @@ func main() {
 	}
 }
 
+func get_control_socket() string {
+	dd := os.Getenv("XDG_DATA_HOME")
```
cpacejo, line 39: Nice!

adapter/ph/src/km.rs
```diff
@@ -0,0 +1,919 @@
...
+
+        // TODO: Ability to decrypt in place. Not sure how to accomplish.  At very least we could use our own buffer pool.
+        let mut decr_buf = [0u8; config::PACKET_BUFFER_SIZE];
+
+        // TODO: pass sa_id into decrypt
+        match encr.decrypt_transport(sa_id, &message.body()[1..], &mut decr_buf) {
+            Ok(len) => {
+                // Copy the decrypted data back into the message.
+                // Note at this point the "padding" space on the message is probably filled with leftover ciphertext.
+                message.body_mut()[1..len + 1].copy_from_slice(&decr_buf[0..len]);
```
cpacejo, line 269: Same here – I think you can leave ZPI in place as we'll want it there for packet capture and (nominally) interpreting the structure layout.

adapter/ph/src/km.rs
```diff
@@ -0,0 +1,919 @@
...
+                None => {
+                    return Err(io::Error::new(
+                        io::ErrorKind::Other,
+                        "KeyManager not running",
+                    ));
+                }
+            }
+        }
+        let mut km_buf = BytesMut::with_capacity(message.len());
+        km_buf.put(message);
```
cpacejo: `Bytes::copy_from_slice(message)` is a shorter way to spell this

adapter/ph/src/km.rs
```diff
@@ -0,0 +1,919 @@
...
+                ));
+            }
+        }
+        Ok(())
+    }
+
+    /// Pass in a full Key Management payload from our peer here (should not include ZDP header).
+    ///
+    /// We copy the payload into our own buffer for processing asynchronously. Caller should free buffer.
+    pub async fn handle_km_message(&self, message: &[u8]) -> io::Result<()> {
```
cpacejo, line 285: What I've realized in the fastpath is we don't want to ever call an async function – i.e. we don't want to block on a queue.  So all the calls into mgmt-land I've turned into best-effort calls (i.e. `try_...`).  If the queue is full they just return an error, and the caller counts it as a drop due to internal congestion.

adapter/ph/src/km.rs
```diff
@@ -0,0 +1,919 @@
...
+/// ```text
+///      *
+///      ↓
+/// Configuring -> Transport
+///      ↓ ↑          ↓
+///     Error <-------+
+///```
+///
+/// Not accounting for "rekeying" -- for now assuming that can happen
+/// internally and machine can just stay in transport state.
```
cpacejo, line 506: I agree with this choice!

adapter/ph/src/km.rs
```diff
@@ -0,0 +1,919 @@
...
+    Transport,
+    Error,
+}
+
+/// A set of constant settings for a particular [KeyManagerStateMachine].  Used
+/// to configure the running of the [KeyManager].
+#[derive(Debug, Clone)]
+pub struct KMSettings {
+    /// The ZDP defined type value for this Key Management system.
+    pub zdp_km_type: u16,
```
cpacejo: Feel free to add a type and consts to `zpr.rs` for this.

cpacejo (approved):
I think this looks good!  I like the design.  I'm curious what will change when you switch to Noise.  Just a few nits/simplifications.  Thanks for your patience as I got to everything :)

adapter/cd/src/cd/mod.rs
mkolehmainen: It's far easier to do it this way (as far as I know how) since the rx side of a mpsc channel does not implement copy.  So what I do is create those in the scope where they are used.

adapter/cd/src/zdp/client.rs
mkolehmainen: I do agree!

adapter/ph/src/km.rs
mkolehmainen: ok

adapter/ph/src/km.rs
mkolehmainen: I have the same issue here as with other uses of mpsc.  I don't yet know how best to create the rx side so that it can be used where it is not created.  I use the rx in the run-loop, so that's where I create the tx.  I'm putting a todo note in the code.

adapter/ph/src/km.rs
mkolehmainen: I will do this on next checking when I add Noise.

adapter/ph/src/km.rs
mkolehmainen: Ok I'll leave this for now.  It would be nice if the KM didn't know anything about ZPI.  Since the ZPI and SA_ID share 1 whole byte, we need to decide which bits are for SA and which for ZPI.

adapter/ph/src/km.rs
mkolehmainen: I like that.  But.... handling km messages should not be in fast path, right?

mkolehmainen: For now I'm adding a non-async version too.  (Using tokio `try_send`).

PR #303, opened 2024-08-08
introduce RCU-ish functionality to the PH

cpacejo:
This is my attempt at a very basic [RCU](https://en.wikipedia.org/wiki/Read-copy-update)-"ish" API for the PH.  It's greatly simplified to restrict our usage of it, so we can swap out the backend implementation if needed.  There are several 3rd party crates which implement something "like" RCU and it's not obvious yet which is the best for us.  (Some already I've ruled out due to API incompatibility / design flaw reasons.)

Also, I've used this for the BPF filtering.  This needs a fast read-side (since we read it on every packet) so is a great candidate.

This needs tests, but they'd need to do concurrent control stuff, so I'm punting that for now.

### DIFF
--- a/adapter/ph/Cargo.lock
+++ b/adapter/ph/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "aarc"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


mkolehmainen (approved):
Looks good. To me, crossbeam-epoch looks to have more brain power behind it, fwiw.

I'll add a story to try to use this stuff in the KM instead of my tired arc+mutex tricks.

PR #304, opened 2024-08-08
Send_sync_req_helper cancellation safe

### LINKED ISSUE #269: Make request/reply handler cancellation-safe
Right now, if someone calls it e.g. in a select!(), but then doesn't finish executing it (because e.g. some other event occurred and they don't care about getting a response any more), the buffer it allocates will never be returned to the pool.  To fix this, you'll need to use a DropGuard (or use the get_buffer_guarded() variant) to properly return the buffer on scope exit.

Blocked on:
- org-zpr/zpr-core#264
- org-zpr/zpr-core#268

owenrhysthomas:


cpacejo:
Agh ok I have dropped the ball in my planning!!  The changes I've made to the management queues conflict with this in a way that makes the drop guard no longer necessary here.  (Enqueueing now no longer blocks.)  So the merge would just stomp this.  I'm sorry about that!  There will be other places in the management path we have to be aware of and address cancellation safety though – you did take the right approach here.

cpacejo:
Oh thinking more!!!  We might actually _want_ blocking interfaces for management egress, which we would use here… reopening for now… but it will be a few days until I get to that… #317

PR #305, opened 2024-08-08
put ph tests run commands on multiple lines

owenrhysthomas:


owenrhysthomas:
Doesn't change any functionality of the runners, just removed `&&` syntax from runners that execute multiple commands and puts each command on its own line. 

@mkolehmainen Just wanted you to see this because you mentioned putting runners on multiple lines on Slack.

nonrenewable (approved):
it is prettier this way ✨

PR #310, opened 2024-08-10
zpr-ext: implement very basic ancillary data support

cpacejo:
This is quick and hacky and has buggy edge cases but it should work for what we're using it for until the nightly feature becomes stable!


owenrhysthomas (approved):
Looks good, thanks!

PR #312, opened 2024-08-12
addition of ancillary data

### LINKED ISSUE #176: pass FD instead of filename for packet capture

owenrhysthomas:


adapter/ph/src/rpc_worker.rs
```diff
@@ -87,15 +96,29 @@ async fn handle_connection(
...
-            },
+            "SET-CAPTURE" => {
+                // Tell debug tool we're ready for the ancillary data
+                buf_writer.write_all("SEND ANCILLARY\n".as_bytes()).await?;
+                buf_writer.flush().await?;
+
+                // Receive ancillary data
+                let mut ancillary_buffer = [0; config::ANCILLARY_BUFFER_SIZE];
+                let mut ancillary = SocketAncillary::new(&mut ancillary_buffer);
+                let mut buf = [0; 8];
```
owenrhysthomas: could use const here instead of `8`

adapter/ph/src/rpc_worker.rs
```diff
@@ -294,11 +317,25 @@ fn values_from_hist(hist_name: &str, units: &str, hist: &Histogram<u64>) -> Stri
...
+async fn set_capture(asm: &Assembly<'_>, ancillary: SocketAncillary<'_>) -> String {
+    // Get the ancillary data
+    let anc_message = ancillary.messages().nth(0).unwrap();
+    // Get the SCM rights from the ancillary data
+    if let AncillaryData::ScmRights(mut scm_rights) = anc_message.unwrap() {
+        // See if there's actually data in the scm_rights, if yes try to open a
+        // capture file, otherwise report failure to open file
+        match scm_rights.nth(0) {
+            Some(fd) => {
+                let file = unsafe { File::from_raw_fd(fd) };
```
owenrhysthomas: should have comment with justification of safety for this unsafe operation, but I'm not sure what that justification is.

adapter/ph-debug/src/main.rs
```diff
@@ -149,9 +155,48 @@ fn handle_perf_sample(duration: u64, frequency: u64, port: &str) -> std::io::Res
...
+        .read(true)
+        .write(true)
+        .create(true)
+        .open(file_path)
+        .unwrap()
+        .as_raw_fd();
+
+    let mut ancillary_buffer = [0; ANCILLARY_BUFFER_SIZE];
+    let mut ancillary = SocketAncillary::new(&mut ancillary_buffer);
+    let borrowed_fd = unsafe { BorrowedFd::borrow_raw(file_descriptor) };
```
owenrhysthomas: also need justification for use of unsafe here

adapter/ph/src/rpc_worker.rs
cpacejo: (Aside – with my latest changes to the ancillary data stuff, you can do `File::from(fd.try_into_owned())` to do this safely!)

As written above though – and generally – to write a good safety justification, you can base it on the safety requirement of the function you're calling.  So for [`from_raw_fd()`](https://doc.rust-lang.org/nightly/std/os/fd/trait.FromRawFd.html#tymethod.from_raw_fd), the requirement is:

> The fd passed in must be an [owned file descriptor](https://doc.rust-lang.org/nightly/std/io/index.html#io-safety); in particular, it must be open.

So a correct safety comment could be something like:

```
// SAFETY: Any FD received in ancillary data is owned by us and open.
```

adapter/ph-debug/src/main.rs
cpacejo: You can use [`.as_fd()`](https://doc.rust-lang.org/std/fs/struct.File.html#impl-AsFd-for-File) to go directly from a `File` to a `BorrowedFd` safely.  (And then, with my new API, `.from()` or `CowFd::into()` to go from a `BorrowedFd` into a `CowFd`.)

adapter/ph-debug/src/main.rs
```diff
@@ -149,9 +155,48 @@ fn handle_perf_sample(duration: u64, frequency: u64, port: &str) -> std::io::Res
 }
 
 fn handle_set_capture(file_path: String, port: &str) -> std::io::Result<()> {
-    let command = format!("SET-CAPTURE {}\n", file_path);
+    let file_descriptor = OpenOptions::new()
+        .read(true)
```
cpacejo: Read permission shouldn't be necessary I think.

adapter/ph-debug/src/main.rs
```diff
@@ -149,9 +155,48 @@ fn handle_perf_sample(duration: u64, frequency: u64, port: &str) -> std::io::Res
...
+        .open(file_path)
+        .unwrap()
+        .as_raw_fd();
+
+    let mut ancillary_buffer = [0; ANCILLARY_BUFFER_SIZE];
+    let mut ancillary = SocketAncillary::new(&mut ancillary_buffer);
+    let borrowed_fd = unsafe { BorrowedFd::borrow_raw(file_descriptor) };
+    ancillary.add_fds(&[borrowed_fd]);
+
+    let buf = [1; 8];
```
cpacejo: I think this can even just be a single byte (`[1; 1]`) – no need to define a constant in that case.  Though I'd add a comment stating _why_ we need this (i.e. because you need to send some real data with ancillary data).

adapter/ph/src/rpc_worker.rs
```diff
@@ -294,11 +318,25 @@ fn values_from_hist(hist_name: &str, units: &str, hist: &Histogram<u64>) -> Stri
...
+async fn set_capture(asm: &Assembly<'_>, ancillary: SocketAncillary<'_>) -> String {
+    // Get the ancillary data
+    let anc_message = ancillary.into_messages().nth(0).unwrap();
+    // Get the SCM rights from the ancillary data
+    if let AncillaryData::ScmRights(mut scm_rights) = anc_message.unwrap() {
+        // See if there's actually data in the scm_rights, if yes try to open a
+        // capture file, otherwise report failure to open file
+        match scm_rights.nth(0) {
+            Some(fd) => {
+                let file = unsafe { File::from_raw_fd(fd.as_raw_fd()) };
```
cpacejo: So now with `into_messages()`, you should be able to call `fd.try_into_owned()?` to get an `OwnedFd` without needing `unsafe`.  (Better than the proposed standard! :) )

adapter/ph/src/rpc_worker.rs
cpacejo: Even just 1 byte I think is enough – and then no need for a const, just a comment that this is receiving the necessary byte you are sending on the other side.

cpacejo (changes requested):
Nice looks good!  Just requesting a couple changes to make to make use of the new API changes I made so you can avoid `unsafe`.  (I can explain the "Cow" stuff in person/over Slack too.)

Also I'm realizing how glad I am to have this functionality!  I am reminded of past products I've worked on that _didn't_ have this functionality and how anxious I and others would get using a debug tool in production, that we might accidentally overwrite some critical system file when asking the main process to log something somewhere…

cpacejo (approved):
Perfect thanks!!  Weird that Tokio File doesn't implement `From<OwnedFd>`, what you have is good tho!

PR #313, opened 2024-08-12
Simplify mgmt workers

cpacejo:
This eliminates the final redundant queue – since the outbound processor only is encoding mgmt messages now, we can just do that directly to the interface queue.  I've moved all the message encoding into a new `mgmt` module.

Also, since the inbound processor now only processes management messages, I've renamed it to `mgmt_processor` and eliminated batching.

Finally – bump the PH version to 0.2.0 to reflect all these structural changes!


adapter/ph/src/mgmt.rs
```diff
@@ -0,0 +1,76 @@
...
+    fastpath::substrate_egress(
+        asm,
+        link_id,
+        zpr::ZPI_0, // TODO
+        packet,
+    );
+}
+
+pub async fn send_report<'pktbuf>(asm: &Assembly<'pktbuf>, link_id: zpr::LinkId, report: &str) {
+    // this condition will need to be adjusted when we have complete ZPR packets
```
owenrhysthomas, line 59: How come `send_report` and `send_discard` are in here, but we leave `send_hello_req` in the Assembly?

owenrhysthomas (approved):
Nice!

adapter/ph/src/mgmt.rs
cpacejo: Oh good question!  The difference is it's a synchronous message – I started trying to factor them out also but realized it would get complicated and messy, especially since we don't have a proper notion of "link" in place yet, where we'd keep the sync request state.  So that is coming later!  (May be a task you can do once I figure out how we want to manage link state!)

PR #314, opened 2024-08-12
more documentation

### LINKED ISSUE #309: more documenting

owenrhysthomas:


adapter/ph-debug/src/main.rs
```diff
@@ -122,7 +128,7 @@ fn handle_watch(time: u64, port: &str) -> std::io::Result<()> {
         let counts: Vec<&str> = response.split('\n').collect(); // Split the messages
 
         // TODO error checking, make sure actually got a message back, and that it's the correct message
-        for n in 1..7 {
+        for n in 1..=NUM_COUNTERS {
```
cpacejo, line 130: Tangent – I think you can do without this constant by using `for count in &counts[1..]` here, and defining `values` as a `Vec::new()` above, and `push()`ing to it in the body here.  Then there's no need to keep `NUM_COUNTERS` in sync with the number of counters in the PH.  (If you do change this do it in a separate PR! :) from a docs point of view this is good!)

adapter/ph-debug/src/main.rs
```diff
@@ -228,6 +243,8 @@ impl CtrlcHandle {
         self.cv.notify_one();
     }
 
+    // Waits for a specified duration, but the timeout can be interrupted if wait becomes
+    // false
```
cpacejo, line 246: Triple slashes will make this a rustdoc comment!  (Same comment for other `fn` comments in this PR.)

adapter/ph/src/flow_control.rs
```diff
@@ -1,3 +1,6 @@
+// Controls whether a packet will be copied and captured in the packet
+// capture framework
+
```
cpacejo: Here you can use `//!` to make this a rustdoc comment on the module itself, which will show up with `cargo doc --open`.

(Same comment for the other module-level comments.)

cpacejo (approved):
Thanks for doing this!!  This will help future me a lot haha.  Just noted some places you can turn the normal comments into rustdoc comments (by adding an extra `/` or `!` as appropriate – next-item doc or enclosing-item doc) so they'll show up in `cargo doc --open`.  Feel free to push after modifying those, or change in a separate commit.

adapter/ph-debug/src/main.rs
owenrhysthomas: Will do, I'll create a new pr to do this

PR #318, opened 2024-08-13
rework UnixStream ancillary stuff to not suffer from a resource leak

cpacejo:
Rust's proposed ancillary data API suffers from a potential resource leak / DoS attack – FDs in received ancillary data are not owned, and therefore if they aren't extracted and converted to an `OwnedFd` by the caller, they will never be closed (a resource leak).  So this change allows FDs to be owned directly by the `SocketAncillary` struct, so if they aren't accessed, they're simply closed.  This also required changing the API slightly (thus deviating from Rust's proposed API); namely `messages()` became `into_messages()`.

(And because the same `SocketAncillary` struct is used for FDs being _sent_ as ancillary data, but sending does _not_ take ownership, I needed to implement a borrowed-or-owned type, called `CowFd`, which allows representing either in `SocketAncillary`.  Long story short I think Rust's proposed API still needs work.)

### DIFF
--- a/zpr-ext/Cargo.lock
+++ b/zpr-ext/Cargo.lock
@@ -367,7 +367,7 @@ dependencies = [
 
 [[package]]
 name = "zpr-ext"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "bytes",
  "libc",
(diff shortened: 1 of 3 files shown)


owenrhysthomas (approved):
Looks good. ~~Is this an issue we'd want to bring up on the issue for the proposed implementation for the standard version? I don't see it~~ ~~Seems someone's found it https://github.com/rust-lang/rust/issues/76915#issuecomment-1505924484~~ Haha I just saw you reacted to that comment. So just ignore this.

PR #324, opened 2024-08-14
add noise as key manager

### LINKED ISSUE #253: Adatper and Node set up a Security Assocaition (SA)
A previous story sets up code paths for key messages between the node and adapter.  This story is about implementing  real keying exchange on the link so that a security association is set up.

Ideally the SA is then used by all other ZDP chatter on the wire.

prerequisite:
- org-zpr/zpr-core#251

mkolehmainen:
This adds a key management machine for the Noise protocol.  The node and adapter have also been modified to do a little demo.

Some shared code used in the demo has been added to `ph/km_demo.rs` -- I don't expect that to live very long, it was just to avoid duplicating the code.

Note that the KM is modified a bit from original PR.  One thing it does is add a length field outside of the transport buffer. This was for cases where we wanted to just encrypt part of the buffer.  That may go away entirely since we are now talking about not doing that.

There is no KM interface for doing any of the HMAC stuff we've been talking about.  Also the snow library does not expose the "AD" interface of AEAD which is irritating but fixable.

I see there are a bunch of older commits on here that are already in main.  Not sure why.  This branch was branched from mk-251 and I did a rebase --onto operation after the mk-251 PR was squashed.  Anyway, that makes this PR extra confusing.


adapter/ph/src/km.rs
```diff
@@ -142,56 +175,49 @@ impl KeyManager<'_> {
...
-            }
-            encr = state.statemachine.get_transport_encryptor();
-            sa_id = state.sa_id;
+        let mut state = self.shared.state.lock().unwrap();
+        if state.statemachine.get_state() != KMSMState::Transport {
+            return Err(KMError::InvalidState);
+        }
+        if state.sa_id == 0 {
+            // programming error
+            panic!("SA_ID is zero");
```
svjaffer: Where do we prevent calling this function if this is true?

mkolehmainen: Maybe I should just return an error?  Bascially a 0 SA is invalid so you shouldn't be trying to encrypt or decrypt packets if SA is not set.  But really, it is hard to guarantee that it will be set when you call due to the async nature of the maintenance messages.  So yes I think I'll take that panic out!

mkolehmainen: Oh and as I changed it I noticed I was returning an error in the decrypt leg already, so now it's symmetrical.

adapter/cd/src/cd/zpr.rs
```diff
@@ -126,6 +126,19 @@ impl Zpr {
         Some(format!("{}:{}", conf.get_dock_host(), conf.get_dock_port()))
     }
 
+    /// Returns noise public key for dock
+    pub fn copy_dock_noise_key(&self, name: &str, key_out: &mut [u8; 32]) -> bool {
```
cpacejo: FWIW I think returning `Option<[u8; 32]>` is also fine here.  It's returned on the stack, then just a couple of register moves to copy it to wherever the caller wants it: https://godbolt.org/z/M37MK5Ecr (hooray for 128-bit registers!)

adapter/ph/src/zdp.rs
```diff
@@ -112,7 +121,29 @@ pub struct ZdpReportHeader {
 #[derive(FromZeroes, FromBytes, AsBytes, Unaligned)]
 #[repr(packed)]
 pub struct ZdpHelloResponseHeader {
-    pub status: u16,
```
cpacejo, line 117: Oh good catch!

cpacejo (commented):
(still reviewing… got through everything but the two big files)

adapter/ph/src/km_noise.rs
```diff
@@ -0,0 +1,695 @@
+//! KeyManager implementation using Noise protocol.
+//!
+//! Configured to use the "IK" pattern (see Noise paper).  This is same as what wireguard uses
+//! and requires that initiators know the 25519 public key of the responder.  The idea here
+//! is that an adapter wanting to connect to a remote node will have access to this key.
+//! Maybe through a special DNS record.
```
cpacejo, line 6: I wonder if the responder can return an x509 cert (i.e., a signed public key + identity) instead of just a plain public key?  Then the initiator doesn't need to know the responder's public key beforehand, just its identity and the signing authority's public key.

(I think predistributing the responder's public key is fine for now though!)

adapter/ph/src/km_noise.rs
```diff
@@ -0,0 +1,695 @@
...
+//! Note that we do not handle padding here. If padding is desireable
+//! caller needs to apply padding before calling encrypt.
+//!
+//! Note that the encrypt/decrypt functions expect to operate on entire
+//! buffer.  Caller is responsible for dealing with cases when only
+//! part of the buffer is to be encrypted or decrypted.
+//!
+//!
+//! Note you can create noise keys with the `wg` command line tool on recent versions of linux.
+//! For example, to generate a private key: `wg genkey`.  Then you can pass that private key
```
cpacejo, line 33: FWIW `openssl` can be used to generate these as well.  `openssl genpkey -algorithm X25519` seems to do the trick.  (Confirmed after extracting the private key that `wg pubkey` happily generates the correct public key.)

adapter/ph/src/km_noise.rs
```diff
@@ -0,0 +1,695 @@
...
+/// Not multi-thread safe.
+/// TODO: Figure out how to make encryption/decryption parallelizable.
+pub struct KMNoise {
+    settings: KMSettings,
+    state: KMSMState,
+    initiate: bool,
+    peer_pub_key: Option<Vec<u8>>, // required if initiator
+    local_keypair: snow::Keypair,
+    hs_state: Option<snow::HandshakeState>,
+    t_state: Option<snow::TransportState>,
```
cpacejo, line 63: Have you looked into [`StatelessTransportState`](https://docs.rs/snow/latest/snow/struct.StatelessTransportState.html)?  That looks like it'd let us manage the nonces and encrypt in parallel.

Thinking more on this… I think our sequence numbers suffice as nonces for mgmt messages… in which case we don't actually need to include the sequence number in the plaintext message with window size = 1 (because we know what the sequence number should be, and it won't decrypt if it's wrong).  (And in the case we want to allow a window size > 1 (Sarah is asking Frank for this right now), we'd want to move the sequence number, or at least a suffix of it, _outside_ encryption so we can easily handle reordered packets, else we'll be guessing at nonce values within the window when reordering occurs.)

For transit messages, in the case that encryption _is_ enabled on them, it's tempting to say just use 0 as a nonce, since we don't care about replay for them.  But apparently, this actually breaks the security of ChaChaPoly.  So for these, we'll still have to have a nonce, but we can grab it from an atomic counter (possibly in batches) to allow for parallelization.  But – in any case – the nonce will have to be emitted in plaintext outside the encryption, so the receiver knows what the heck was used.

The catch with all the above is that we are moving trust that a nonce isn't reused from the crypto code to the packet handler to do the right thing.  (E.g. when retransmitting a mgmt packet, right now we regenerate it expecting it should have the same body – if this isn't true, it would break encryption!)

So I suppose the "safe" thing to do generally is, choose nonce values from an atomic counter in _this module_, and paste them (or at least, a suffix of them, relying on loose synchronization to keep track of the prefix) in plaintext on the front of the ciphertext for _all messages_.  The only "downside" I see to this is the nonce value is redundant-ish with the mgmt sequence number.  So we're wasting a couple bytes.  BUT, I feel like this may be the right choice for now??

(If we do this – I think the notion that (a prefix of) the nonce is emitted before the ciphertext must be part of the RFC, defined per KM scheme.  Also whether we "factor out" the mgmt sequence number for a given KM scheme.  Since both sides need to agree on whether/how to do this.)

adapter/ph/src/km_noise.rs
```diff
@@ -0,0 +1,695 @@
...
+}
+
+impl KMNoise {
+    /// Create the Noise KeyManager.
+    ///
+    /// This requires Noise keys.  Eventually (TODO) we will also pass certificates through here
+    /// so that each side can check for certificate authority signautre.
```
cpacejo, line 70: Ah answering my question above! :)

adapter/ph/src/km_noise.rs
```diff
@@ -0,0 +1,695 @@
...
+        peer_pub_key: Option<Vec<u8>>,
+        local_keypair: Option<snow::Keypair>,
+    ) -> Result<Self, KMError> {
+        if initiate && peer_pub_key.is_none() {
+            error!("noise: peer public key required for initiator");
+            return Err(KMError::ConfigurationError);
+        }
+
+        let settings = KMSettings {
+            zdp_km_type: 2,
```
cpacejo: Grab `2` the from `zpr` module?

adapter/ph/src/km_noise.rs
```diff
@@ -0,0 +1,695 @@
...
+    ///
+    /// This requires Noise keys.  Eventually (TODO) we will also pass certificates through here
+    /// so that each side can check for certificate authority signautre.
+    ///
+    /// - `peer_pub_key` is required for initiator, and should match the `local_keypair` of the responder.
+    /// - `local_keypair` is optional. If not provided, a new keypair will be generated.
+    pub fn new(
+        initiate: bool,
+        peer_pub_key: Option<Vec<u8>>,
+        local_keypair: Option<snow::Keypair>,
```
cpacejo, line 77: I think in normal operation we'd want to require both of these, so both peers can validate each other?  (Which is annoying for testing of course but when we move to PKI model becomes not a problem.)

adapter/cd/src/cd/zpr.rs
mkolehmainen: Ah!

adapter/ph/src/km_noise.rs
mkolehmainen: Great!

adapter/ph/src/km_noise.rs
```diff
@@ -0,0 +1,697 @@
...
+        if self.state == KMSMState::Configuring {
+            let mut hs: snow::HandshakeState;
+            if self.hs_state.is_none() {
+                error!("noise: handle_message called but no handshake state set up");
+                return Err(KMError::InvalidState);
+            }
+            if self.hs_state.is_some() {
+                hs = self.hs_state.take().unwrap();
+                let mut buf = BytesMut::zeroed(1024);
+                match hs.read_message(&message[..], &mut buf) {
```
cpacejo: Nit: the `&[..]` I believe is unnecessary here as `message` is already a slice.

adapter/ph/src/km_noise.rs
mkolehmainen: We (I) should definitely rework this to use the stateless thing.  I somehow didn't notice that module before.  I'll do that in a follow on PR.  I'll add story to rethink nonces and sequence numbers, etc.

adapter/ph/src/km_noise.rs
```diff
@@ -0,0 +1,697 @@
...
+            Some(t) => t,
+            None => {
+                error!("noise: encrypt_transport called in wrong state");
+                return Err(KMError::InvalidState);
+            }
+        };
+        let nonce = ts.sending_nonce();
+        match ts.write_message(payload, message) {
+            Ok(len) => {
+                message[len..len + 8].copy_from_slice(&nonce.to_be_bytes());
```
cpacejo, line 293: Possible to insert the nonce at the beginning?  (Packet fields located at fixed offsets is always nice.)

Future improvement: we should be able to get away with just sending the low-order 4 bytes of the nonce.  The high-order 4 bytes we track simply by incrementing any time the received 4 bytes drops by more than 2^31.  (DTLS does something similar.)

adapter/ph/src/km.rs
```diff
@@ -200,24 +225,30 @@ impl KeyManager<'_> {
...
-                // Now we need to push a our SA id onto the front.
-                // (Note ZPI should be part of integrity protected data)
-                let zpi_buf = message.alloc_zeroed_headroom(1);
-                zpi_buf[0] = sa_id;
+                message.shrink(message.body().len()); // remove body
+                message.put(&encr_buf[0..len]); // write a new body
+
+                // Now write our headroom info:
+                let head_buf = message.alloc_zeroed_headroom(3); // ZPI + LEN
+                head_buf[0] = state.sa_id;
```
cpacejo, line 242: We did chat earlier just leaving a note for completeness – ZPI will already be present in the packet (and will be a different value than SAID).

adapter/ph/src/km.rs
```diff
@@ -227,52 +258,50 @@ impl KeyManager<'_> {
...
-                    "SA not in transport state",
-                ));
-            }
-            encr = state.statemachine.get_transport_encryptor();
-            sa_id = state.sa_id;
+        let mut state = self.shared.state.lock().unwrap();
+        if message.body()[0] == 0 {
+            return Err(KMError::SaIdZero);
+        }
+        if state.sa_id != message.body()[0] {
```
cpacejo, line 267: Same note here – ZPI value will be unrelated to SAID.

adapter/ph/src/km.rs
```diff
@@ -227,52 +258,50 @@ impl KeyManager<'_> {
...
+            return Err(KMError::InvalidState);
         }
 
         // TODO: Ability to decrypt in place. Not sure how to accomplish.  At very least we could use our own buffer pool.
         let mut decr_buf = [0u8; config::PACKET_BUFFER_SIZE];
 
-        // TODO: pass sa_id into decrypt
-        match encr.decrypt_transport(sa_id, &message.body()[1..], &mut decr_buf) {
+        // read the size of the encrypted payload.  Size follows the ZPI/SA_ID value:
+        let encr_len: usize = u16::from_be_bytes([message.body()[1], message.body()[2]]) as usize;
```
cpacejo, line 278: Possible to infer ciphertext length from the packet length, instead of including it in the packet body?

adapter/ph/src/km.rs
```diff
@@ -393,10 +413,27 @@ impl KeyManager<'_> {
             match prev_state {
                 KMSMState::Error => {
                     // If error, send reset and loop again
+                    match km_signals_out.send(KMSignal::Error).await {
+                        Ok(_) => {}
+                        Err(_) => {
```
cpacejo, line 415: `Err` can only occur if the channel is closed.  We probably want to bail out in that case I would think.  (Assuming these signals are important and someone ought to be paying attention to them?)

cpacejo (approved):
I left lots of comments!

I think the main "code" thing that I would change is how we handle errors writing to the signal channel.  But maybe I'm misunderstanding what that channel is for.  If it's OK to lose those messages maybe `mpsc` is the wrong channel type and some sort of pub/sub channel would be better.

Design-wise I had a couple comments on packet format (nonce at beginning, no need for length prefix) but otherwise looks great!!

adapter/ph/src/km_noise.rs
mkolehmainen: tracking this as https://github.com/org-zpr/zpr-core/issues/358

adapter/ph/src/km.rs
mkolehmainen: yes will be in next pr

adapter/ph/src/km.rs
mkolehmainen: This goes away in next pr since we do not use this interface for partial encryption.

adapter/ph/src/km_noise.rs
mkolehmainen: will do on next PR.

Issue #322, opened 2024-08-14
Investigate using SO_REUSEPORT to parallelize socket reads


cpacejo:
SO_REUSEPORT supposedly allows reading from one socket on multiple threads, with traffic being balanced by flow hash: https://lwn.net/Articles/542629/  Let's see (a) whether this works based on IPv6 flow label also, and (b) if so, use this in the PH so we can concurrently read from the same tether connection in a useful fashion.  (Otherwise, for parallelization, we'd need to have multiple ZDP sessions.)

cpacejo:
Doing this now because (a) I want concurrent ingress pipelines to be in place so our code is correctly structured around them, and (b) I want packet flows distributed amongst said pipelines, to flush out any bugs we'd have with them.

PR #327, opened 2024-08-14
publicize BPF codes

cpacejo:
Makes the cBPF instruction codes public so other code can use them!  (Needed for packet steering #322.)


owenrhysthomas (approved):
Nice! Is this something you'd want to break out into its own file eventually now that it's a module? Or keep it in the lib.rs?

cpacejo:
> Nice! Is this something you'd want to break out into its own file eventually now that it's a module? Or keep it in the lib.rs?

Oh I could do that!  I'm still thinking about futzing with it a bit (make more of the stuff in /usr/include/linux/filter.h available, just not sure the best naming scheme).  Maybe even break it out into a separate crate.

PR #328, opened 2024-08-14
zpr-ext: attach_reuse_port_cbpf

cpacejo:
Add a wrapper for SO_ATTACH_REUSEPORT_CBPF.  Prerequisite for #322 packet steering.

### DIFF
--- a/zpr-ext/src/tokio.rs
+++ b/zpr-ext/src/tokio.rs
@@ -25,25 +25,29 @@ pub mod net {
 
         #[cfg(any(doc, target_os = "android", target_os = "linux"))]
         fn attach_reuse_port_cbpf(&self, filter: &[libc::sock_filter]) -> io::Result<()> {
-            unsafe {
-                let fprog = libc::sock_fprog {
-                    len: filter.len() as u16,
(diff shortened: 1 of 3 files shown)


zpr-ext/src/tokio.rs
```diff
@@ -20,6 +22,30 @@ pub mod net {
                 Err(errno) => Err(io::Error::from(errno)),
             }
         }
+
+        #[cfg(any(doc, target_os = "android", target_os = "linux"))]
+        fn attach_reuse_port_cbpf(&self, filter: &[libc::sock_filter]) -> io::Result<()> {
+            unsafe {
```
owenrhysthomas: Maybe just a comment for justification of usage of unsafe function.

owenrhysthomas (approved):

zpr-ext/src/tokio.rs
cpacejo: Ahhh keeping me honest 😅 there's a lot of unsafety in this block lemme see what I can do…

cpacejo:
Was actually able to shrink the scope of the `unsafe` block!  & added a comment addressing the remaining unsafety.

PR #329, opened 2024-08-14
Eliminiated need to keep NUM_COUNTERS const in ph-debug

### LINKED ISSUE #323: remove NUM_COUNTERS from ph-debug
https://github.com/org-zpr/zpr-core/pull/314#discussion_r1716978040

owenrhysthomas:


adapter/ph-debug/src/main.rs
```diff
@@ -126,21 +126,32 @@ fn handle_watch(frequency: u64, port: &str) -> std::io::Result<()> {
         // with each line as a different index of the vector
         let counts: Vec<&str> = response.split('\n').collect(); // Split the messages
 
+        if first_run {
+            for _n in &counts[1..] {
+                values.push(0);
+            }
```
cpacejo: You can alternatively use `values.resize()` to set the Vec size in one operation.

adapter/ph-debug/src/main.rs
```diff
@@ -126,21 +126,32 @@ fn handle_watch(frequency: u64, port: &str) -> std::io::Result<()> {
...
+        }
         // TODO error checking, make sure actually got a message back, and that it's the correct message
-        for n in 1..=NUM_COUNTERS {
+        let mut n = 0;
+        for count in &counts[1..] {
```
cpacejo: I just learned another way to write this is:
```
for (n, count) in counts[1..].iter().enumerate() {
  ...
}
```
which will give you n = 0, 1, 2, etc... paired with each successive item from the slice.  But what you have is fine also!

cpacejo (approved):
Nice!  Always good to decouple systems like this.

adapter/ph-debug/src/main.rs
owenrhysthomas: I tried doing that with the .enumerate(), but it didn't work because the structure wasn't an iterator, I didn't realize you could do .iter()! Very nice.

PR #330, opened 2024-08-14
use SO_REUSEPORT to support multiple ingress workers

cpacejo:
Two changes here:

1. I use the SO_REUSEPORT socket option to allow opening multiple susbtrate ingress sockets, and therefore multiple substrate ingress workers.  This sets us up for correctly writing code which is shared by multiple workers (such as the Dock Lookup Table).  In this configuration, packets get distributed by the kernel to workers based on their IP flow hash, meaning that links (which are identified by IP source address+port) get pinned to a worker.  This is good because DLT keys (link:flow) can be partitioned by link ID.  This is bad because it can lead to uneven load balancing (e.g. if there's only one link, only one worker will get used); that will be addressed in another PR.

2. The sockets no longer `connect()`, we now use `send_to()` to specify the destination for each packet.  This was necessary for load balancing to work correctly.  (If the sockets were connected Linux did not load balance amongst them.)  It's labeled as a "TEMP HACK" for now because I have a follow-on task to actually make a link lookup table, vs. right now it only works for a single link.

These mostly address #322.

### DIFF
--- a/adapter/ph/Cargo.lock
+++ b/adapter/ph/Cargo.lock
@@ -579,6 +579,7 @@ dependencies = [
  "open-enum",
  "openssl",
  "openssl-sys",
+ "socket2",
  "tokio",
  "tokio-tun",
  "tokio-util",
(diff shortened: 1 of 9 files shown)


svjaffer (approved):

cpacejo:
Oh also I enabled multiple agent output queues in the TUN interface!

PR #331, opened 2024-08-15
add some more documentation to packet

mkolehmainen:
I've attempted to add some documentation that I would have found helpful with the hope that it will aid future explorers.

### DIFF
--- a/adapter/ph/src/packet.rs
+++ b/adapter/ph/src/packet.rs
@@ -4,7 +4,7 @@
 //! is responsible for allocating the buffer and keeping it around for as long as
 //! the packet is needed.
 //!
-//! For example usags, see the [Packet] documentation.
+//! For usage examples see the [Packet] documentation.
 
 use crate::config;
 use crate::net_defs::*;
@@ -38,7 +38,7 @@ use zpr_ext::std::mem::DropGuard;
 /// extended into either of these, but not beyond.  The size of these
 /// is defined when the `Packet` handle is created.
 ///
-/// Note that a packet is [bytes::BufMut] (among other things) so you can use the
+/// Note that a packet is [bytes::BufMut] and [bytes::Buf] so, for example, you can use the
 /// `put` method on Packet to append data to the packet body. Note that to get
 /// this (and other interesting) functionality you must have the correct traits in
 /// scope.
@@ -62,7 +62,8 @@ use zpr_ext::std::mem::DropGuard;
 ///     let mut sock_buf = [0u8; 4096];
 ///     let (read_len, source_addr) = sock.recv_from(&mut sock_buf)?;
 ///
-///     // We need a backing byte buffer for the packet.
+///     // We need a backing byte buffer for the packet.  This would tupically be allocated
+///     // from a buffer pool.
 ///     let mut pkt_buf = [0u8; config::PACKET_BUFFER_SIZE];
 ///
 ///     // Create the packet. Since we are reading this packet with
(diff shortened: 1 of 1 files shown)


adapter/ph/src/packet.rs
```diff
@@ -1,4 +1,10 @@
 //! This module contains all state of a packet which is moving through the system.
+//!
+//! A [Packet] is a lightweight wrapper around a byte buffer. Creator of the packet
+//! is responsible for allocating the buffer and keeping it around for as long as
+//! the packet is needed.
+//!
+//! For example usags, see the [Packet] documentation.
```
cpacejo: Typo: "usags"

adapter/ph/src/packet.rs
```diff
@@ -8,22 +14,129 @@ use zerocopy::{AsBytes, ByteOrder, FromBytes, FromZeroes, NetworkEndian};
...
+/// *Headroom* is space in the buffer which is set aside when the packet
+/// is created (see [Packet::new]). It is useful for when you need to slap headers or other
+/// front matter onto the packet.  Use the `alloc_*` series of functions to
+/// push strucutres onto the packet body by taking bytes from the headroom.
+///
+/// The packet *body* resides between headroom and tailroom.  It can be
+/// extended into either of these, but not beyond.  The size of these
+/// is defined when the `Packet` handle is created.
+///
+/// Note that a packet is [bytes::BufMut] (among other things) so you can use the
```
cpacejo: Also `bytes::Buf`, through which data can be read from the front of the packet body (removing it), e.g. through [`copy_to_slice()`](https://docs.rs/bytes/latest/bytes/buf/trait.Buf.html#method.copy_to_slice).  (`Buf` and `BufMut` are independent traits & don't imply one another.)

adapter/ph/src/packet.rs
```diff
@@ -8,22 +14,129 @@ use zerocopy::{AsBytes, ByteOrder, FromBytes, FromZeroes, NetworkEndian};
...
+///     let mut pkt_buf = [0u8; config::PACKET_BUFFER_SIZE];
+///
+///     // Create the packet. Since we are reading this packet with
+///     // full (ZDP) headers already on it and we don't plan on pushing
+///     // anything on to the front, we don't need any headroom so
+///     // we set it to 0.
+///     let mut pkt = packet::Packet::new(&mut pkt_buf, 0);
+///
+///     // Write (copy) the received data into the packet.
+///     pkt.put(&sock_buf[..read_len]);
```
cpacejo: Ah so this is where the magic of `BufMut` comes into play!  `std::net:UdpSocket` doesn't support it, but `tokio::net::UdpSocket` does.  With which you can directly `sock.recv_buf(&mut pkt)` ([reference](https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.recv_buf)).  No need for the temporary `sock_buf` or extra copy.  So complete sequence then is:
```
    let mut pkt_buf = [0u8; config::PACKET_BUFFER_SIZE];  // would typically be allocated from a buffer pool
    let mut pkt = packet::Packet::new(&mut pkt_buf, 0);
    sock.recv_buf(&mut pkt);
```

cpacejo (approved):
This is great, thank you for clarifying this!  When I write code because it came out of my own brain it's not always apparent to me where explanation is lacking.

Just a nit in the first example you can simplify it by taking advantage of `BufMut`.  I can also clarify this myself if you check it in as-is.

adapter/ph/src/packet.rs
mkolehmainen: Ah!  I'm actually using tokio:sockets in the demo code for KM and didn't know that!

Issue #332, opened 2024-08-15
per-flow packet steering should offset by IP hash


cpacejo:
This is necessary to avoid "clumping" due to stream ID correlation between links (e.g. multiple links using low-numbered stream IDs).  However it requires using an eBPF steerer, not a cBPF steerer, which involves extra system calls I don't have time to wrap right now.  See <https://github.com/torvalds/linux/blob/master/net/core/sock_reuseport.c#L595-L598> and <https://github.com/torvalds/linux/blob/master/net/core/filter.c#L11426-L11428> for context.

PR #334, opened 2024-08-15
ingress packet steering

cpacejo:
Uses a cBPF program to tell the kernel how to steer packets into our queues.  As-is, this code steers unencrypted per-flow messages (which includes, importantly, transit packets) by stream ID, and all other packets by IP hash (effectively, by link ID).

This still suffers from the issue that stream IDs may be correlated between links (e.g. multiple links all using low-numbered stream IDs).  The solution to this issue is to add the IP hash to the stream ID.  Unfortunately that information is not available to a cBPF packet steerer (which I think must be an oversight by the kernel folk, since it's available to cBPF filters).  It _is_ available to an eBPF packet steerer, but that interface is more complex, so I'm just making that a TODO for now (#332).

Also, `main()` is getting messy, I've opened #333 to address that.

### DIFF
--- a/adapter/ph/Cargo.lock
+++ b/adapter/ph/Cargo.lock
@@ -143,7 +143,7 @@ checksum = "8318a53db07bb3f8dca91a600466bdb3f2eaadeedfdbcf02e1accbad9271ba50"
 
 [[package]]
 name = "cbpf-rs"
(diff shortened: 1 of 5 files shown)


mkolehmainen (approved):
Looks good to me!

PR #335, opened 2024-08-15
capture integration test

### LINKED ISSUE #320: packet capture integration test

owenrhysthomas:


owenrhysthomas:
So far this program 
- creates and sets a capture program
- pings the client and server
- closes the capture file and deletes the capture program
- opens vim with the resultant capture file for manual verification of successful capture

Edited 9/19/24
- checks how many inbound packets were captured and ensures correctness
- checks to make sure capture file for ph with no program set is empty

owenrhysthomas:
When/if this is approved, I'll create a GitHub workflow for the test

adapter/ph/test/capture-test.sh
```diff
@@ -0,0 +1,261 @@
...
+
+  sudo ip netns exec zpr-a ping -q -c 5 -w 5 "$B_ZPR_ADDR6" & wait -f $!
+  sudo ip netns exec zpr-b ping -q -c 5 -w 5 "$A_ZPR_ADDR6" & wait -f $!
+}
+
+function set_program() {
+  SOCKET=$1
+  FILE_NAME=$2
+  PROGRAM=$3
+  sudo "$PH_DEBUG_BIN" -c SET-CAPTURE -p "$SOCKET" --file-path "$FILE_NAME"
```
cpacejo: I think `sudo` should be unnecessary here, since the PH is running as the invoking user and thus creates the socket file with according permissions.  (This is what `sudo -E -u "$ZPR_USER"` does for us.)

You also then should be able to get rid of the `sudo` when invoking `tcpdump` below.

(More generally (not relevant for this test) – the debug tool only needs to run with enough permission to access the socket file.  So so long as that is created with permissions for the invoking user to access it, the debug tool doesn't need elevated privileges.)

adapter/ph/test/capture-test.sh
```diff
@@ -0,0 +1,261 @@
...
+B_ZPR_ADDR=192.168.1.2
+A_ZPR_ADDR6=fd00:1:1::1
+B_ZPR_ADDR6=fd00:1:1::2
+
+A_ZPR_SOCK=server.sock
+B_ZPR_SOCK=client.sock
+
+#
+# Helper functions
+#
```
cpacejo, line 24: All the functions which are shared between this and `basic-test.sh`, you can split out into a separate file.  Call it `common-funcs.sh` or such.  Then, in each of the tests, `source "$(dirname $0)/common_funcs.sh"` will pull them in.

adapter/ph/test/capture-test.sh
```diff
@@ -0,0 +1,261 @@
...
+if ! ping_test
+then PASS=1
+fi
+
+close_program "$A_ZPR_SOCK"
+close_program "$B_ZPR_SOCK"
+
+# Make sure correct number of incoming packets were captured, either 23 or 24 depending 
+# on whether two hello requests are received
+sudo tcpdump -r cap_test1.pcap 'link[0] = 1' > checker1.txt
```
cpacejo: Let's stick `checker1.txt` into `$TMPDIR` too.

adapter/ph/test/capture-test.sh
```diff
@@ -0,0 +1,261 @@
...
+echo "Carrier has arrived."
+
+stty sane || true
+
+#
+# Run test
+#
+
+set_program "$A_ZPR_SOCK" cap_test1.pcap 'link[0] == 1 || link[0] == 0'
+set_program "$B_ZPR_SOCK" cap_test2.pcap None
```
cpacejo: Let's stick these files into $TMPDIR, like: "$TMPDIR/cap_test1.pcap". This avoids the chance to overwrite any local files named the same.

adapter/ph/test/capture-test.sh
```diff
@@ -0,0 +1,261 @@
...
+if [[ "$SIZE" != "0" ]]
+then PASS=1
+fi
+
+#
+# Cleanup
+#
+
+sudo rm cap_test1.pcap
+sudo rm cap_test2.pcap
```
cpacejo: & if we stick these in `$TMPDIR`, no need to clean up explicitly, as they'll be removed when the directory is removed by `cleanup()`.

adapter/ph/test/capture-test.sh
```diff
@@ -0,0 +1,261 @@
...
+PACKET_COUNT="$(grep -c '0x0000:  01' checker1.txt)"
+
+if [[ ("$PACKET_COUNT" != "23" || "$HELLO_COUNT" != "1") &&  ("$PACKET_COUNT" != "24" || "$HELLO_COUNT" != "2") ]]
+then PASS=1
+fi
+
+# Make sure no data was captured when program is not set
+SIZE="$(wc -c cap_test2.pcap | awk '{print $1}')"
+
+if [[ "$SIZE" != "0" ]]
```
cpacejo: An alternative Bash trick is:
```
if [ -s "$TMPDIR/cap_test2.pcap" ]
```
`-s` checks whether the file is non-empty.  `help test` gives a list of these.  (`test` and `[` `]` are the same thing.)

(Also more generally – alternative to using `wc` to determine file size – since `wc` reads the whole file counting characters – is `stat -c %s` which directly reads the file size.)

cpacejo (approved):
Nice thanks for doing this!! I'll feel better making sweeping pipeline changes now haha.  Just a couple things I think we can improve on.

adapter/ph/test/capture-test.sh
owenrhysthomas: I realized I actually can't check if the file is empty, because it will have the 24 byte pcap file header, not sure how this was working before. I did switch to using stat.

PR #337, opened 2024-08-16
added run conditional for GitHub workflows

### LINKED ISSUE #306: only execute runners when you've edited relevant code

owenrhysthomas:
This PR added path conditionals when a workflow is triggered by a pull request, meaning that the associated workflows only run when the associated with a certain path. An extension of this could be to split some of the workflow files up further, for example give ph-debug its own runner so the ph-debug runners only execute when ph-debug files are changed, not when any file in the adapter/ph or adapter/cd or adapter/ph-debug folder are edited.

### DIFF
--- a/.github/workflows/go-visaservice.yml
+++ b/.github/workflows/go-visaservice.yml
@@ -7,6 +7,9 @@ on:
   push:
     branches: [ "main" ]
   pull_request:
+    paths:
+      - visaservice/mods/**
+      - visaservice/core/**
 
 jobs:
 
--- a/.github/workflows/go-visaservice.yml
+++ b/.github/workflows/go-visaservice.yml
@@ -10,6 +10,7 @@ on:
     paths:
       - visaservice/mods/**
       - visaservice/core/**
+      - visaservice/thrift/**
 
 jobs:
 
--- a/.github/workflows/rust-adapter.yml
+++ b/.github/workflows/rust-adapter.yml
@@ -5,8 +5,8 @@ on:
     branches: [ "main" ]
   pull_request:
     paths:
-      - adapter/ph
-      - adapter/cd
+      - adapter/ph/**
+      - adapter/cd/**
 
 env:
   CARGO_TERM_COLOR: always
--- a/.github/workflows/rust-adapter.yml
+++ b/.github/workflows/rust-adapter.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches: [ "main" ]
   pull_request:
+    paths:
+      - adapter/ph
+      - adapter/cd
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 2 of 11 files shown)


owenrhysthomas:
I'm not 100% sure what all the various dependencies between various modules are. I did my best to make sure that associated runners will execute at the appropriate time, but I'm not positive I did that, perhaps people working on the different pieces before we commit this just to make sure its what they want/need/expect.

.github/workflows/go-visaservice.yml
```diff
@@ -7,6 +7,9 @@ on:
   push:
     branches: [ "main" ]
   pull_request:
+    paths:
+      - visaservice/mods/**
+      - visaservice/core/**
```
mkolehmainen, line 12: add `visaservice/thrift/**`

.github/workflows/rust-node.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches: [ "main" ]
   pull_request:
+    paths: 
+      - node/**
```
mkolehmainen, line 8: add `visaservice/thrift/**`

.github/workflows/rust-visaservice.yml
```diff
@@ -4,7 +4,9 @@ on:
   push:
     branches: [ "main" ]
   pull_request:
-
+    paths:
+      - visaservice/cli/**
```
mkolehmainen, line 8: add `visaservice/thrift/**`

.github/workflows/rust-zprext.yml
```diff
@@ -1,9 +1,11 @@
-name: Rust-cbpfrs
+name: Rust-zprext
 
 on:
   push:
     branches: [ "main" ]
   pull_request:
+    paths:
```
mkolehmainen, line 7: If you alter zpr-ext, you better also recompile:
- adapter/cd
- adapter/ph
- adapter/ph-debug
- node

.github/workflows/rust-adapter.yml
```diff
@@ -4,6 +4,10 @@ on:
   push:
     branches: [ "main" ]
   pull_request:
+    paths:
+      - adapter/ph/**
```
mkolehmainen, line 8: note that if you alter ph, you should recompile node.

nonrenewable (approved):
I really like the concept!! Thank you for doing this 🙇 

But i don't know all the dependencies so either @svjaffer and @cpacejo also need to look/guarantee, or just merge it in and as things get weird - we know about this and fix it as we go!

.github/workflows/rust-adapter.yml
```diff
@@ -4,6 +4,11 @@ on:
...
+      - adapter/cd/**
+      - adapter/ph-debug/**
+      - zpr-ext/**
```
cpacejo, line 11: `cbpf-rs` and `cslab` also are dependencies of both the adapter and node

cpacejo (approved):
Awesome!!  Just a couple extra dependencies to add for the adapter and node

PR #341, opened 2024-08-19
RCU-friendly concurrent slab

cpacejo:
This is for our node and flow tables :)

It is a basic slab data structure (= allocation pool indexed by integer).  However, it supports concurrent access between readers and writers with _very_ low cost read accesses, by leveraging support of an RCU structure (or similar coordination provided by the caller).


cslab/src/lib.rs
```diff
@@ -0,0 +1,916 @@
...
+// This is the role of `RcuCslab` -- so named as it is a
+// read-copy-update/RCU-friendly structure.
+//
+// The design of `RcuCslab` is as follows.  The `RcuCslab` structure is the
+// "writer" interface.  It provides fully synchronized read-write access to
+// the slab.  Reader handles may be cloned from this.  Each reader handle is
+// associated (_n_-to-1) with a "generation" -- reads performed in a given
+// generation definitively cannot see items removed prior to the creation of
+// the generation.  Generations are totally ordered.  New generations are
+// created explicitly by calling `.collect()` on the writer.  Removals are
```
cpacejo: `.schedule_finalization()`, not `.collect()`.  (Old name.)

mkolehmainen (approved):
This looks good to me.  Would be nice to see an example documentation section for the rcuslab, but actually the comments in the file are pretty good.

PR #345, opened 2024-08-20
capture test to workflow

owenrhysthomas:


.github/workflows/rust-adapter.yml
```diff
@@ -101,6 +101,20 @@ jobs:
...
+    steps: 
+    - uses: actions/checkout@v4
+    - name: Check test warnings
+      run: |
+        cd adapter/ph
+        cargo build
+        sudo apt update
+        sudo apt install libpcap-dev
+        cargo build --manifest-path ../ph-debug/Cargo.toml
+        test/capture-test.sh
```
nonrenewable: potentially too naive question you may have already discussed this with someone else -- does it make sense to put this capture test as part of the "test" execution for ph-debug? 

i ask because the test is capture-related but it's actually in the root `ph` dir but build `ph-debug` a basis so feels like a test for `ph-debug-test` section? (looking below, i'd probably add it in the `ph-debug-test-warnings` job but i'm curious whether you think that makes sense!)

owenrhysthomas: That's an interesting point. We have to build both the ph and ph-debug because the test uses both the ph and then connects the ph-debug into the ph. It is testing the ability of the ph to send commands to the ph, but it is also testing the ability of the ph to properly set the program and filter programs. Theoretically it is testing both the ph and ph-debug, so it's in a weird in between place. It could make sense to add it below the ph-debug section and add ph-debug to the needs section, to reflect that it's really testing both.

nonrenewable: oh i gotcha, i see now it's building both up there! 

maybe name it somehow to make that "common" testing clear/make it more generic than capture (so if someone wants to add more common tests to both here they can easily)? and then you also `needs` the ph-debug-build (which i should've noticed before 😂 ) but i gotcha, i'm here for it!

nonrenewable (approved):
Minor notes but on board with goals now so approving (feel free to merge as you feel ready)

PR #347, opened 2024-08-20
added pkt.advance to different branches of conditionals, altered cap …

owenrhysthomas:
...test

This PR does two things:
- I altered the capture test to actually make sure that the program is filtering correctly: now the program should only capture inbound packets, but the tcpdump will count the number of inbound and outbound packets, to make sure packets are being properly filtered
- I also added pkt.advance to two places in the maybe_capture function where a `match` or conditional could cause the function to reach a place where the link-layer header was not removed, even though one had been added.


cpacejo (approved):
Awesome good catch!!

PR #344, opened 2024-08-20
Fuzz testing for filter and validate

owenrhysthomas:

PR #348, opened 2024-08-20
cbpf-rs: fix bugs found thanks to fuzz testing!

cpacejo:
@owenrhysthomas's fuzz testing work (#344) has paid off!!  Running on my laptop found a bunch of bugs within 5 minutes.  All relating to opcode handling differences between validation and filtering.  This PR contains fixes for them.

Also this fixes a bug I found by observation, that the validator incorrectly rejects programs which end in `ret A`.

Running now for about 1/2 hour fuzzing finds no more issues, so I think we're good now?

For reference: `cargo fuzz run filter` should run fuzz testing for packet filtering.

(Note, excepting the `ret A` bug, all of these were also present in the `libpcap` source I modeled this after.)

Once ZPR is open-sourced we will have to add this to the `rust-fuzz` [trophy case](https://github.com/rust-fuzz/trophy-case) Owen found :)

### DIFF
--- a/cbpf-rs/.gitignore
+++ b/cbpf-rs/.gitignore
@@ -1,1 +1,3 @@
-/target
+target
+fuzz/corpus
+fuzz/target
--- a/cbpf-rs/src/lib.rs
+++ b/cbpf-rs/src/lib.rs
@@ -527,7 +527,10 @@ impl BpfProgram {
             }
         }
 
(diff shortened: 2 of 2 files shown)


mkolehmainen (approved):
I can't believe I didn't notice these errors on my initial review ;)

Looks good!

PR #350, opened 2024-08-21
hopefully added A2A SAID and MAC at end of pkts

### LINKED ISSUE #349: temporary data at end of packets to appease dissector
So if you take the route of fixing the packet format in the PH itself, the functions you'll want to modify are `agent_input()` and `agent_output()` in the `fastpath` module.  Add the A2A SAID and MICV in `agent_input()` before forwarding, and remove them in `agent_output()` before sending the packet.  For the SAID (one byte as a header) just use 0 for now.  For the hash (four bytes at the end), you could put all 0s, or if you want to be fancy, use `blake3::hash()`: https://docs.rs/blake3/latest/blake3/fn.hash.html (and just use the first 4 bytes of the hash) which is what we'll be using eventually anyway! :slightly_smiling_face:

owenrhysthomas:


cpacejo (approved):
Perfect thank you!!

PR #355, opened 2024-08-22
Adapter lookup tables

cpacejo:
Also promulgate `FiveTuple` type.

Also add hack to prepopulate the tables.

Also temporarily disable IPv6 integration test (hack doesn't support IPv6).

### DIFF
--- a/adapter/ph/src/adapter_tables.rs
+++ b/adapter/ph/src/adapter_tables.rs
@@ -8,21 +8,21 @@ use crate::defs::FiveTuple;
 use crate::zpr::{CompressionMode, StreamId};
 use dashmap::DashMap;
 
-pub struct AgentPep {
-    pub stream_id: StreamId,
+pub struct AltPep {
     pub compression_mode: CompressionMode,
+    pub stream_id: StreamId,
 }
 
 pub struct AgentLookupTable {
-    table: DashMap<FiveTuple, AgentPep>,
+    table: DashMap<FiveTuple, AltPep>,
 }
 
-pub struct Ref<'a>(dashmap::mapref::one::Ref<'a, FiveTuple, AgentPep>);
+pub struct AltRef<'a>(dashmap::mapref::one::Ref<'a, FiveTuple, AltPep>);
 
-impl std::ops::Deref for Ref<'_> {
-    type Target = AgentPep;
+impl std::ops::Deref for AltRef<'_> {
+    type Target = AltPep;
 
-    fn deref(&self) -> &AgentPep {
+    fn deref(&self) -> &AltPep {
         self.0.deref()
     }
 }
@@ -34,16 +34,56 @@ impl AgentLookupTable {
         }
     }
 
-    pub fn get<'a>(&'a self, key: &FiveTuple) -> Option<Ref<'a>> {
-        self.table.get(key).map(|r| Ref(r))
+    pub fn get<'a>(&'a self, key: &FiveTuple) -> Option<AltRef<'a>> {
+        self.table.get(key).map(|r| AltRef(r))
     }
 
     // FIXME: ideally we want `try_insert()` but dashmap doesn't support that…
-    pub fn insert(&self, key: FiveTuple, value: AgentPep) {
+    pub fn insert(&self, key: FiveTuple, value: AltPep) {
         self.table.insert(key, value);
     }
 
     pub fn remove(&self, key: &FiveTuple) {
+        self.table.remove(key);
+    }
+}
+
+pub struct DltPep {
+    pub compression_mode: CompressionMode,
+    pub five_tuple: FiveTuple,
+}
+
+pub struct DockLookupTable {
+    table: DashMap<StreamId, DltPep>,
+}
+
+pub struct DltRef<'a>(dashmap::mapref::one::Ref<'a, StreamId, DltPep>);
+
+impl std::ops::Deref for DltRef<'_> {
+    type Target = DltPep;
+
+    fn deref(&self) -> &DltPep {
+        self.0.deref()
+    }
+}
+
+impl DockLookupTable {
+    pub fn new() -> Self {
+        Self {
+            table: DashMap::new(),
+        }
+    }
+
+    pub fn get<'a>(&'a self, key: &StreamId) -> Option<DltRef<'a>> {
+        self.table.get(key).map(|r| DltRef(r))
+    }
+
+    // FIXME: ideally we want `try_insert()` but dashmap doesn't support that…
+    pub fn insert(&self, key: StreamId, value: DltPep) {
+        self.table.insert(key, value);
+    }
+
+    pub fn remove(&self, key: &StreamId) {
         self.table.remove(key);
     }
 }
--- a/adapter/ph/src/adapter_tables.rs
+++ b/adapter/ph/src/adapter_tables.rs
@@ -1,30 +1,26 @@
-//! Agent Lookup Table
+//! Adapter lookup tables
 //!
 //! RFC 6.5 § 5.1
 
 #![allow(dead_code)]
 
 use crate::defs::FiveTuple;
-use crate::zpr::{CompressionMode, StreamId};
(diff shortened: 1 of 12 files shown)


adapter/ph/src/packet.rs
```diff
@@ -153,51 +154,51 @@ pub struct PacketMetadata {
     offset: usize,    // packet offset (must be >= PACKET_BODY_BUFFER_MIN_OFFSET)
     len: usize,       // packet length
     pub flow_id: u32, // flow ID for load-balancing purposes; not otherwise meaningful
-    src_address: IpAddress,
-    dst_address: IpAddress,
-    src_port: u16,
-    dst_port: u16,
-    protocol: u8,
-    _padding: [u8; 7],
+    five_tuple: FiveTuple,
```
nonrenewable, line 157: this makes me so very happy! 

naive question re: padding below - did it lose a bit? and why? (i'm sure i should know and have forgotten...)

adapter/ph/test/common_funcs.sh
```diff
@@ -89,8 +89,9 @@ function ping_test() {
   sudo ip netns exec zpr-a ping -q -c 5 -w 5 "$B_ZPR_ADDR" & wait -f $!
   sudo ip netns exec zpr-b ping -q -c 5 -w 5 "$A_ZPR_ADDR" & wait -f $!
 
-  sudo ip netns exec zpr-a ping -q -c 5 -w 5 "$B_ZPR_ADDR6" & wait -f $!
```
nonrenewable, line 92: naive and irrelevant question but since i'm here - count is 5 but it counts as 6 messages? initial exchange or something? curious how the count above is affected exactly - we can also just talk about this if easier than typing!

adapter/ph/src/zpr.rs
```diff
@@ -40,3 +40,23 @@ pub const KM_ID_IKEV2: KmId = 1;
...
+#[repr(u8)]
+pub enum L3Type {
+    IPv4 = 4,
+    IPv6 = 6,
+}
+
+/// Bitmask indicating how an agent packet is compressed.
+pub type CompressionMode = u8;
+
+/// CompressionMode constants (RFC 6.5 § 6.3.11)
```
nonrenewable, line 55: love ref to RFCs ✊

adapter/ph/src/net_defs.rs
```diff
@@ -37,18 +37,55 @@ impl IpAddress {
...
+pub type IpVersion = u8;
+
+pub fn ip_version(pkt: &[u8]) -> IpVersion {
     pkt[0] >> 4
 }
 
-pub fn ip_ethertype(ip_version: u8) -> u16 {
+pub fn ip_ethertype(ip_version: IpVersion) -> u16 {
     match ip_version {
         4 => ethertype::IP,
```
nonrenewable, line 81: Should this leading numbers reference the zpr.rs L3Type enum values or is the context here different?

adapter/ph/src/main.rs
```diff
@@ -166,6 +167,36 @@ fn main() -> ExitCode {
...
+                    tun_devs[0].destination().unwrap().into(),
+                    protocol,
+                    0,
+                    0,
+                );
+                alt.insert(
+                    five_tuple,
+                    adapter_tables::AltPep {
+                        l3_type: zpr::L3Type::IPv4,
+                        compression_mode: 0,
```
nonrenewable, line 186: is 0 here always an option? or also just a temp stop-gap?

otherwise was petty going to offer to add it in the enum for compression modes

adapter/ph/src/fastpath.rs
```diff
@@ -372,15 +373,30 @@ pub fn substrate_ingress<'pktbuf>(
...
+    stream_id: zpr::StreamId, // TODO: should we keep this in metadata? or per-flow header?
     mut pkt: Packet<'pktbuf>,
 ) {
-    // TODO: decompress
+    // lookup stream ID in DLT
+    if asm
+        .dlt
+        .inspect(stream_id, |pep| {
+            // decompress
+            if pep.compression_mode != 0 {
```
nonrenewable, line 384: asked a q somewhere else, temporary `0` compression mode is probably the answer... i'm going places 😂

adapter/ph/src/main.rs
nonrenewable: i've kind of answered some of this myself - it seems that for now you put some values in here for testing (i know you said that and there's "hack" notes everywhere but i was missing the forest for the trees) but then later on, these will be simply initialized here and then dynamically populated based on interactions with the node/agent requests?

adapter/ph/src/defs.rs
```diff
@@ -8,3 +11,55 @@ pub enum Direction {
...
 }
+
+/// IP 5-tuple used for hashing.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, AsBytes, FromBytes, FromZeroes)]
+#[repr(C)]
+pub struct FiveTuple {
+    pub src_address: net_defs::IpAddress,
+    pub dst_address: net_defs::IpAddress,
+    pub protocol: net_defs::IpProtocol,
+    pub padding: [u8; 1],
```
nonrenewable, line 22: oh maybe this is that stolen bit from the other place?? what is it? 😁

adapter/ph/src/adapter_tables.rs
```diff
@@ -0,0 +1,92 @@
...
+    }
+}
+
+pub struct DltPep {
+    pub l3_type: L3Type,
+    pub compression_mode: CompressionMode,
+    pub five_tuple: FiveTuple,
+}
+
+pub struct DockLookupTable {
```
nonrenewable, line 57: I want to ask you a little more about this data struct in person if you don't mind - i just lack some of the technical depth here that i don't even know how to express exactly and talking would waste way less of your time! (whenever, not urgent at all)

nonrenewable (approved):
Looks good and all of this makes sense! I think i even answered most of my own gaps of knowledge along the way which was nice - means it all ties together ok 😁 

(Note: Def didn't look for performance choices due to my lack of ability to at this point!)

adapter/ph/src/defs.rs
cpacejo: Yes exactly it migrated here! :)  (`zerocopy` insists that everything be explicitly padded, so that is needed here to properly align the `port` fields.  The fields were in a different order before so it implicitly lived at the end of these five fields.)

adapter/ph/test/common_funcs.sh
cpacejo: One of the messages is the `HELLO` we exchange! :)

adapter/ph/src/net_defs.rs
cpacejo: Different context!  (These are IPv4/v6 header version numbers.  I should define these as constants…)

adapter/ph/src/main.rs
cpacejo: Compression modes is a bitmask not an enum :)  So yes 0 is implicitly always valid! :)

adapter/ph/test/common_funcs.sh
nonrenewable: ah, excellent!

adapter/ph/src/net_defs.rs
nonrenewable: i gotcha, thanks for clarifying!

adapter/ph/src/main.rs
nonrenewable: hah great point, slightly missed the forest for the trees

adapter/ph/src/adapter_tables.rs
nonrenewable: for posterity and future me - the look up here is simply the reverse as above (stream id -> 5 tuple) and the part i was missing is that the 5 tuple is all that's necessary to be able to route this packet back to the agent itself!

PR #357, opened 2024-08-22
Fix ph instances not tearing down at the end of integration tests

svjaffer:


adapter/ph/test/basic-test.sh
```diff
@@ -86,8 +86,9 @@ fi
 # Cleanup
 #
 
-kill "${CHILDREN[@]}" 2> /dev/null || true
+sudo kill "${CHILDREN[@]}" || true
 sleep 1  # FIXME: let's do something better here
+stty sane
```
owenrhysthomas, line 92: || true? for GitHub

owenrhysthomas (approved):

PR #365, opened 2024-08-26
actually compress/expand packets (L3 only for now)

### LINKED ISSUE #364: Packet compression / expansion

cpacejo:
also:
* move L3 type into FiveTuple
* fill L3 type in classifier
* L3Type is now an open_enum
* rename 5t protocol -> l4_protocol

### DIFF
--- a/adapter/ph/src/adapter_tables.rs
+++ b/adapter/ph/src/adapter_tables.rs
@@ -6,7 +6,7 @@
 
 use crate::defs::FiveTuple;
 use crate::rcu::RcuBox;
-use crate::zpr::{CompressionMode, L3Type, StreamId};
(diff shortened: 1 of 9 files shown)


mkolehmainen (approved):
This looks good to me.

Did you consider and dismiss using ethertype as the L3_Type values?  I know it's two bytes instead of one.

cpacejo:
> Did you consider and dismiss using ethertype as the L3_Type values? I know it's two bytes instead of one.

I did!  I don't feel strongly either way.  The values I chose here were to match the "IP Version" field of the Bind Request message.  (I guess I could have called the type `IpVersion`.)  So, allows us to use the enum directly in that structure.

PR #366, opened 2024-08-26
Generate and exchange symmetric keys (and ZPIs) during KM handshake

### LINKED ISSUE #339: Iterate on Noise KM - add transit packet "hmac" ability
We have a general idea that we will add an hmac value to a transit packet based on the current SA.  So stop short of encryption, but add some crypto value that can be checked hop by hop.

Will require a little work on the KM to expose the correct functions.

Plan of attack -- expose a key value that another module can use to perform the HMAC.

Also, looks like we do not need to partially encrypt a buffer, so at same time I plan to remove the length encoding feature in the KM.  That gets us 2 bytes back.  

As part of this will also communicate our chosen pair of ZPIs to peer, and accept a pair back.

mkolehmainen:
Uses the ability of our noise pattern to send a message along with the handshake to have both sides tell the other what symmetric key to use for HMACs, and what ZPIs to use.

Another small improvement here makes use of the stateless encrypt/decrypt functions in the noise library we are using.  This means we don't need to hold a lock during encrypt/decrypt.  This is still not as lock free or efficient as possible.

### DIFF
--- a/adapter/ph/src/km.rs
+++ b/adapter/ph/src/km.rs
@@ -168,18 +168,18 @@ impl KeyManager<'_> {
         state.sa_id
     }
 
-    /// Returns the ZPIs used for sending messages, format is (FULL_ENCRYPTION_ZPI, TRANSIT_HMAC_ZPI).
+    /// Returns the ZPIs used for sending messages.
     /// Only set/valid after handshake is complete and SA is established.a
     /// Currently safe to cache this after handshake is done because (TODO) these do not change.
-    pub fn get_send_zpis(&self) -> (u8, u8) {
+    pub fn get_send_zpis(&self) -> ZPIPair {
         let state = self.shared.state.lock().unwrap();
         state.ts.send_zpis
     }
 
-    /// Returns the ZPIs used for receiving messages, format is (FULL_ENCRYPTION_ZPI, TRANSIT_HMAC_ZPI)
-    /// Only set/valid after handshake is complete and SA is established.a/// 
+    /// Returns the ZPIs used for receiving messages.
+    /// Only set/valid after handshake is complete and SA is established.a///
     /// Currently safe to cache this after handshake is done because (TODO) these do not change.
-    pub fn get_recv_zpis(&self) -> (u8, u8) {
+    pub fn get_recv_zpis(&self) -> ZPIPair {
         let state = self.shared.state.lock().unwrap();
         state.ts.recv_zpis
     }
@@ -199,7 +199,7 @@ impl KeyManager<'_> {
         let state = self.shared.state.lock().unwrap();
         state.ts.recv_hmac_key
     }
-    
+
 
     #[cfg(test)]
     fn set_sa_id(&self, sa_id: u8) {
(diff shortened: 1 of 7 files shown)


adapter/ph/src/km_noise.rs
```diff
@@ -113,6 +150,55 @@ pub fn derive_public_key(private_key: &[u8; 32]) -> [u8; 32] {
...
+    snow_state: snow::StatelessTransportState,
+}
+
+impl Codec for NoiseCodec {
+    fn encrypt_transport_stateless(
+        self: &Self,
+        payload: &[u8],
+        message: &mut [u8],
+    ) -> Result<usize, KMError> {
+        rand_bytes(&mut message[..NOISE_NONCE_LEN]).unwrap();
```
cpacejo, line 163: I'd suggest using a [shared atomic counter](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU64.html) here instead (kept in `NoiseCodec`)… it will be faster and equally (actually slightly more) correct.  (Sequential nonces are fine, just reuse is bad.)  But this is fine & correct for the demo.

adapter/ph/src/km.rs
```diff
@@ -154,58 +156,100 @@ impl KeyManager<'_> {
...
+        let codec: Arc<dyn Codec>;
+        {
+            // TODO: Here we take a quick lock to get the state. This would be better as an RLock
+            //       or really should use the RCU thing.
+            let state = self.shared.state.lock().unwrap();
+            if state.sa_id == 0 {
+                return Err(KMError::SaIdZero);
+            }
+
+            if !matches!(state.statemachine.get_state(), KMSMState::Transport { .. }) {
```
cpacejo, line 241: Oh nice I forgot this macro exists!

adapter/ph/src/km.rs
```diff
@@ -215,34 +259,22 @@ impl KeyManager<'_> {
...
             Ok(len) => {
                 info!(
                     "noise: encrypt input {} bytes, output {} bytes",
                     encr_len, len
                 );
-                // Copy the encrypted data back into the message -- there should be sufficient room for it since
-                // caller should know our required padding space and alignment.
-
-                message.shrink(message.body().len()); // remove body
+                message.shrink(encr_len); // remove body, leavign ZPI
```
cpacejo: You'll have a merge conflict, I renamed this to `shrink_by`, sorry!

adapter/ph/src/km.rs
```diff
@@ -215,34 +259,22 @@ impl KeyManager<'_> {
...
             ZdpPacketType::TransitPacket => {
-                panic!("Transit packet encryption not implemented");
+                return Err(KMError::InvalidPacketType);
             }
-            _ => message.body().len(),
+            _ => message.body().len() - 1,
         };
 
         // TODO: Ability to encrypt in place. Not sure how to accomplish. At very least we could use our own buffer pool.
         let mut encr_buf = [0u8; config::PACKET_BUFFER_SIZE];
```
cpacejo, line 268: We'll need to be careful of max-size packets here, adding padding + nonce could result in a too-large ciphertext, overflowing this buffer (causing a crash).

Ideally I think the codec has two functions, one which simply computes the size of the ciphertext, the other which actually performs the encryption.  So the first function can be called first, and then we can check whether there's enough space in the buffer and handle it gracefully if there isn't.

That said – the buffer is probably large enough in all cases here – since 8 byte nonce + 0 padding is less than the size of the metadata at the start of each packet.  So no need to worry about that now.

BUT, below when calling `put`, that will panic if there's not enough space in the packet, which there might not be.  You can check by calling `remaining_mut()` on the packet.

mkolehmainen: My idea was that caller would check `KeyManagerStateMachine::get_setting::padlen` which is the amount of space needed for ciphertext expansion.

mkolehmainen: So if you are using an encryption scheme, you better know how much space it needs and allocate your buffers accordingly.

svjaffer (approved):

PR #367, opened 2024-08-26
A2A checksumming

### LINKED ISSUE #368: compute+check A2A MICV

cpacejo:
Actually generate/validate A2A checksum… almost.  This uses an _unkeyed_ BLAKE3 hash which of course gives no security against an adversary.  This is just an exercise in making sure the code to generate the checksum is in the right place and doing the right thing.

Also rename `Packet::shrink()` to `Packet::shrink_by()` because I forgot what it did.

### DIFF
--- a/adapter/ph/src/fastpath.rs
+++ b/adapter/ph/src/fastpath.rs
@@ -15,11 +15,11 @@ use crate::queues::TryEnqueueError;
 use crate::zdp;
 use crate::zdp_ll;
 use crate::zpr;
+use blake3;
 use bytes::{Buf, BufMut};
-use std::mem::size_of;
 use std::net::SocketAddr;
 use std::time::SystemTime;
-use zerocopy::{AsBytes, FromBytes};
+use zerocopy::FromBytes;
 use zpr_ext::std::mem::{drop_guard, DropGuard};
 use zpr_ext::zerocopy::*;
 
@@ -237,7 +237,7 @@ pub fn decrypt<'pktbuf>(
             return Err(DecryptError::MicvFailure);
         }
 
-        pkt.shrink(2); // remove checksum
+        pkt.shrink_by(2); // remove checksum
 
         Ok(())
     } else {
@@ -377,9 +377,25 @@ pub fn agent_input<'pktbuf>(
     stream_id: zpr::StreamId, // TODO: should we keep this in metadata? or per-flow header?
     mut pkt: Packet<'pktbuf>,
 ) {
-    // "Check" A2A checksum
-    pkt.advance(size_of::<zdp::ZdpSaidHeader>());
-    pkt.shrink(size_of::<zdp::ZdpMicvEnd>());
+    // extract A2A MAC
+    let Some(a2a_hdr) = zdp::ZdpA2aHeader::read_from_buf(&mut pkt) else {
+        drop_and_count(asm, pkt, CounterType::BadStructure);
+        return;
+    };
+
+    if a2a_hdr.a2a_said != 0 {
+        todo!("A2A SAID");
+    }
+
+    let a2a_mac_size = zdp::ZDP_A2A_MAC_SIZE; // TODO: checksum may be shorter depending on A2A SA
+
+    if pkt.body().len() < a2a_mac_size {
+        drop_and_count(asm, pkt, CounterType::BadStructure);
+        return;
+    }
+    let mut a2a_mac = [0u8; zdp::ZDP_A2A_MAC_SIZE];
+    a2a_mac[..a2a_mac_size].copy_from_slice(&pkt.body()[pkt.body().len() - a2a_mac_size..]);
+    pkt.shrink_by(a2a_mac_size);
 
     // lookup PEP in DLT and expand compressed packet
     if asm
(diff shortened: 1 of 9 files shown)


adapter/ph/src/fastpath.rs
```diff
@@ -377,9 +377,25 @@ pub fn agent_input<'pktbuf>(
...
+    }
+
+    let a2a_mac_size = zdp::ZDP_A2A_MAC_SIZE; // TODO: checksum may be shorter depending on A2A SA
+
+    if pkt.body().len() < a2a_mac_size {
+        drop_and_count(asm, pkt, CounterType::BadStructure);
+        return;
+    }
+    let mut a2a_mac = [0u8; zdp::ZDP_A2A_MAC_SIZE];
+    a2a_mac[..a2a_mac_size].copy_from_slice(&pkt.body()[pkt.body().len() - a2a_mac_size..]);
```
nonrenewable, line 397: wow this line threw me for a loop...i think i'm finally following but just to confirm
1. what you're grabbing is the ending bits of length a2a_mac_size from the packet body (edit: just looked below, looks like it's the last thing added on before you forward so that makes sense why you pick it off the end...)
2. you copy that into a2a_mac but use a fixed length of where to copy since that a2a_mac_size might be != zdp::ZDP_A2A_MAC_SIZE per your note above? (edit: this actually happens a lot so i feel like there's something i'm missing re: specific length copies you're doing!)

adapter/ph/src/fastpath.rs
```diff
@@ -431,34 +454,42 @@ pub fn agent_output<'pktbuf>(asm: &Assembly<'pktbuf>, mut pkt: Packet<'pktbuf>)
         }
     }
 
+    // compute A2A MAC
+    // TODO: use actual A2A SAID & keyed hash
+    let a2a_said: zpr::A2aSaid = 0;
+    let a2a_mac_size = zdp::ZDP_A2A_MAC_SIZE; // TODO: may be smaller depending on A2A SAID
+    let mut a2a_mac = [0u8; zdp::ZDP_A2A_MAC_SIZE];
+    // SECURITY: truncating BLAKE3 is safe
+    a2a_mac[..a2a_mac_size].copy_from_slice(&blake3::hash(pkt.body()).as_bytes()[..a2a_mac_size]);
```
nonrenewable, line 463: could the blake3 hash result be < zdp::ZDP_A2A_MAC_SIZE?

tools/zdp_dissector.lua
```diff
@@ -79,10 +79,10 @@ function zdp_proto.dissector(buffer, pinfo, tree)
         -- Transit Packet
         zdp_header_subtree:add(stream_id, buffer(5, 4))
         -- zdp_header_subtree:add(pad, buffer(9, 8))
-        zdp_header_subtree:add(d2d_said, buffer(9, 1))
+        zdp_header_subtree:add(a2a_said, buffer(9, 1))
         zdp_header_subtree:add(agent_packet, buffer(10, real_len - TRANSIT_NON_AGENT_DATA))
-        zdp_header_subtree:add(d2d_mac, buffer(real_len - 8, 4))
-        zdp_header_subtree:add(mac_addr, buffer(real_len - 4, 4))
+        zdp_header_subtree:add(a2a_mac, buffer(real_len - 12, 8))
```
nonrenewable, line 84: question - i put 2 and 2 together that because we're adding from the end and the a2a mac is 8 bits, these have changed. What i don't quite follow is how it was 4 bits before when looking at the ZdpMicvEnd it's 32bits and that's what we seem to have replaced?

nonrenewable (approved):
I think this is fine! just some questions

adapter/ph/src/fastpath.rs
cpacejo: Yes exactly!  I'll clarify this post-demo when I rework this to be a function of the A2A SAID.

adapter/ph/src/fastpath.rs
cpacejo: BLAKE3, no – it's exactly 32 bytes: https://docs.rs/blake3/latest/blake3/struct.Hash.html#method.as_bytes

(Line 460 should really be… `let a2a_mac_size = <MAC size specified by the A2A SA>`, which would always be a value compatible with the SA's hashing function.  This will be clarified when we parameterize this by A2A SA…)

tools/zdp_dissector.lua
cpacejo: The numbers above are bytes!  So we're replacing a 4 byte (= 32 bit) field with an 8 byte field.

adapter/ph/src/fastpath.rs
nonrenewable: oh good ✔️ all good then!

adapter/ph/src/fastpath.rs
nonrenewable: got it! i figured you had a plan, was just behind on it 😁

tools/zdp_dissector.lua
nonrenewable: oh i get the original then, you might have to walk me through the getting of 8bytes with the new mac but that's not urgent

PR #370, opened 2024-08-27
handle bind request in adapter

cpacejo:
Also:
* relocate mgmt handling stuff into mgmt module
* thread ingress link ID through mgmt code
* eliminate most panics from mgmt handling path

### DIFF
--- a/adapter/ph/src/mgmt.rs
+++ b/adapter/ph/src/mgmt.rs
@@ -6,7 +6,8 @@ use crate::fastpath;
 use crate::packet::{self, Packet};
 use crate::zdp;
 use crate::zpr;
-use bytes::BufMut;
+use bytes::{Buf, BufMut};
+use zerocopy::FromBytes;
 
 /// Send a unidirectional non-flow management message on the given link.
 /// The packet should contain only the message body.
@@ -78,4 +79,60 @@ pub async fn send_discard<'pktbuf>(asm: &'pktbuf Assembly<'pktbuf>, link_id: zpr
     let buf = asm.buffer_stack.get_buffer().await;
     let pkt = Packet::new(buf, config::DEFAULT_MESSAGE_HEADROOM);
     send_non_flow_mgmt(asm, link_id, zdp::ZdpPacketType::Discard, pkt).await;
+}
+
+pub async fn handle_report<'pktbuf>(
+    asm: &Assembly<'pktbuf>,
+    ingress_link_id: zpr::LinkId,
+    mut pkt: Packet<'pktbuf>,
+) {
+    let hdr = zdp::ZdpReportHeader::ref_from_prefix(pkt.body()).expect("too-short inbound packet");
+    // TODO handle protocol errors i.e. if the body is shorter
+    let report_data_length: usize = hdr.report_data_length.into();
+    pkt.advance(std::mem::size_of::<zdp::ZdpReportHeader>());
+    if pkt.body().len() >= report_data_length {
+        // TODO printing to stderr blocks indefinitely, this is just temporary
+        eprintln!(
+            "{}: {}",
+            ingress_link_id,
+            std::str::from_utf8(&pkt.body()[..report_data_length]).unwrap()
+        );
+    }
+    asm.buffer_stack.put_buffer(pkt.destroy());
+}
+
+pub async fn handle_discard<'pktbuf>(
+    asm: &Assembly<'pktbuf>,
+    ingress_link_id: zpr::LinkId,
+    pkt: Packet<'pktbuf>,
+) {
+    // TODO print to debug log, when implemented
+    eprintln!("Discard message received from {}", ingress_link_id);
+    asm.buffer_stack.put_buffer(pkt.destroy());
+}
+
+pub async fn handle_hello_request<'pktbuf>(
+    asm: &Assembly<'pktbuf>,
+    ingress_link_id: zpr::LinkId,
+    pkt: Packet<'pktbuf>,
+) {
+    let mut send_pkt = Packet::new(pkt.destroy(), config::DEFAULT_MESSAGE_HEADROOM);
+    let hdr = send_pkt.alloc_zeroed_header::<zdp::ZdpHelloResponseHeader>();
+    hdr.status = 0.into();
+    send_non_flow_mgmt(
+        asm,
+        ingress_link_id,
+        zdp::ZdpPacketType::HelloResponse,
+        send_pkt,
+    )
+    .await;
+    eprintln!("Received HelloRequest");
+}
+
+pub async fn handle_bind_agent_address_request<'pktbuf>(
+    _asm: &Assembly<'pktbuf>,
+    _ingress_link_id: zpr::LinkId,
+    _stream_id: zpr::StreamId,
+    mut _pkt: Packet<'pktbuf>,
+) {
 }
--- a/adapter/ph/src/mgmt.rs
+++ b/adapter/ph/src/mgmt.rs
@@ -2,6 +2,7 @@
 
 use crate::assembly::Assembly;
 use crate::config;
+use crate::counters_enum;
 use crate::fastpath;
 use crate::packet::{self, Packet};
 use crate::zdp;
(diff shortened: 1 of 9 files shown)


adapter/ph/src/mgmt.rs
```diff
@@ -79,3 +84,213 @@ pub async fn send_discard<'pktbuf>(asm: &'pktbuf Assembly<'pktbuf>, link_id: zpr
...
+            HandleMgmtError::UnknownType(_type) => Self::UnknownType,
+            HandleMgmtError::UnexpectedMgmtResponse => Self::UnexpectedMgmtResponse,
+            HandleMgmtError::BadStructure => Self::BadStructure,
+        }
+    }
+}
+
+pub type HandleMgmtResult<'pktbuf> = Result<(), (HandleMgmtError, Packet<'pktbuf>)>;
+
+pub async fn handle_report<'pktbuf>(
```
cpacejo, line 106: These all were just moved, had panics changed to error returns, and had `ingress_link_id` threaded through.

adapter/ph/src/mgmt.rs
```diff
@@ -79,3 +84,213 @@ pub async fn send_discard<'pktbuf>(asm: &'pktbuf Assembly<'pktbuf>, link_id: zpr
...
+        zdp::ZdpPacketType::HelloResponse,
+        rsp_pkt,
+    )
+    .await;
+
+    Ok(())
+}
+
+// RFC 6.5 § 6.3.11
+pub async fn handle_bind_agent_address_request<'pktbuf>(
```
cpacejo, line 164: This is the only method here which is actually new.

adapter/ph/src/mgmt_processor_worker.rs
```diff
@@ -46,62 +54,39 @@ async fn handle_packet<'pktbuf>(asm: &Assembly<'pktbuf>, mut pkt: Packet<'pktbuf
...
+
+            ZdpPacketType::BindAgentAddressRequest => {
+                mgmt::handle_bind_agent_address_request(asm, ingress_link_id, stream_id, pkt).await
+            }
+
+            packet_type => Err((HandleMgmtError::UnknownType(packet_type.0), pkt)),
         }
     } else {
         match base_hdr.packet_type {
-            ZdpPacketType::Report => {
```
cpacejo, line 72: These all got moved into the mgmt module.

zpr-ext/src/zerocopy.rs
```diff
@@ -28,14 +32,29 @@ mod zerocopy_bytes {
...
+        }
+
         fn read_from_buf(buf: &mut impl Buf) -> Option<Self>
         where
             Self: Sized,
         {
             if buf.remaining() < std::mem::size_of::<Self>() {
                 None
             } else {
-                Self::read_from(buf.copy_to_bytes(std::mem::size_of::<Self>()).as_ref())
```
cpacejo: Drive-by premature optimization.  I realized `copy_to_bytes()` is performing a heap allocation.  So instead I rewrote this to allocate space for a `T` on the stack.  But that was very roundabout because of how the `zerocopy` traits are defined.

adapter/ph/src/mgmt.rs
```diff
@@ -79,3 +84,213 @@ pub async fn send_discard<'pktbuf>(asm: &'pktbuf Assembly<'pktbuf>, link_id: zpr
...
+    let Some(hdr) = zdp::ZdpReportHeader::ref_from_prefix(pkt.body()) else {
+        return Err((HandleMgmtError::BadStructure, pkt));
+    };
+
+    // TODO handle protocol errors i.e. if the body is shorter
+    let report_data_length: usize = hdr.report_data_length.into();
+    pkt.advance(std::mem::size_of::<zdp::ZdpReportHeader>());
+    if pkt.body().len() >= report_data_length {
+        // TODO printing to stderr blocks indefinitely, this is just temporary
+        eprintln!(
```
mkolehmainen, line 120: Why not use tracing? Eg, `info!`.

adapter/ph/src/mgmt.rs
```diff
@@ -79,3 +84,213 @@ pub async fn send_discard<'pktbuf>(asm: &'pktbuf Assembly<'pktbuf>, link_id: zpr
...
+    if compression_mode & zpr::compression_mode::SOURCE_PORT_PRESENT != 0 {
+        if pkt.remaining() < 2 {
+            return Err((HandleMgmtError::BadStructure, pkt));
+        }
+        src_port = pkt.get_u16();
+    }
+
+    // read destination port (optional)
+    let mut dst_port = 0;
+    if compression_mode & zpr::compression_mode::SOURCE_PORT_PRESENT != 0 {
```
mkolehmainen, line 230: Does `SOURCE_PORT_PRESENT` imply a destination port too?

adapter/ph/src/mgmt.rs
```diff
@@ -79,3 +84,213 @@ pub async fn send_discard<'pktbuf>(asm: &'pktbuf Assembly<'pktbuf>, link_id: zpr
...
+
+    let ip_version = hdr.ip_version;
+    let compression_mode = hdr.compression_mode;
+
+    // read addresses (always present)
+    let src_address;
+    let dst_address;
+    match ip_version {
+        zpr::L3Type::Ipv4 => {
+            let Some(src_addr) = <[u8; 4]>::read_from_buf(&mut pkt) else {
```
mkolehmainen, line 185: Shouldn't we have a header or message struct we can read into instead of reading bytes ourselves?  Wouldn't it be nicer if the addresses were actually ip address types?

adapter/ph/src/mgmt.rs
```diff
@@ -79,3 +84,213 @@ pub async fn send_discard<'pktbuf>(asm: &'pktbuf Assembly<'pktbuf>, link_id: zpr
...
+    // recycle request buffer for response
+    let mut rsp_pkt = Packet::new(pkt.destroy(), config::DEFAULT_MESSAGE_HEADROOM);
+
+    // attempt to insert into DLT
+    let stream_id;
+    match asm.dlt.insert(pep) {
+        Ok(sid) => {
+            // success; respond with stream ID
+            // TODO: maybe tick a counter somewhere?
+            zdp::ZdpBindAgentAddressResponseHeader {
```
mkolehmainen, line 259: Haven't looked at spec, but feels like there should be a sequence number or something here that ties the response to the request.

adapter/ph/src/peer_table.rs
```diff
@@ -7,13 +7,23 @@ use std::sync::Mutex;
...
+
 pub struct PeerState {
+    pub peer_type: PeerType,
     pub substrate_addr: SubstrateAddr,
 }
 
 impl PeerState {
-    pub fn new(substrate_addr: SubstrateAddr) -> Self {
-        Self { substrate_addr }
+    pub fn new(peer_type: PeerType, substrate_addr: SubstrateAddr) -> Self {
```
mkolehmainen, line 22: In the prototype we use separate ports for dock and link connections.  Are we doing that in ph?  If not do we need an unknown type or should we use an Option?

adapter/ph/src/mgmt.rs
cpacejo: I agree, not new code though.

mkolehmainen (approved):
I have some questions/comments for you, but looks good.

adapter/ph/src/mgmt.rs
cpacejo: No, § 6.3.11 of RFC 6 has them as separate.

adapter/ph/src/mgmt.rs
cpacejo: Yeah this bugged me too but we already have a proliferation of "IP address" types.  (Rust's `Ipv4Addr`, `Ipv6Addr`, `IpAddr`, and our own `IpAddress`.)  Maybe I'll define `Ipv4AddressWire` and `Ipv4AddressWire` types which are exactly these wire representations though.

adapter/ph/src/mgmt.rs
mkolehmainen: But it's ok that you are using the same compression mode flag to read both src & dst ports?  `SOURCE_PORT_PRESENT`. ?

adapter/ph/src/mgmt.rs
cpacejo: Ohhhhhhhhhhhhhhhhhhhhhhhhhhhhhh is there some baggage here :)

(Quick summary: we have too many open questions about how sequence numbers are supposed to work.  We need to get Frank in to sort it all out before we can attempt to implement something.)

adapter/ph/src/peer_table.rs
cpacejo: I'm not following, which ports are you referring to?

The `SubstrateAddr` here refers to the remote address:port of the link or docking session.

PR #371, opened 2024-08-27
zpr-ext: eliminate heap allocation for read_from_buf

cpacejo:


nonrenewable (approved):
Yes! I'm here for it!

(from @cpacejo a note that helped me and might help anyone else who ends up here
"`copy_to_bytes()` performs a heap allocation (necessarily, since it returns an owned reference to an arbitrarily sized piece of memory).  To avoid that, instead I allocate space for the read data directly on the stack (as ret), read into that and return it.")

PR #372, opened 2024-08-27
Adapter-Node-Adapter

svjaffer:
- Change the basic integration test to an adapter-node-adapter topology
 - Added node "mode" to the PH
 - Added a second peer_addr argument to the PH.  Node has 2 peer_addrs
 - Node "forwards" between 2 peers (janky temporary hack)
 - Added a system_name argument to the PH for debugging purposes
 - Added packet dump for debugging purposes

### DIFF
--- a/adapter/ph/src/assembly.rs
+++ b/adapter/ph/src/assembly.rs
@@ -44,6 +44,8 @@ use zpr_ext::std::mem::{drop_guard, DropGuard};
 
 pub struct Assembly<'pktbuf> {
     // Shared resources.  These may be accessed by any part of the system.
+    pub system_name: String, // For debugging use
+
     pub buffer_stack: BufferStack<'pktbuf, { config::PACKET_BUFFER_SIZE }>,
 
     pub mgmt_processor: MgmtProcessor<'pktbuf>,
@@ -62,7 +64,7 @@ pub struct Assembly<'pktbuf> {
     pub sync_req_state: SyncReqState<'pktbuf>,
 
     pub peer_table: peer_table::PeerTable,
-    pub adapter_docking_session_id: zpr::LinkId,
+    pub peer_ids: Vec<zpr::LinkId>,
 
     pub alt: adapter_tables::AgentLookupTable,
     pub dlt: adapter_tables::DockLookupTable,
@@ -115,9 +117,10 @@ impl<'pktbuf> Assembly<'pktbuf> {
         &self,
         zdp_request_type: ZdpPacketType,
         zdp_response_type: ZdpPacketType,
+        peer_id: zpr::LinkId,
         pkt_fn: impl Fn(&mut Packet<'_>) + Send + 'static,
     ) -> Result<Packet<'pktbuf>, SyncReqError> {
-        self.send_sync_req_helper(zdp_request_type, zdp_response_type, None, pkt_fn)
+        self.send_sync_req_helper(zdp_request_type, zdp_response_type, None, peer_id, pkt_fn)
             .await
     }
 
@@ -132,10 +135,17 @@ impl<'pktbuf> Assembly<'pktbuf> {
         zdp_request_type: ZdpPacketType,
         zdp_response_type: ZdpPacketType,
         stream_id: u32,
+        peer_id: zpr::LinkId,
         pkt_fn: impl Fn(&mut Packet<'_>) + Send + 'static,
     ) -> Result<(u32, Packet<'pktbuf>), SyncReqError> {
         match self
-            .send_sync_req_helper(zdp_request_type, zdp_response_type, Some(stream_id), pkt_fn)
+            .send_sync_req_helper(
+                zdp_request_type,
+                zdp_response_type,
+                Some(stream_id),
+                peer_id,
+                pkt_fn,
+            )
             .await
         {
             Ok(mut pkt) => {
@@ -161,6 +171,7 @@ impl<'pktbuf> Assembly<'pktbuf> {
         zdp_request_type: ZdpPacketType,
         zdp_response_type: ZdpPacketType,
         stream_id: Option<u32>,
+        peer_id: zpr::LinkId,
         pkt_fn: impl Fn(&mut Packet<'_>) + Send + 'static,
     ) -> Result<Packet<'pktbuf>, SyncReqError> {
         let permit: SemaphorePermit = self.sync_req_state.semaphore.acquire().await.unwrap(); // TODO error handling in case we don't get permit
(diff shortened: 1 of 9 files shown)


adapter/ph/src/assembly.rs
```diff
@@ -240,6 +246,7 @@ impl<'pktbuf> Assembly<'pktbuf> {
             .send_sync_non_flow_req(
                 ZdpPacketType::HelloRequest,
                 ZdpPacketType::HelloResponse,
+                self.peer_ids[0],
```
mkolehmainen: Should peer id index be hard coded here?  Would it make more sense for send_hello_req to take to_peer_id as an argument?

mkolehmainen: Or is this just a hack function of some sort? I notice below that it seems to send and wait for response.  That doesn't seem right...  Why would we expect a hello-response to be the next packet received?

adapter/ph/src/main.rs
```diff
@@ -376,9 +415,11 @@ fn main() -> ExitCode {
...
             }
 
-            mgmt::send_report(asm, asm.adapter_docking_session_id, "Reporting for Duty!").await;
-            mgmt::send_discard(asm, asm.adapter_docking_session_id).await;
-            asm.send_hello_req().await;
+            if !is_node {
+                mgmt::send_report(asm, asm.peer_ids[0], "Reporting for Duty!").await;
+                mgmt::send_discard(asm, asm.peer_ids[0]).await;
+                asm.send_hello_req().await;
+            }
```
mkolehmainen, line 394: Ah I see.  This is where adapter decides to start sending some messages.  I see it is scripted... (and where I should put the KM handshake).

mkolehmainen (approved):
Looks good. I left some comments. Only thing that I would change is the signature for send-hello-req.

adapter/ph/src/assembly.rs
svjaffer: Oh, this is 100% a short-term hack for the demo

adapter/ph/src/main.rs
svjaffer: Yeah, I've got the link state machine code stashed away while we prep for the demo.  That will replace this with the long-term logic.

adapter/ph/src/main.rs
```diff
@@ -174,29 +205,36 @@ fn main() -> ExitCode {
...
+            if !is_node {
+                for (remote_stream_id, protocol) in [1, 6, 17].into_iter().enumerate() {
+                    let five_tuple = defs::FiveTuple::new(
+                        zpr::L3Type::Ipv4,
+                        tun_devs[0].address().unwrap().into(),
+                        if zpr_peer.is_some() {
+                            zpr_peer.unwrap().into()
+                        } else {
+                            tun_devs[0].destination().unwrap().into()
+                        },
```
cpacejo: Obv. not needed here as this is throwaway code, but the Rust-y way to write this is:

```
match zpr_peer {
  Some(peer) => peer.into(),
  None => tun_devs[0].destination().unwrap().into(),
}
```

adapter/ph/src/packet.rs
```diff
@@ -394,6 +414,30 @@ impl<'buf> Packet<'buf> {
...
+                print!("\n{:04x} ", i);
+            } else if i % 8 == 0 {
+                print!(" ");
+            }
+            print!(" {:02x}", self.buf[i]);
+        }
+        println!("");
+    }
+
+    pub fn dump(&self) {
```
cpacejo, line 435: I love this!!!!!!!!  Best would be if you could make it a [`Debug`](https://doc.rust-lang.org/std/fmt/trait.Debug.html) trait impl.  (Easy way: use `write!(f, ...)` instead of `print!(...)`)

adapter/ph/src/packet.rs
```diff
@@ -394,6 +414,30 @@ impl<'buf> Packet<'buf> {
...
+
+    pub fn dump_packet_buffer(&self) {
+        // TODO This is a super hacky and inefficient way to do this
+        for i in 0..self.metadata().len {
+            if i % 16 == 0 {
+                print!("\n{:04x} ", i);
+            } else if i % 8 == 0 {
+                print!(" ");
+            }
+            print!(" {:02x}", self.buf[i]);
```
cpacejo, line 430: `buf` here includes the metadata, but `len` above refers to only the packet body.

Maybe here you mean to reference `self.body()[i]` instead?

adapter/ph/src/packet.rs
cpacejo: Actually I've done this in my own branch (and also addressed the above bug) so I will just merge that!

adapter/ph/src/main.rs
```diff
@@ -92,12 +99,16 @@ fn emit_counts(counts_map: &EnumMap<CounterType, Counter>) {
...
     let sock_path = cmd_line.control_path;
-    let peer_addr = cmd_line.dock_addr;
     let self_addr = cmd_line.self_addr;
+    let peer_addr1 = cmd_line.peer_addr1;
+    let peer_addr2 = cmd_line.peer_addr2;
+    let zpr_peer = cmd_line.zpr_peer;
     let _ca_file = cmd_line.ca_file;
     let _cert_file = cmd_line.certificate_file;
     let _priv_key_file = cmd_line.private_key_file;
+    let mut is_node = false;
```
cpacejo, line 107: You can declare this as `let is_node;` and then initialize it as true/false later on.  The compiler will ensure it's initialized exactly once, and not used before initialization, without allowing further mutation.

adapter/ph/src/main.rs
```diff
@@ -84,7 +86,8 @@ struct CmdLine {
     tun_if: Option<String>,
 }
 
-fn emit_counts(counts_map: &EnumMap<CounterType, Counter>) {
+fn emit_counts(system_name: &String, counts_map: &EnumMap<CounterType, Counter>) {
```
cpacejo, line 89: For passing around read-only string references `&str` is more flexible.  (To which a `String` can be cast using `.as_str()`.)  `str` being a simple view of a string, vs. an owned object.  (Same difference as `&[u8]` vs. `&Vec<u8>`.)

adapter/ph/test/basic-test.sh
cpacejo, line 1: I wonder if it would actually make sense to clone this file, so we have an adapter<->adapter test, and an adapter<->node<->adapter test?  Could help distinguish more "involved" issues from simpler ones.

cpacejo: (That said… maintaining both might become annoying… I think what's here is fine for now!!!!!)

cpacejo:
Closing as this has been incorporated into #379.

PR #373, opened 2024-08-28
async trait fns supported as of Rust 1.75

### LINKED ISSUE #55: rework async trait fns
Rust doesn't support async trait methods yet.  We have some functions that "want to be" though, which are currently just free-functions.  However Tokio uses a trick to make trait methods which "act" asynchronous, e.g. <https://docs.rs/tokio/latest/src/tokio/io/util/async_write_ext.rs.html#124-129>… let's see if we can copy this, would clean up the code a bit.

cpacejo:
cleanup here as I noticed this is now possible


mkolehmainen (approved):
nice!

PR #378, opened 2024-08-29
Hook KM subsystem into the ph 

### LINKED ISSUE #353: hook KM into PH
Put a KM instance on the PeerState in the packet handler.  Figure out correct initialization.

mkolehmainen:
While not actually hooked up per se, this has integrated the KM into the fastpath.   Facility is provided (with small example in main.rs) to add KM protocol to links when they come up and remove it when they go down.  New flags have been added to the binary that allow the KM to be turned off, and also allow for insecure use of ZPI=0.  Note that these flags allow the basic-test to run.

What this does-

- Refactor the KM so that the encry/decrypt stuff is just a codec interface created by the (eg) Noise state machine.
- New KM signal that tells you when an SA is established AND sends all state required to use it.
- Rework encrypt and decrypt in fastpath a bit.
- Some code in main and assembly to set up state for KM, and to illustrate how to add an adapter link.
- Two new "threads" that monitor the link key management machines: one to watch for signals and update assembly state, and one to watch for KM payloads and emit ZDP Key Management messages.
- I have kept the old node and cd code in sync for now.


adapter/ph/src/assembly.rs
```diff
@@ -70,6 +76,25 @@ pub struct Assembly<'pktbuf> {
...
+}
+
+pub struct PhFlags {
+    /// If set TRUE this allows any messages on ZPI 0.  VERY INSECURE!!
+    pub allow_insecure_zpi_zero: bool,
+}
+
+impl PhFlags {
+    /// Reasonable defaults
+    pub fn new() -> Self {
```
cpacejo: You can instead `impl Default for PhFlags { fn default() -> Self { ... } }`.

adapter/ph/src/assembly.rs
```diff
@@ -70,6 +76,25 @@ pub struct Assembly<'pktbuf> {
     pub dlt: adapter_tables::DockLookupTable,
 
     pub adapter_manager: AdapterManager<'pktbuf>,
+    pub km_state: KmState<'pktbuf>,
+
+    // The `pa_states` is written to by the KM Multiplexor and is read whenever we need to figure
+    // out the correct ZPI to use on a link, or do encryption/HMAC.
+    pub sa_states: Arc<DashMap<zpr::LinkId, SAState>>,
```
cpacejo: Let's put the `SAState` into the `peer_table` -- that's the home for all link state.

adapter/ph/src/main.rs
```diff
@@ -288,8 +322,18 @@ fn main() -> ExitCode {
                 alt,
                 dlt,
                 adapter_manager,
+                km_state,
+                sa_states: Arc::new(DashMap::new()),
```
cpacejo: No need for `Arc` here I don't think.  (Though see my other comment about leveraging `peer_state` to store this state.)

adapter/ph/src/main.rs
```diff
@@ -344,6 +388,28 @@ fn main() -> ExitCode {
                 cap_outq,
             ));
 
+
+            // Watch for outbound Key Management messages
+            let km_loop_ctok = Box::leak(Box::new(km_mpx_ctok.clone()));
```
cpacejo: Instead of boxing this, you can move it into the `async` block, just write `async move { }` instead of `async { }`.

adapter/ph/src/main.rs
```diff
@@ -344,6 +388,28 @@ fn main() -> ExitCode {
...
+            let km_loop_rx = Box::leak(Box::new(km_rx));
+            js.spawn(async {
+                loop {
+                    tokio::select! {
+                        _ = km_loop_ctok.cancelled() => {
+                            break;
+                        }
```
cpacejo: What cancels this thread?

(Right now no other worker thread gets cancelled for any reason; the only way "out" is to terminate the process.  Link up/down let's have working independently of process start/stop.  I'm somewhat of the mind to _only_ have terminate-as-shutdown; every product I've worked on with a "clean" shutdown (a) had tons of teardown-synchronization bugs in the clean shutdown path, and (b) never actually relied on the clean shutdown path.)

adapter/ph/src/main.rs
```diff
@@ -344,6 +388,28 @@ fn main() -> ExitCode {
...
+                        Some(km_buf_msg) = km_loop_rx.recv() => {
+                            mgmt::send_key_management(&*asm, km_buf_msg.link_id, zpr::KM_ID_NOISE, &km_buf_msg.msg).await;
+                        }
+                    }
+                }
+            });
```
cpacejo: Let's stick this worker thread in its own module like how the other worker threads are.

adapter/ph/src/km.rs
```diff
@@ -105,11 +107,89 @@ pub enum KMSignal {
...
+    pub recv_zpis: ZPIPair,
+
+    /// These are the ZPIs which our peer has given us to use for sending messages.
+    pub send_zpis: ZPIPair,
+
+    /// This is the key, shared with our peer, which we should use for sending HMAC messages to the peer.
+    pub send_hmac_key: [u8; 32],
+
+    /// The is the key, shared with our peer, which our peer will use to send HMAC messages to us.
+    pub recv_hmac_key: [u8; 32],
```
cpacejo, line 197: Why not use the same key for both sides?

adapter/ph/src/km.rs
```diff
@@ -105,11 +107,89 @@ pub enum KMSignal {
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        if self.sa_id == 0 {
+            return write!(f, "KMTransportSA {{ sa_id: 0 }}");
+        }
+        write!(
+            f,
+            "KMTransportSA {{ sa_id: {}, send_zpis: {:?}, recv_zpis: {:?}, send_hmac_key: {:?}, recv_hmac_key: {:?} }}",
+            self.sa_id, self.send_zpis, self.recv_zpis, self.send_hmac_key, self.recv_hmac_key
+        )
+    }
```
cpacejo: Can just call `Debug::fmt(self, f)` here.

adapter/ph/src/km.rs
```diff
@@ -584,6 +536,135 @@ impl KeyManager<'_> {
...
         Ok(())
     }
+
+    // Helper to reduce verbosity slightly
+    fn send_signal<'a>(
+        &self,
+        chan: &'a mpsc::Sender<KMLinkMsg<KMSignal>>,
+        link_id: zpr::LinkId,
+        signal: KMSignal,
+    ) -> impl Future<Output = Result<(), mpsc::error::SendError<KMLinkMsg<KMSignal>>>> + 'a {
```
cpacejo: Often I find these need to be `+ Send` as well to cooperate with Tokio.

Though, you should be able to write this as an `async fn`, no?

adapter/ph/src/km.rs
```diff
@@ -584,6 +536,135 @@ impl KeyManager<'_> {
...
+            sa_id: 0,
+            send_zpis,
+            recv_zpis,
+            send_hmac_key: send_key,
+            recv_hmac_key: recv_key,
+            codec,
+        }
+    }
+
+    pub fn new_empty() -> KMTransportSA {
```
cpacejo: Maybe implement as `Default::default()`?

adapter/ph/src/km_multiplexor.rs
```diff
@@ -0,0 +1,401 @@
...
+
+/// Global state for the KM system.
+/// TODO: Move to assembly?
+pub struct KmState<'pktbuf> {
+    // This token is used to create cancellation tokens for all the link KeyManager state machines.
+    // So cancellign this will shutdown them all.
+    ctok: CancellationToken,
+    km_tx: mpsc::Sender<KMLinkMsg<Bytes>>,
+    km_sig_tx: mpsc::Sender<KMLinkMsg<KMSignal>>,
+    table: DashMap<zpr::LinkId, KMHandle<'pktbuf>>, // Not sure how to set the lifetime variable here.
```
cpacejo: Yes let's move the `KMHandle` into `peer_state` if possible.  Then this structure just becomes the "public interface" to the KM, call it `KmCtl` or some such.

adapter/ph/src/km_multiplexor.rs
```diff
@@ -0,0 +1,401 @@
...
+use std::future::Future;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use tokio::sync::mpsc;
+use tokio_util::sync::CancellationToken;
+use tracing::{error, info};
+
+/// Global state for the KM system.
+/// TODO: Move to assembly?
+pub struct KmState<'pktbuf> {
```
cpacejo: Consistency nit – here it's `Km` but other places it's `KM`.  I don't have a strong feeling either way (though Rust style seems to lean toward `Km`).

adapter/ph/src/km.rs
```diff
@@ -121,228 +201,64 @@ struct KMShared<'mgr> {
 }
 
 struct KMState<'mgr> {
-    // Warning - have no idea what I'm doing with lifetimes.
-    // What I'm trying to assert here is that the impl of KeyManagerStateMachine must live as long as the KeyManager it is passed to.
+    // Lifetime hint here asserts that the impl of KeyManagerStateMachine must live as long as the KeyManager it is passed to.
     statemachine: Box<dyn KeyManagerStateMachine + 'mgr>,
```
cpacejo: No lifetime annotation is needed here.  `Box<dyn KeyManagerStateMachine>` ensures that the referenced item lives at least as long as this structure and anything that references it.  So I think most/all of the KM structures do not need lifetime parameters.

adapter/ph/src/km_multiplexor.rs
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    fn add_link_handle(&self, link_id: zpr::LinkId, handle: KMHandle<'pktbuf>) {
+        self.table.insert(link_id, handle);
+    }
+}
+
+// I'd like the KeyManager to live as long as the hashmap that holds the handle.
+// Which for us is forever -- since map should be created on KmState and stored
+// forever in the Assembly.
```
cpacejo: See my comment in `km.rs` about lifetime parameters.  I don't think you need any anywhere – default Rust lifetime rules ensure the behavior you want.

adapter/ph/src/km_multiplexor.rs
```diff
@@ -0,0 +1,401 @@
...
+    mgr: KeyManager<'pktbuf>, // The manager must remnain valid for lifetime of the link
+}
+
+/// SAState is placed in the assembly so that other parts of the code can check
+/// to see if an SA is established, and if so get all the details.
+///
+/// Do not use any values in here until `sa_established` is true.
+///
+/// - TODO: Replace this lame synchronization method with RCU ??
+/// - TODO: This could be more tightly integrated into the assembly.
```
cpacejo: Yes – this can go into an `RcuBox` in `peer_table`.

adapter/ph/src/km_multiplexor.rs
```diff
@@ -0,0 +1,401 @@
...
+                    _ => {} // TODO: Handle other signals.
+                }
+            }
+        }
+    }
+}
+
+/// Start the signal watcher worker.
+///
+/// (I've totally copied the pattern that Chris used for other workers but don't really understand it. --mk)
```
cpacejo: I don't understand it all either! :)  (Wrote it while I was still wrapping my head around Rust-isms and haven't revisited since…)

adapter/ph/src/km_multiplexor.rs
```diff
@@ -0,0 +1,401 @@
...
+                    }
+                    KMSignal::SaEstablished(sa) => {
+                            if let Some(mut sa_state) = state_table_p.get_mut(&linkmsg.link_id) {
+                                sa_state.transport_sa = sa;
+                                sa_state.sa_established.store(true, Ordering::Relaxed);
+                            } else {
+                                error!("km_multiplexor: no SA state for link {}", linkmsg.link_id);
+                            }
+                    }
+                    _ => {} // TODO: Handle other signals.
```
cpacejo: There is a `todo!()` macro which panics.

adapter/ph/src/km_multiplexor.rs
```diff
@@ -0,0 +1,401 @@
...
+    if let Some(sa_state) = asm.sa_states.get_mut(&link_id) {
+        sa_state.sa_established.store(false, Ordering::Relaxed);
+    }
+
+    // remove handle from our km state, if found
+    let handle = asm.km_state.table.remove(&link_id);
+
+    // Stop the KM
+    if handle.is_some() {
+        handle.unwrap().1.ctok.cancel(); // stop the KM
```
cpacejo: It may be best if we also wait on the join handle here (returned by `tokio::spawn()`) so we know the KM has stopped "doing stuff".  (Avoids the class of bugs that a KM worker is stuck in some state continuing to "do stuff" long after we think it's dead.)  That of course would make this an async function but I think that's fine/correct.

adapter/ph/src/km_multiplexor.rs
```diff
@@ -0,0 +1,401 @@
...
+    let child_ctok = asm.km_state.ctok.child_token();
+    let spawn_ctok = child_ctok.clone();
+    let spawn_km_tx = asm.km_state.km_tx.clone();
+    let spawn_sig_tx = asm.km_state.km_sig_tx.clone();
+    tokio::spawn(async move {
+        match spawn_mgr.start(spawn_ctok, spawn_km_tx, spawn_sig_tx).await {
+            Ok(_) => (),
+            Err(e) => {
+                error!("KeyManager failed on link {}: {:?}", link_id, e);
+            }
```
cpacejo, line 221: I feel like this should take down the link somehow, e.g. by sending a signal to the link state machine.

(Some of the KM failure modes include "bad packet" which I think should always immediately bring down a link.)

adapter/ph/src/km_multiplexor.rs
```diff
@@ -0,0 +1,401 @@
...
+    Ok(())
+}
+
+/// When an inbound key management ZDP message arrives on a link, send it here after parsing.
+///
+/// The km_payload should be the km-payload part of the ZDP KM message.  We copy the payload before
+/// returning.
+///
+/// TODO: Note this will block if the KeyManager queue is full.  Should this instead be using
+///       the non-blocking call?
```
cpacejo: Depends where we're calling into this from.  If from the `mgmt_processor` worker, we're already in management land, blocking calls are fine and correct (I don't think we should be shedding load in management land).  But if from the fastpath, then yes, this should be a non-async function and use the `try_handle_km_message` method.  Fastpath must never block.

adapter/ph/src/km_multiplexor.rs
```diff
@@ -0,0 +1,401 @@
...
+pub async fn handle_inbound_km_msg<'pktbuf>(
+    asm: &Assembly<'pktbuf>,
+    from_link: zpr::LinkId,
+    km_payload: &[u8],
```
cpacejo, line 252: Unless you have a strong objection I think this should be a `Packet`, rather than making a heap allocation and copying the data into it in `handle_km_message()`.  (The caller certainly has no need for the data any longer.)  No need to change in this PR though; I opened a separate task.

Thinking more – I think the interface is _cleaner_ how you have it.  I'm wondering whether there's a way I can make `Packet` implement `Bytes`.  Hm.............

adapter/ph/src/main.rs
mkolehmainen: Yeah, ok that's fine with me.  That KM thread in particular should run until end of program.

adapter/ph/src/main.rs
mkolehmainen: Will do.  Not sure why I didn't.

adapter/ph/src/assembly.rs
mkolehmainen: on it.

adapter/ph/src/km_multiplexor.rs
```diff
@@ -0,0 +1,401 @@
...
+) -> Result<(), String> {
```
cpacejo: Instead of a `String`, possible to return an enum of some sort here?  That way it can be mapped to a counter type and appropriately counted.

Also I would say that `handle_km_message()` should return a more specific error type than `KMError` – many of the values of `KMError` are things that can only happen while processing the message, whereas only two kinds of errors can occur during enqueueing (queue full, and queue not present).  Best if the caller need only be prepared to handle those cases that actually occur.

(Notably – if we do switch to passing a `Packet` in here, if the enqueue fails, that `Packet` should get passed back so the caller can drop it.  This is how e.g. the `mgmt_processor` queue works.)

adapter/ph/src/main.rs
cpacejo: I know why, because `main` is a mess right now so it doesn't look out of place :)

adapter/ph/src/fastpath.rs
```diff
@@ -181,24 +186,41 @@ pub fn maybe_capture_batch<'a, 'pktbuf: 'a>(
...
+pub fn encrypt_hmac<'pktbuf>(_send_hmac_key: [u8; 32], _pkt: &mut Packet<'pktbuf>) {
+    // Run the blake hash key'd with the `send_hmac_key` on the correct parts of the packet
+    // and write the (truncated) hash value to the correct header location.
+
+    info!("encrypt_hmac: not implemented - NOP");
+}
+
+pub fn encrypt_full<'pktbuf>(
+    asm: &Assembly<'pktbuf>,
+    codec: Arc<dyn Codec>,
```
cpacejo: Let's pass a `&dyn Codec` here.  (No need for this function to be able to adjust reference counts.)

adapter/ph/src/fastpath.rs
```diff
@@ -181,24 +186,41 @@ pub fn maybe_capture_batch<'a, 'pktbuf: 'a>(
...
-    if zpi == zpr::ZPI_0 {
-        // RFC 6.5 § 5.25.2
-        pkt.put(
-            net_defs::inet_checksum(&pkt.body()[std::mem::size_of::<zdp::ZdpZpiHeader>()..])
-                .as_slice(),
-        );
-    } else {
-        todo!("encryption");
+    let enc_buf = asm.buffer_stack.try_get_buffer().unwrap();
+    let encr_len = pkt.body().len() - 1; // Everything except the ZPI byte
```
cpacejo: `std::mem::size_of::<ZdpZpiHeader>()` is a more verbose way to spell "1" :)

adapter/ph/src/fastpath.rs
```diff
@@ -181,24 +186,41 @@ pub fn maybe_capture_batch<'a, 'pktbuf: 'a>(
...
 
-    if zpi == zpr::ZPI_0 {
-        // RFC 6.5 § 5.25.2
-        pkt.put(
-            net_defs::inet_checksum(&pkt.body()[std::mem::size_of::<zdp::ZdpZpiHeader>()..])
-                .as_slice(),
-        );
-    } else {
-        todo!("encryption");
+    let enc_buf = asm.buffer_stack.try_get_buffer().unwrap();
```
cpacejo: Just allocate space on the stack here (like you do in the `km` module).  Grabbing and return a buffer is expensive (= mutex lock + unlock for each).  Allocating space on the stack is near zero-cost.

adapter/ph/src/fastpath.rs
```diff
@@ -181,24 +186,41 @@ pub fn maybe_capture_batch<'a, 'pktbuf: 'a>(
...
+    let encr_len = pkt.body().len() - 1; // Everything except the ZPI byte
+
+    match codec.encrypt_transport_stateless(&pkt.body()[1..encr_len + 1], enc_buf) {
+        Ok(len) => {
+            pkt.shrink_by(encr_len); // remove cleartext body, leavign ZPI
+            pkt.put(&enc_buf[0..len]); // copy ciphertext body over
+        }
+        Err(e) => {
+            error!("encrypt failed: {}", e);
```
cpacejo: What are the reasons encryption can fail?

Let's not log an error here (logging can block, and if encryption is failing for some state-based reason, we'll flood the log with messages).  Propagate it up a level to `substrate_egress_common()`, and there use `drop_and_count()` to drop `pkt` and count the reason why.

adapter/ph/src/fastpath.rs
```diff
@@ -181,24 +186,41 @@ pub fn maybe_capture_batch<'a, 'pktbuf: 'a>(
 }
 
 /// Encrypt a ZDP packet according to its ZPI header (which is not encrypted).
-pub fn encrypt<'pktbuf>(
-    _asm: &Assembly<'pktbuf>,
-    _link_id: zpr::LinkId,
+pub fn encrypt_zero<'pktbuf>(pkt: &mut Packet<'pktbuf>) {
```
cpacejo: Let's call it `null` instead of `zero` to match the "NULL SA" terminology.

adapter/ph/src/fastpath.rs
```diff
@@ -221,44 +243,118 @@ impl From<DecryptError> for CounterType {
...
-        // RFC 6.5 § 5.25.2
-        if !net_defs::validate_inet_checksum(
-            &pkt.body()[std::mem::size_of::<zdp::ZdpZpiHeader>()..],
-        ) {
-            return Err(DecryptError::MicvFailure);
-        }
+/// Decrypt a ZDP packet according to its ZPI header (which is not removed).
+pub fn decrypt_full<'pktbuf>(
+    asm: &Assembly<'pktbuf>,
+    codec: Arc<dyn Codec>,
```
cpacejo: Same comments here re: `&dyn Codec`, 1, and buffers.

adapter/ph/src/fastpath.rs
```diff
@@ -221,44 +243,118 @@ impl From<DecryptError> for CounterType {
...
+    pkt: &mut Packet<'pktbuf>,
+) -> Result<(), DecryptError> {
+    if pkt.body().len() < 1 {
+        return Err(DecryptError::BadStructure);
+    }
+    let encr_len = pkt.body().len() - 1;
+    if encr_len == 0 {
+        // empty?
+        return Err(DecryptError::BadStructure);
+    }
```
cpacejo: Shouldn't be an error in and of itself I don't think?

adapter/ph/src/fastpath.rs
```diff
@@ -221,44 +243,118 @@ impl From<DecryptError> for CounterType {
...
-    } else {
-        todo!("decryption");
+    match codec.decrypt_transport_stateless(&pkt.body()[1..encr_len + 1], decr_buf) {
+        Ok(len) => {
+            // Copy the decrypted data back into the message -- do not overwrite ZPI.
+            pkt.shrink_by(encr_len); // remove ciphertext body, leave ZPI
+            pkt.put(&decr_buf[0..len]); // copy over cleartext body
+        }
+        Err(e) => {
+            error!("decryption failed: {}", e);
```
cpacejo, line 292: Same comment re: logging in the fastpath.  Notably a MitM can trigger this codepath at will.

adapter/ph/src/fastpath.rs
```diff
@@ -310,11 +404,13 @@ pub async fn substrate_egress_blocking<'pktbuf>(
...
         Ok(()) => (),
-        Err(pkt) => drop_and_count(asm, pkt.into_inner(), CounterType::OutPacksErr),
+        Err(pkt) => {
+            drop_and_count(asm, pkt.into_inner(), CounterType::OutPacksErr);
+        }
     }
 }
 
-/// Process packets ingressing from the specified SA.
+/// Process packets ingressing from the specified SA (SocketAddr).
```
cpacejo: Substrate Address, not SocketAddr.

(Actually the parameter below should be a `&zpr::SubstrateAddr`.  My mistake there.)

adapter/ph/src/fastpath.rs
```diff
@@ -310,11 +404,13 @@ pub async fn substrate_egress_blocking<'pktbuf>(
         .await
     {
         Ok(()) => (),
-        Err(pkt) => drop_and_count(asm, pkt.into_inner(), CounterType::OutPacksErr),
+        Err(pkt) => {
+            drop_and_count(asm, pkt.into_inner(), CounterType::OutPacksErr);
+        }
```
cpacejo, line 423: Stray change from testing?

adapter/ph/src/fastpath.rs
```diff
@@ -327,16 +423,88 @@ pub fn substrate_ingress<'pktbuf>(
...
+                "ingress: link {}: link not in SA state table",
+                ingress_link_id
+            );
+            false
+        }
+    };
+
+    if !secure {
+        // Not under a security assocation  which means only ZPI 0 is allowed
+        if zpi_hdr.zpi != 0 {
```
cpacejo: `zpr::ZPI_0` instead of 0

adapter/ph/src/fastpath.rs
```diff
@@ -327,16 +423,88 @@ pub fn substrate_ingress<'pktbuf>(
...
+                            drop_and_count(asm, pkt, err);
+                            return;
+                        }
+                    }
+                } else {
+                    // We have an SA and ZPI does not match.
+                    info!(
+                        "ingress: link {}: unexpected ZPI value {} (expected {:?})",
+                        ingress_link_id, zpi_hdr.zpi, sa_state.transport_sa.recv_zpis
+                    );
```
cpacejo: I think in general we should refrain from logging in the fastpath as it can block.  (Also is an extra point of synchronization, and places pressure on the Icache, etc.)

(more occurrences below)

adapter/ph/src/fastpath.rs
```diff
@@ -327,16 +423,88 @@ pub fn substrate_ingress<'pktbuf>(
...
+                        Err(err) => {
+                            drop_and_count(asm, pkt, err);
+                            return;
+                        }
+                    }
+                } else if zpi_hdr.zpi == sa_state.transport_sa.recv_zpis.encr {
+                    // TODO: Put padlen in state somewhere too
+                    match decrypt_full(
+                        asm,
+                        sa_state.transport_sa.codec.clone(),
```
cpacejo: If we pass as a reference (`&*codec`) we can avoid the clone (which entails an atomic operation).

adapter/ph/src/fastpath.rs
```diff
@@ -327,16 +423,88 @@ pub fn substrate_ingress<'pktbuf>(
...
+        drop_and_count(asm, pkt, CounterType::BadStructure);
+        return;
+    };
+
+    // If a ZPI is setup on this link, then we expect the message to use one of the valid
+    // ZPI values.  Need to check spec, but I think best to drop any ZPI 0 packets IF we
+    // have an SA established.
+    let secure = match asm.sa_states.get(&ingress_link_id) {
+        Some(sa_state) => {
+            if sa_state.sa_established.load(Ordering::Relaxed) {
```
cpacejo: Normally I would be a stickler about atomic load/store ordering correctness :) but as discussed earlier this should all be moved into an `RcuBox` inside of `peer_state`.

adapter/ph/src/fastpath.rs
```diff
@@ -327,16 +423,88 @@ pub fn substrate_ingress<'pktbuf>(
...
+    let Some(zpi_hdr) = zdp::ZdpZpiHeader::read_from_prefix(&pkt.body()) else {
+        drop_and_count(asm, pkt, CounterType::BadStructure);
+        return;
+    };
+
+    // If a ZPI is setup on this link, then we expect the message to use one of the valid
+    // ZPI values.  Need to check spec, but I think best to drop any ZPI 0 packets IF we
+    // have an SA established.
+    let secure = match asm.sa_states.get(&ingress_link_id) {
+        Some(sa_state) => {
```
cpacejo: If written as `Some(sa_state) if sa_state.sa_established =>`, then both the "unestablished" and "not present" cases can be caught by a single `_ =>` branch, and the NULL decryption handled there.  (Rather than using the `secure` boolean.)

adapter/ph/src/fastpath.rs
```diff
@@ -349,13 +517,29 @@ pub fn substrate_ingress<'pktbuf>(
         return drop_and_count(asm, pkt, CounterType::BadStructure);
     };
 
+    // In ZPI zero only KM messages are allowed (well, and APR ARP which we don't support yet)
+    // Can be overridden (FOR TESTING ONLY) in the flags.
+    if !secure && base_hdr.packet_type != zdp::ZdpPacketType::KeyManagement {
```
cpacejo, line 525: & if you get rid of `secure` above, here instead of `secure` we can then just check that `zpi == 0`, since we've established by this point that `zpi == 0` if-and-only-if `!secure`.

adapter/ph/src/fastpath.rs
```diff
@@ -221,44 +243,118 @@ impl From<DecryptError> for CounterType {
...
+    let mut transit = true;
+    let mut transport_sa: Option<km::KMTransportSA> = None;
+
+    let real_zpi = match asm.sa_states.get(&link_id) {
+        Some(sa_state) => {
+            if sa_state.sa_established.load(Ordering::Relaxed) {
+                // We have SA.
+                // If agent transit packet, use HMAC ZPI.
+                // Else use our ENCR ZPI.
+                transport_sa = Some(sa_state.transport_sa.clone());
```
cpacejo: Let's avoid the atomic ref count uptick by passing a reference (`&*sa_state`).

adapter/ph/src/fastpath.rs
```diff
@@ -221,44 +243,118 @@ impl From<DecryptError> for CounterType {
...
+                if let Some(zdp_hdr) =
+                    ZdpBaseHeader::ref_from_prefix(&pkt.body()[ZDP_BASE_HEADER_OFFSET..])
+                {
+                    if zdp_hdr.packet_type == zdp::ZdpPacketType::TransitPacket {
+                        sa_state.transport_sa.send_zpis.hmac
```
cpacejo: Near-term future enhancement: hmac vs. encr here should be also controlled by flags in the assembly.

adapter/ph/src/fastpath.rs
```diff
@@ -221,44 +243,118 @@ impl From<DecryptError> for CounterType {
...
+                    } else {
+                        transit = false;
+                        sa_state.transport_sa.send_zpis.encr
+                    }
+                } else {
+                    error!("egress: link {}: failed to parse ZDP header on egress packet, assigning ZPI 0", link_id);
+                    0
```
cpacejo: Hmm… I would just `drop_and_count()` here.

adapter/ph/src/fastpath.rs
```diff
@@ -221,44 +243,118 @@ impl From<DecryptError> for CounterType {
...
 ) -> Option<zpr::SubstrateAddr> {
     // TODO: should we add ZDP header here also??
 
-    encap_zpi(asm, link_id, zpi, pkt);
+    let mut transit = true;
+    let mut transport_sa: Option<km::KMTransportSA> = None;
+
+    let real_zpi = match asm.sa_states.get(&link_id) {
+        Some(sa_state) => {
+            if sa_state.sa_established.load(Ordering::Relaxed) {
```
cpacejo: `Some(sa_state) if sa_state.sa_established =>` can be used to combine the two "else" cases into one.

adapter/ph/src/fastpath.rs
```diff
@@ -221,44 +243,118 @@ impl From<DecryptError> for CounterType {
...
 
     maybe_capture(asm, Direction::Outbound, pkt);
 
-    encrypt(asm, link_id, pkt);
+    // To "encrypt" we need to know which ZPI is in use, then we may need the HMAC or
+    // we may need the codec.
+    if real_zpi == 0 {
+        encrypt_zero(pkt);
+    } else {
+        let tsa = transport_sa.unwrap();
```
cpacejo: Maybe `match transport_sa { None => encrypt_zero(pkt); Some(tsa) => ... }`

adapter/ph/src/fastpath.rs
```diff
@@ -221,44 +243,118 @@ impl From<DecryptError> for CounterType {
...
+        Some(sa_state) => {
+            if sa_state.sa_established.load(Ordering::Relaxed) {
+                // We have SA.
+                // If agent transit packet, use HMAC ZPI.
+                // Else use our ENCR ZPI.
+                transport_sa = Some(sa_state.transport_sa.clone());
+                if let Some(zdp_hdr) =
+                    ZdpBaseHeader::ref_from_prefix(&pkt.body()[ZDP_BASE_HEADER_OFFSET..])
+                {
+                    if zdp_hdr.packet_type == zdp::ZdpPacketType::TransitPacket {
```
cpacejo: Let's hoist "is it a transit packet?" out of this `match` and perform it first.

adapter/ph/src/fastpath.rs
```diff
@@ -221,44 +243,118 @@ impl From<DecryptError> for CounterType {
...
+    if encr_len == 0 {
+        // empty?
+        return Err(DecryptError::BadStructure);
+    }
+    if encr_len < padlen {
+        return Err(DecryptError::BadStructure);
+    }
 
-        pkt.shrink_by(2); // remove checksum
+    let decr_buf = asm.buffer_stack.try_get_buffer().unwrap();
```
cpacejo, line 283: (Also -- I think this is missing getting returned to the buffer pool.  But -- switching to a stack allocation avoids the issue entirely.)

cpacejo (changes requested):
Overall looks good!  Lots of comments mostly nits, some buffer management issues in `fastpath` (hence "request changes").

adapter/ph/src/fastpath.rs
cpacejo: I understand the desire to log in the fastpath though – and it is possible to do without impacting performance by using a lossy ring buffer – I've opened #386 to look into this

adapter/ph/src/main.rs
mkolehmainen: Adding `move` to that opens a whole world of hurt...

adapter/ph/src/km_multiplexor.rs
mkolehmainen: good call. I forgot that spawn returns something...

adapter/ph/src/main.rs
cpacejo: Oh because then it tries to copy/move `asm`.  This issue is addressed by the structure of the `launch()` functions I write for workers (all of which use `async move`).  If you do the same for this worker it should be trivial to pass in `km_mpx_ctok.clone()` as an argument which then gets moved into the worker.

adapter/ph/src/peer_table.rs
```diff
@@ -67,23 +71,34 @@ pub struct PeerTable<'pktbuf> {
     peer_slab: Mutex<RcuCslab<PeerState<'pktbuf>>>,
     peer_slab_reader: RcuBox<RcuCslabReader<PeerState<'pktbuf>>>,
     sa_to_link: DashMap<SubstrateAddr, LinkId>,
+
+    // `link_so_sec_assoc` is state info for the security association per link. This is managed by the
+    // KM Multiplexor and used during encrypt/decrypt.
+    link_to_sec_assoc: DashMap<LinkId, SAState>,
```
cpacejo: Sorry what I meant more specifically is – `SAState` can/should move into `PeerState` here -- `peer_slab` is where the actual contents of the peer table live.  (`peer_slab_reader` is just a read-only handle, and `sa_to_link` is just a reverse mapping.)

I recognize this is probably a headache to do so feel free to address in a separate commit / or I can take a stab at it as I know the Rcu and Cslab stuff more intimately.

adapter/ph/src/fastpath.rs
mkolehmainen: Yes, I got well into this before I realized that you were not using "sa" to mean "Security Association".  I am removing all instances of "sa" meaning "security assocaition" now.

adapter/ph/src/fastpath.rs
mkolehmainen: This I don't understand!

adapter/ph/src/km_noise.rs
```diff
@@ -147,42 +147,56 @@ impl Codec for NoiseCodec {
...
         {
             Ok(len) => Ok(len + NOISE_NONCE_LEN),
-            Err(e) => {
-                error!("noise: error encrypting message: {:?}", e);
-                Err(KMError::EncryptionError)
-            }
+            Err(e) => match e {
+                snow::error::Error::Input => Err(EncryptionError::MessageTooLarge),
+                _ => Err(EncryptionError::InternalError(format!(
+                    "noise failed to encrypt message: {}",
```
cpacejo: I think in this case it's OK to panic – I think the only cases that encryption can fail other than `Error::Input` is due to a logic/setup error on our part, in which case it's not possible to continue in any way.

adapter/ph/src/km_noise.rs
```diff
@@ -147,42 +147,56 @@ impl Codec for NoiseCodec {
...
             .snow_state
             .write_message(nonce, payload, &mut message[NOISE_NONCE_LEN..])
         {
             Ok(len) => Ok(len + NOISE_NONCE_LEN),
-            Err(e) => {
-                error!("noise: error encrypting message: {:?}", e);
-                Err(KMError::EncryptionError)
-            }
+            Err(e) => match e {
+                snow::error::Error::Input => Err(EncryptionError::MessageTooLarge),
```
cpacejo, line 167: It looks like this can [only be generated](https://docs.rs/snow/latest/src/snow/stateless_transportstate.rs.html#64-65) if either the output buffer is too small (which we should be able to statically ensure it's not) or the overall message is > 64 KiB (which we should also be able to ensure doesn't happen).  I wish `snow` exposed these constants but they are found [here](https://docs.rs/snow/latest/src/snow/constants.rs.html#8).

What I'm aiming for is encryption should never fail "dynamically", with the exception that we've encrypted too many bytes and therefore need to / should have already performed a rekey.  Simply because there's nothing really useful we can do to respond to a dynamic failure other than drop the packet, which doesn't help the sending application figure out why its packets aren't going through and adjust its behavior.

Not worth solving now I don't think but maybe worth a note.

adapter/ph/src/peer_table.rs
mkolehmainen: Ah! Yes, let me make another task for that.  At same time can iron out the use of clone and reference which is also causing me issues.

adapter/ph/src/km_noise.rs
```diff
@@ -147,42 +147,56 @@ impl Codec for NoiseCodec {
...
+    ) -> Result<usize, DecryptionError> {
         // nonce is first 8 bytes of the message.
         let plen = payload.len();
         if plen < NOISE_NONCE_LEN {
             error!("noise: message too short");
-            return Err(KMError::EncryptionError);
+            return Err(DecryptionError::MessageTooShort);
         }
-        let nonce: u64 = u64::from_be_bytes(payload[0..NOISE_NONCE_LEN].try_into().unwrap());
+        let nonce: u64 = u64::from_be_bytes(payload[0..NOISE_NONCE_LEN].try_into().unwrap()); // pretty sure this cannot fail
```
cpacejo, line 184: Alternative: the `array_ref![]` macro: https://docs.rs/arrayref/latest/arrayref/macro.array_ref.html  which allows slicing an array to another array.  (This is fine/correct as-is though.)

adapter/ph/src/fastpath.rs
```diff
@@ -304,58 +304,53 @@ fn substrate_egress_common<'pktbuf>(
...
+        Some(zdp_hdr) => zdp_hdr,
+        None => {
+            return Err(km::EncryptionError::ParseError);
+        }
+    };
+    let transit = zdp_hdr.packet_type == zdp::ZdpPacketType::TransitPacket;
 
-    let real_zpi = match asm
+    // Get the security association for this link and extrant the correct ZPI.
+    // TODO: Is there some way to avoid a clone here?
```
cpacejo, line 316: If the SA stuff lives inside a `peer_table` entry then you can just `inspect()` the entry and will have access to whatever's in there for the duration of the call to `inspect()`.

adapter/ph/src/fastpath.rs
cpacejo: The `if xxx` is termed a "guard".  What it means is – the match scrutinee (the expression after `match`) is first matched against the match arm's pattern (`Some(sa_state)` in this case).  If it matches, the variables are bound (`sa_state` here), and then the arm's guard expression is checked.  If it's true, the match arm's block is then executed.  If the guard expression is false, the arm is considered not to match, and the next pattern is checked.  So instead of writing:

```
match foo {
  Some(x) => {
    if x > 5 { do stuff when x > 5 }
    else { do stuff when x <= 5 }
  }
  None => { do stuff when there's no x }
}
```
you can write:
```
match foo {
  Some(x) if x > 5 => { do stuff when x > 5 }
  Some(x) => { do stuff when x <= 5 }
  None => { do stuff when there's no x }
}
```
Or, if you want to do the same stuff when `x <= 5` or there is no `x`:
```
match foo {
  Some(x) when x > 5 => { do stuff when x > 5 }
  Some(_) | None => { do stuff when x <= 5 or there's no x }
}
```
or you can save characters at the expense of fewer static checks like so:
```
match foo {
  Some(x) if x > 5 => { do stuff when x > 5 }
  _ => { do stuff when x <= 5 or there's no x }
}
```

Not a code issue here, I think it's fine to be checked in as-is.  Just an easier way to express what's written in my opinion.

adapter/ph/src/main.rs
mkolehmainen: Yes, this is moved into a worker now.

adapter/ph/src/km.rs
mkolehmainen: I don't know... I just copied how I was exchanging the ZPIs.

adapter/ph/src/fastpath.rs
mkolehmainen: thanks for explanation!

adapter/ph/src/km_noise.rs
mkolehmainen: ok!

cpacejo (approved):
Thanks for making all these changes!  I think everything major is taken care of… like I mentioned feel free to leave the `peer_state` stuff for me to sort out.

adapter/ph/src/km.rs
mkolehmainen: phew

adapter/ph/src/km_multiplexor.rs
mkolehmainen: Yes.  I think so. We need to centralize the link-setup procedure.  When we create a link in the peer-table there are things that should happen, including starting up KM.  There are many things that can bring a link down (eg, KM completely fails after too many retries, or user kills it, or visa service/admin kills it, or it is lost based on some other retry mechanism like keep alives go missing...).  I'll add a task but it is pretty vague at the moment.

PR #379, opened 2024-08-29
Add Dock Forwarding Table, etc.

### LINKED ISSUE #376: Dock Forwarding Table

### LINKED ISSUE #377: spawn mgmt worker per peer
Everything is single-threaded through `mgmt_worker` right now.  Each peer should have its own `mgmt_worker` servicing only that peer.

cpacejo:
Multiple changes in here:

* incorporates @svjaffer's changes from #372 to support adapter<->node<->adapter forwarding, and add extra debugging
* Dock Forwarding Table now exists
* SyncReqState moved into PeerState
* mgmt_processor moved into PeerState
* `Cslab` grew a notion of "vacant entry" (idea borrowed from the `slab` crate), so we can create self-referencing entries (necessary for mgmt_processor)
* docks now can issue Bind Requests
* temp hack in a node for issuing a bind request to "the" egress adapter when a bind request is received (pending visas or any forwarding support)

Probably others.

The changes in this PR allow the Adapter<->Node<->Adapter packet traffic to flow with dynamically issued Bind requests.

IPv6 testing/support is temporary disabled, because there is some race condition (other other concurrency issue) which gets tripped up due to the kernel sending many IPv6 packets at once while assigning IPv6 link-local addresses.  I will sort that out next week.  Possible that we need proper mgmt req sequence numbers and/or windowing to fix this properly.

Also lots of debug messages added; they're noisy and I will remove them next week after fixing the above.

Apologies this is not my usual small self-contained PR.  I was unable to break it up further due to time pressures.  If this PR is still open next week I'll restructure it into smaller pieces.


adapter/ph/src/dock_tables.rs
```diff
@@ -0,0 +1,63 @@
...
+// visas may be more narrowly scoped than what can be compressed out)
+pub struct DftPep {
+    pub next_hop: DftNextHop,
+}
+
+pub struct DockForwardingTable {
+    table: Mutex<RcuCslab<DftPep>>,
+    reader: RcuBox<RcuCslabReader<DftPep>>,
+}
+
```
mkolehmainen, line 32: Nice straightforward example of how to use the RcuBox and RcuCSlab.  I'll soon be copying...

mkolehmainen (approved):
Looks good to me.

svjaffer (approved):

PR #380, opened 2024-08-30
adds a helper function for testing with assembly

mkolehmainen:
This is a helper function for test mode that allows you to create an assembly with various bits stubbed out.

This is not used in this PR but I am using it in my next one involving hooking up KM.  The "allow dead code" stuff is just temporary.

### DIFF
--- a/adapter/ph/src/assembly.rs
+++ b/adapter/ph/src/assembly.rs
@@ -10,7 +10,7 @@ use crate::mgmt;
 use crate::packet::*;
 use crate::peer_table;
 use crate::queues::*;
-use crate::tun_ctl::TunCtl;
+use crate::tun_ctl::CarrierSetter;
 use crate::zdp::*;
 use crate::zpr;
 use bytes::Buf;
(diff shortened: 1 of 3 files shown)


svjaffer (approved):

cpacejo:
I'd much rather the interface be called `TunCtl` and the implementation be called `TunCtlImpl`.  This may/will eventually do more than simply allow carrier control.

adapter/ph/src/assembly.rs
```diff
@@ -57,7 +57,7 @@ pub struct Assembly<'pktbuf> {
 
     pub counters: EnumMap<CounterType, Counter>,
 
-    pub tun_ctl: TunCtl<'pktbuf>,
+    pub tun_ctl: &'pktbuf dyn CarrierSetter,
```
cpacejo, line 60: Let's make this an owning reference (`Box<dyn CarrierSetter>`).  This retains the original ownership model, and no need to worry about lifetimes then.

cpacejo:
Thanks for adding this test harness!!  Hopefully we'll be making much use of this soon…

PR #381, opened 2024-09-03
restore name of TunCtl and use owning reference

cpacejo:
Implementation of my comments on PR #380.

Just renames CarrierSetter back to TunCtl (since the intent is for the interface to support more than just setting the carrier), replaces the shared reference in Assembly with an owning reference to retain the original ownership model, and reflects the implementation/interface split in the comments.


mkolehmainen (approved):
Looks good to me!

PR #383, opened 2024-09-04
add some fastpath diagrams

cpacejo:
Made some diagrams with PlantUML.  I don't love them yet but they're decent.  Considering combining them into one big diagram.


mkolehmainen (approved):
Great! I think maybe I would take your message in slack and put it in a README here?

Also, consider checking in the generated images?  I think these will be useful to include in other documentation.  On their own they are a little dry ;)

PR #393, opened 2024-09-06
RcuBox get

cpacejo:
Finally figured out how to implement a simple `get()` for `RcuBox`!

Did that, then implemented `get()` using that for the `RcuCslab`-based tables, and then used `get()` in some places.

(Not all -- some I'm about to rewrite in another branch so don't want to conflict, others actually work better using `inspect()` since it's effectively `.get().map()`.)

Since there's a lot of copy-paste in the table implementations now, I opened #392 to address that.


mkolehmainen (approved):
Looks good to me!

PR #396, opened 2024-09-09
slowdown KM retry

### LINKED ISSUE #394: Slow down KM retry
The KM state machine goes immediately to reset after an error which can cause a TON of messages to get sent.  Slow that down, make configurable.  Or alternatively, bump that transition up to next level and let the PH trigger it (where it can count errors, and back off, etc).

### LINKED ISSUE #408: add timeout detect to KM noise
Outstanding TODO in km_noise.rs is to retry on timeout during the simple handshake.  Important, especially for case when node is not up yet.

mkolehmainen:
Moved the retry logic out of KM state machine as it was just running way to fast and I think we will want to have more control over that.  Now the logic is up to use of the KM.  So for us that is km_multiplexor.

Also put some noise key material into "main" so that it makes sense if anybody ever tries to run this with KM enabled.

I believe we are reworking main anyway, so this is probably very temporary.

### DIFF
--- a/adapter/ph/src/km.rs
+++ b/adapter/ph/src/km.rs
@@ -576,24 +576,26 @@ impl KeyManager {
             }
         }
 
-        // Tick machine, even during error.
-        let resp: Option<Bytes>;
-        {
-            let mut state = self.shared.state.lock().unwrap();
-            resp = match state.statemachine.tick() {
-                Ok(h) => h,
-                Err(e) => {
-                    warn!("error during tick processing: {}", e);
-                    None
-                }
-            };
-        }
-        if let Some(r) = resp {
-            match km_buffers_out.send(KmLinkMsg::new(link_id, r)).await {
-                Ok(_) => {}
-                Err(_) => {
-                    error!("failed to enqueue oubound KM message");
-                    return Err(KmError::EnqueueFailed);
+        // Unless we did a reset, tick machine, even during error.
+        if !did_reset {
+            let resp: Option<Bytes>;
+            {
+                let mut state = self.shared.state.lock().unwrap();
+                resp = match state.statemachine.tick() {
+                    Ok(h) => h,
+                    Err(e) => {
+                        warn!("error during tick processing: {}", e);
+                        None
+                    }
+                };
+            }
+            if let Some(r) = resp {
+                match km_buffers_out.send(KmLinkMsg::new(link_id, r)).await {
+                    Ok(_) => {}
+                    Err(_) => {
+                        error!("failed to enqueue oubound KM message");
+                        return Err(KmError::EnqueueFailed);
+                    }
                 }
             }
         }
--- a/adapter/ph/src/km.rs
+++ b/adapter/ph/src/km.rs
@@ -23,7 +23,7 @@ use bytes::{BufMut, Bytes};
 
 use zerocopy::FromBytes;
 
-use tracing::{error, warn, info};
+use tracing::{error, info, warn};
 
 use crate::config;
 use crate::packet::Packet;
@@ -425,7 +425,6 @@ impl KeyManager {
                 prev_state = state.statemachine.get_state();
             }
 
-
             tokio::select! {
                 _ = ctok.cancelled() => {
                     match self.send_signal(&km_signals_out, link_id, KmSignal::Termination).await {
@@ -535,7 +534,6 @@ impl KeyManager {
         km_signals_out: &mpsc::Sender<KmLinkMsg<KmSignal>>,
         cur_state: &KmSMState,
     ) -> KmResult<()> {
-
         let mut resp: Option<Bytes> = None;
         let mut did_reset = false;
 
@@ -720,7 +718,6 @@ impl KeyManager {
         }
         Ok(())
     }
-
 }
 
 impl KmTransportSA {
--- a/adapter/ph/src/km.rs
+++ b/adapter/ph/src/km.rs
@@ -23,7 +23,7 @@ use bytes::{BufMut, Bytes};
 
 use zerocopy::FromBytes;
 
-use tracing::{error, info};
+use tracing::{error, warn, info};
 
 use crate::config;
 use crate::packet::Packet;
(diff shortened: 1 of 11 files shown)


adapter/ph/src/km_multiplexor.rs
```diff
@@ -62,18 +76,46 @@ async fn signal_worker<'pktbuf>(
...
             _ = sp_ctok.cancelled() => {
                 info!("KM Multiplexor shutting down");
                 break;
             }
 
+            _ = interval.tick() => {
+                let now = time::Instant::now();
+                for (link_id, stat) in status.iter_mut() {
+                    if stat.error_count > 0 && stat.last_error_t >= stat.restart_t {
+                        if now - stat.last_error_t > time::Duration::from_secs(stat.error_count as u64) {
```
svjaffer: This might be a small thing, but I don't think these if statements need to be nested like this.  Why not combine them?

adapter/ph/src/km.rs
```diff
@@ -427,51 +447,73 @@ impl KeyManager {
```
svjaffer, line 528: This is a long loop inside of an even longer function.  Any chance that you could break this up into smaller functional pieces for readability?

adapter/ph/src/km.rs
```diff
@@ -427,51 +447,73 @@ impl KeyManager {
...
                         }
                     }
-                    let resp: Option<Bytes>;
-                    {
+                    if restart_request {
                         let mut state = self.shared.state.lock().unwrap();
                         resp = match state.statemachine.reset() {
                             Ok(h) => h,
                             Err(e) => {
                                 return Err(KmError::MachineError(e.to_string()));
```
svjaffer: If you error our here without a panic or anything, the KM will just stop running without taking the rest of the system down, leaving it effectively disabled unless there's something that restarts this thread?

adapter/ph/src/km.rs
```diff
@@ -427,51 +447,73 @@ impl KeyManager {
...
+                        {
+                            let mut state = self.shared.state.lock().unwrap();
+                            state.error_signaled = true;
                         }
                     }
-                    let resp: Option<Bytes>;
-                    {
+                    if restart_request {
                         let mut state = self.shared.state.lock().unwrap();
                         resp = match state.statemachine.reset() {
```
svjaffer: Why not just do this in the other thread instead of setting restart_required if it's under a lock anyway?

adapter/ph/src/km_multiplexor.rs
mkolehmainen: yes! good point.

adapter/ph/src/km.rs
mkolehmainen: Love this idea. I'll give it a try.

adapter/ph/src/km.rs
mkolehmainen: This ends up getting caught in km_multiplexor (link below).  But even there we don't deal with it.  The whole link-up/link-down/link-error thing needs thought.

https://github.com/org-zpr/zpr-core/blob/cf32e9ce075c7169c1c6f0960c94207428c48122/adapter/ph/src/km_multiplexor.rs#L289

adapter/ph/src/km.rs
mkolehmainen: Good point, but I really wanted one place were I mess with the underlying state machine, handle messages and detect transitions.

svjaffer: I like that as a goal, but in this particular case I worry it's causing more complexity and potential future bugs than it's saving.

Maybe there's another way to go about it.  I'll think about it.

adapter/ph/src/km.rs
svjaffer: So, more generally, the keying erroring out is supposed to cause the link to close.  FWIW:
image

So, the error needs to propagate into the link state stuff I'm working on (which I'll tie in, since it's not there for you to hook into right now), but probably the thread should stay alive?

adapter/ph/src/km_multiplexor.rs
```diff
@@ -62,18 +76,44 @@ async fn signal_worker<'pktbuf>(
...
             _ = sp_ctok.cancelled() => {
                 info!("KM Multiplexor shutting down");
                 break;
             }
 
+            _ = interval.tick() => {
+                let now = time::Instant::now();
+                for (link_id, stat) in status.iter_mut() {
+                    if stat.error_count > 0 && stat.last_error_t >= stat.restart_t && now - stat.last_error_t > time::Duration::from_secs(stat.error_count as u64) {
+                        info!("km_multiplexor: restarting KM on link {}", link_id);
```
svjaffer: Do we want this message to print if the following line evaluates to false (however unlikely that may be)?  If you move this inside the if clause, you could also fold the if statements together and eliminate another branch :)

adapter/ph/src/km.rs
mkolehmainen: done!

adapter/ph/src/km.rs
mkolehmainen: Doesn't need to.  I think the link should go "down" meaning be destroyed from memory and a new one started up.  Each link has its own KeyManager running on it.  That is stored in the peer_table.

So, ideally, when we bring up a link we dispatch to key manager and nothing happens on link until security association is established.  If key process fails or the security assoc fails later for any reason, that should trigger a link down.

When link is brought down administratively (node could close a link for a variety of reasons, or user of an adapter could close a link explicitly) then that code should also notify the key manager that the link is going down.

I think maybe somewhere in the peer table is the right place to put that logic?  Right now if you look in main all these steps are done manually.

adapter/ph/src/km_multiplexor.rs
mkolehmainen: oh I see... I'll change the wording a bit...

adapter/ph/src/km.rs
svjaffer: I see, got it.

svjaffer (approved):

PR #397, opened 2024-09-09
use LocalSet for all mgmt workers

cpacejo:
It's getting annoying/prohibitive not to be able to hold `Send` objects across async calls in mgmt threads.

This is be fixed by instructing Tokio to run all mgmt threads on the same OS thread.

This is effected by launching the main async task with an associated `LocalSet`, which allows the use of `spawn_local()`.  Tasks spawned with `spawn_local()` are then just run on the main thread and don't need to be `Send`.

This should not be an issue performance-wise.

Also, this commit breaks up the big main async block a bit.  Best to select "hide whitespace" in the diff view settings to see what's happening.


mkolehmainen (approved):
Looks good to me!

Issue #401, opened 2024-09-09
[BLOCKED] generate async request sequence numbers


cpacejo:
Blocked on:
* org-zpr/zpr-core#398

cpacejo:
Eliminating request-response system.  See also #839.

Issue #400, opened 2024-09-09
[BLOCKED] validate request sequence numbers


cpacejo:
Blocked on:
* org-zpr/zpr-core#401

cpacejo:
Eliminating request-response system.

Issue #399, opened 2024-09-09
make sync req handlers async funs


cpacejo:

cpacejo:
eehhhhh this would require serializing the packet outside of these functions which is kinda hairy, let's punt this for now.

Issue #398, opened 2024-09-09
sync req: separate window state from listener state


cpacejo:

cpacejo:
We no longer have sync requests.

PR #402, opened 2024-09-09
send & check sequence numbers on sync reqs

### LINKED ISSUE #389: message sequence numbers
Let's get these working…

cpacejo:
Imbues SyncReqState with a notion of sequence numbers, which are handed out with each window permit granted.  These sequence numbers are plumbed through the sync request->reply loop and matched up on reception.

I don't love this code, and it should be replaced long-term once we figure out our broader story for sequence numbers.  But it's here for now.

Some follow-on TODOs --

* The two halves of SyncReqState, the window state, and the listener state, are mostly independent and can/should be split out.  The window state can then move up a layer and be used for all management messages and/or implemented as part of a transport layer. This is #398.
* We need to generate sequence numbers for async requests also.  This is #401.
* On receipt, sequence numbers of requests should be validated that they are nondecreasing, to prevent replays and out-of-order execution.  This is #400.
* The synchronous request handlers should be async functions, and the logic to send a reply can then be centralized (rather than stuck in each handler function).  This is #399.
* It would be nice if sequence numbers served as nonces.  This is #358.

### DIFF
--- a/adapter/ph/src/mgmt.rs
+++ b/adapter/ph/src/mgmt.rs
@@ -49,7 +49,15 @@ pub async fn send_non_flow_mgmt_response<'pktbuf>(
     sequence_number: zpr::SeqNum,
     packet: Packet<'pktbuf>,
 ) {
-    send_mgmt_helper(asm, link_id, packet_type, None, Some(sequence_number), packet).await
+    send_mgmt_helper(
+        asm,
+        link_id,
+        packet_type,
+        None,
+        Some(sequence_number),
+        packet,
+    )
+    .await
 }
 
 pub async fn send_per_flow_mgmt_response<'pktbuf>(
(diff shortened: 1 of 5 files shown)


adapter/ph/src/mgmt.rs
```diff
@@ -378,6 +416,25 @@ pub async fn send_bind_agent_address_request<'a, 'pktbuf>(
     }
 }
 
+/// Send a key management message out the given link.
+pub async fn send_key_management<'pktbuf>(
```
cpacejo, line 478: This just moved to be with the rest of the "send" functionality.  No changes.

adapter/ph/src/mgmt.rs
```diff
@@ -689,37 +749,19 @@ pub async fn handle_bind_agent_address_request<'pktbuf>(
...
+        seq_num,
         rsp_pkt,
     )
     .await;
 
     Ok(())
 }
 
-/// Send a key management message out the given link.
-pub async fn send_key_management<'pktbuf>(
```
cpacejo, line 750: Moved, not removed.  See above.

PR #404, opened 2024-09-10
Fix mgmt deadlock

### LINKED ISSUE #403: fix mgmt deadlock

cpacejo:
We had a deadlock in mgmt processing, due to the fact that the mgmt processor both was routing responses, and also processing requests.

So, rather than enqueue all mgmt packets immediately into the mgmt processor, the fastpath now performs the dispatch of request vs. response packets, sending requests to the mgmt processor queue as before, and responses get routed to the active response listener.

This is a little more heavyweight than I'd like in the fastpath, but at least it's nonblocking.  In the future, we could add a queue to perform this mgmt dispatching.

Also, this patch re-enables IPv6 processing (which now no longer deadlocks!) and removes many of the debug statements I added to track down this issue.  (Notably, there are no eprintlns left in the PH.)

Finally, I also moved KM enqueueing out of the mgmt processor and into the newly created `dispatch_mgmt_packet()`.  Now `mgmt_processor` is dedicated to actually processing mgmt requests.

### DIFF
--- a/adapter/ph/src/fastpath.rs
+++ b/adapter/ph/src/fastpath.rs
@@ -537,7 +537,7 @@ pub fn substrate_ingress<'pktbuf>(
         // TODO: should we peel off the ZDP header here??
         // (instead of this silly code to restore it?)
         *pkt.alloc_zeroed_header() = base_hdr;
-        mgmt::route_mgmt_packet(asm, ingress_link_id, pkt);
+        mgmt::dispatch_mgmt_packet(asm, ingress_link_id, pkt);
         return;
     }
 
--- a/adapter/ph/src/fastpath.rs
+++ b/adapter/ph/src/fastpath.rs
@@ -23,7 +23,7 @@ use crate::{compress, km};
 use blake3;
 use bytes::{Buf, BufMut};
 use std::time::SystemTime;
-use tracing::{error, info, warn};
+use tracing::{debug, error, info, warn};
 use zerocopy::FromBytes;
 use zpr_ext::std::mem::{drop_guard, DropGuard};
 use zpr_ext::zerocopy::*;
@@ -35,7 +35,7 @@ pub fn drop_and_count<'pktbuf>(
     reason: impl Into<CounterType>,
 ) {
     let reason = reason.into();
-    eprintln!("{}: dropping packet because {}", asm.system_name, reason);
+    debug!("{}: dropping packet because {}", asm.system_name, reason);
     asm.buffer_stack.put_buffer(pkt.destroy());
     asm.counters[reason.into()].increment();
 }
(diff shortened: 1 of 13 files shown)


adapter/ph/src/mgmt.rs
```diff
@@ -215,12 +216,66 @@ fn match_received<'pktbuf>(
...
+                Ok(()) => (),
+                Err(queues::TryEnqueueError::Full(pkt)) => {
+                    fastpath::drop_and_count(asm, pkt, CounterType::QueueBackpressure);
+                }
+            }
+        }
+    }
+}
+
+fn handle_response<'pktbuf>(
```
cpacejo, line 256: Not new code; factored out from `mgmt_processor_worker()`.

adapter/ph/src/fastpath.rs
```diff
@@ -523,9 +522,7 @@ pub fn substrate_ingress<'pktbuf>(
     // In ZPI zero only KM messages are allowed (well, and APR ARP which we don't support yet)
     // Can be overridden (FOR TESTING ONLY) in the flags.
     if !secure && base_hdr.packet_type != zdp::ZdpPacketType::KeyManagement {
-        if asm.flags.allow_insecure_zpi_zero {
-            warn!("operating in insecure mode - allow_insecure_zpi_zero is ENABLED");
```
cpacejo, line 527: This message was very noisy (appearing on every transit packet) so I removed it!

mkolehmainen: That's fine and always fine to add/remove log messages within reason in my opinion.  The point I was trying to make was that this should never run with "allow_insecure_zpi_zero" mode once we have working KM.  So being noisy was supposed to help get the KM integrated properly.  Ideally we remove the flag altogether at some point so that we never get insecure behavior.

adapter/ph/src/mgmt_processor_worker.rs
```diff
@@ -54,32 +50,18 @@ async fn handle_packet<'pktbuf>(
...
-        asm.peer_table
-            .inspect(ingress_link_id, |peer_state| {
-                peer_state
-                    .sync_req_state
-                    .forward_response((packet_type, pkt.take().unwrap()))
-                    .map_err(|pkt| (HandleMgmtError::UnexpectedMgmtResponse, pkt))
-            })
-            .unwrap_or_else(|| Err((HandleMgmtError::UnexpectedMgmtResponse, pkt.take().unwrap())))
-    } else if base_hdr.packet_type.is_per_flow() {
+    debug_assert!(
```
mkolehmainen: hmm... panic in "debug" mode else continue?  Maybe log it "release" mode?

mkolehmainen (approved):
One raised eyebrow about `panic_debug` but otherwise looks good!

adapter/ph/src/mgmt_processor_worker.rs
cpacejo: You're right this should be `assert!` to match with the `panic!`s below!

PR #405, opened 2024-09-10
add calls to really do the link-2-link hmac

mkolehmainen:
Fill in code for encrypt_hmac and decrypt_hmac.

Puts 8 byte HMAC on end of transit packets.

### DIFF
--- a/adapter/ph/src/zdp.rs
+++ b/adapter/ph/src/zdp.rs
@@ -106,6 +106,7 @@ pub struct ZdpPerFlowHeader {
     pub stream_id: U32,
 }
 
+
 #[derive(FromZeroes, FromBytes, AsBytes, Unaligned)]
 #[repr(packed)]
 pub struct ZdpEchoHeader {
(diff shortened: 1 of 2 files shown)


adapter/ph/src/zdp.rs
```diff
@@ -178,5 +178,9 @@ pub struct ZdpA2aHeader {
 /// Config-specified size of A2A MAC.  Algorithm-specified MAC may be smaller (but not larger).
 pub const ZDP_A2A_MAC_SIZE: usize = 8;
 
+/// Size of the ZDP "link" HMAC which is set link-by-link for transit packets.
+/// This HMAC is tacked on to the end of the packet (following the A2A HMAC).
```
cpacejo, line 182: Technically not specifically transit packets, but any packet which is MACed but not encrypted.  (In our default config, that is transit packets, but I think we'll eventually also want to support a config where transit packets are encrypted (for the privacy conscious), and one where nothing is encrypted (for the performance conscious).)

cpacejo (commented):
Perfect!!  Just a nit on a comment.

cpacejo (approved):

PR #407, opened 2024-09-11
store first packet on a new stream until bind completes

### LINKED ISSUE #406: figure out the "first packet" problem -- especially for DNS

cpacejo:
This solves the "first packet" problem, which, beside categorically slowing down TCP (due to long ACK timeouts), also completely broke DNS over UDP (since each retry uses a fresh source port).

The solution is simple -- store the first packet with the "pending" entry in the ALT, and send it once the bind completes.

As an extra nicety, if another packet is seen in the interim, the initial packet is flagged as no longer needed, causing it to be dropped rather than sent when the bind completes.  (It's not dropped immediately to avoid the extra overhead of synchronizing with the mgmt plane.)  This is a trivial mechanism which avoids sending a now out-of-order packet which apparently is not missed anyway.

Basic test shows all packets passing now.  (Haven't tested with DNS yet but I expect it to work.)

### DIFF
--- a/adapter/ph/src/adapter_manager_worker.rs
+++ b/adapter/ph/src/adapter_manager_worker.rs
@@ -42,17 +42,23 @@ async fn do_request_tether_id<'pktbuf>(asm: &Assembly<'pktbuf>, pkt: Packet<'pkt
 
     // just extract 5t and drop packet for now, storing & resending it later is a TODO
     let five_tuple = *pkt.metadata().five_tuple();
-    fastpath::drop_and_count(asm, pkt, CounterType::DroppedAwaitingBind);
 
     // if there's already an entry, this is a duplicate request
     // (NOTE: we should be the only ones modifying this table!)
     if asm.alt.get(&five_tuple).is_some() {
+        fastpath::drop_and_count(asm, pkt, CounterType::DroppedAwaitingBind);
         return;
     }
 
     // mark ALT entry as pending to attempt to (i.e. racily) prevent
     // fastpath from issuing multiple requests
-    asm.alt.insert(five_tuple, AltEntry::Pending);
+    asm.alt.insert(
+        five_tuple,
+        AltEntry::Pending {
+            initial_packet: pkt,
+            more_packets_seen: false.into(),
+        },
+    );
 
     // compress only IP addresses for now
     let compression_mode: zpr::CompressionMode = 0;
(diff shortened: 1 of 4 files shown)


adapter/ph/src/adapter_tables.rs
```diff
@@ -20,27 +21,29 @@ pub struct AltPep {
     pub tether_id: StreamId,
 }
 
-#[derive(Clone, Copy)]
-pub enum AltEntry {
+pub enum AltEntry<'pktbuf> {
     Active(AltPep),
-    Pending,
+    Pending {
```
nonrenewable, line 26: maybe can be a struct like the one above it is? (or there's a reason not to make them the same due to details i don't know about. my suggestion comes from a purely "consistency of behavior is nice" perspective!)

adapter/ph/src/adapter_manager_worker.rs
```diff
@@ -78,15 +84,38 @@ async fn do_request_tether_id<'pktbuf>(asm: &Assembly<'pktbuf>, pkt: Packet<'pkt
...
+                more_packets_seen,
+            } = asm
+                .alt
                 .alter(&five_tuple, |entry| {
-                    assert!(matches!(entry, AltEntry::Pending));
-                    *entry = AltEntry::Active(AltPep {
-                        compression_mode,
-                        tether_id,
-                    });
+                    std::mem::replace(
```
nonrenewable, line 101: just to confirm - this is where your "sanity checking" we're in Pending state is now tucked away via a concurrency safety check? and if there's enough of a delay could it replace an existing Active with a new active or does it somehow guarantee it only replaces Pending states and i can't see it?

adapter/ph/src/adapter_tables.rs
```diff
@@ -20,27 +21,29 @@ pub struct AltPep {
...
+        initial_packet: Packet<'pktbuf>,
+        more_packets_seen: atomic::AtomicBool,
+    },
 }
 
-pub struct AgentLookupTable {
-    table: DashMap<FiveTuple, AltEntry>,
+pub struct AgentLookupTable<'pktbuf> {
```
nonrenewable, line 32: these additions all seem like just syntax niceness/explicitness to the type, is that a correct interpretation? or is there a functional change here?

nonrenewable (approved):
I think this all makes sense, i have one really cosmetic suggestion that you can take or leave and a bunch of "am i reading this right" questions.

One higher level question that I figured you can just answer off the top of your head for me that i spent 2min trying to track down and then stopped trying and figured i'd ask -- 
how does the fast path end up triggering the adapter manager worker code? or is it the other way around? it's not directly related to the PR but since i'm here figured i'd try to fill in the blanks!

adapter/ph/src/adapter_tables.rs
cpacejo: Oh, you know I didn't even notice this 😅

cpacejo: Struct vs. enum args aren't _quite_ the same thing semantically, though they are isomorphic (and almost certainly compile identically).  The `AltPep` is a cohesive unit which can be copied, debugged, etc.  The two arguments of `Pending` are separate and can't be treated as one.  It… made sense to do it this way when I wrote this?

adapter/ph/src/adapter_manager_worker.rs
cpacejo: The swap occurs atomically, since `.alter()` calls its closure under a lock.

However, you're right it _will_ stomp an existing `Active` entry – but if it does so, we immediately panic.  I don't love this though as it allows the system to briefly operate in an incorrect state, so I think I'll change it around.

adapter/ph/src/adapter_tables.rs
nonrenewable: yeah, i figured that might be the case after i read through the rest of things and is totally fine! (since the more_packets_seen has its own behavior!)

adapter/ph/src/adapter_tables.rs
cpacejo: The Rust borrow checker demands it :)  No functional change though.

(Specifically – because `AltEntry`s now contain references to packet buffers, they are restricted to live no longer than the packet buffers do.  The lifetime annotation is necessary to give a syntactic place to "put" this lifetime when referring to `AltEntry`s, and in turn everything that contains an `AltEntry`.)

adapter/ph/src/adapter_manager_worker.rs
nonrenewable: Excellent, makes sense! ✅

adapter/ph/src/adapter_tables.rs
nonrenewable: AH that makes sense, thank you for walking through it since i wasn't sure why you decided to add them now but that explanation helps!

Issue #358, opened 2024-08-22
KM nonces and sequence numbers improvements


mkolehmainen:
Chris comment on initial Noise KM commit-

Thinking more on this… I think our sequence numbers suffice as nonces for mgmt messages… in which case we don't actually need to include the sequence number in the plaintext message with window size = 1 (because we know what the sequence number should be, and it won't decrypt if it's wrong). (And in the case we want to allow a window size > 1 (Sarah is asking Frank for this right now), we'd want to move the sequence number, or at least a suffix of it, outside encryption so we can easily handle reordered packets, else we'll be guessing at nonce values within the window when reordering occurs.)

For transit messages, in the case that encryption is enabled on them, it's tempting to say just use 0 as a nonce, since we don't care about replay for them. But apparently, this actually breaks the security of ChaChaPoly. So for these, we'll still have to have a nonce, but we can grab it from an atomic counter (possibly in batches) to allow for parallelization. But – in any case – the nonce will have to be emitted in plaintext outside the encryption, so the receiver knows what the heck was used.

The catch with all the above is that we are moving trust that a nonce isn't reused from the crypto code to the packet handler to do the right thing. (E.g. when retransmitting a mgmt packet, right now we regenerate it expecting it should have the same body – if this isn't true, it would break encryption!)

So I suppose the "safe" thing to do generally is, choose nonce values from an atomic counter in this module, and paste them (or at least, a suffix of them, relying on loose synchronization to keep track of the prefix) in plaintext on the front of the ciphertext for all messages. The only "downside" I see to this is the nonce value is redundant-ish with the mgmt sequence number. So we're wasting a couple bytes. BUT, I feel like this may be the right choice for now??

(If we do this – I think the notion that (a prefix of) the nonce is emitted before the ciphertext must be part of the RFC, defined per KM scheme. Also whether we "factor out" the mgmt sequence number for a given KM scheme. Since both sides need to agree on whether/how to do this.)

See also org-zpr/zpr-core#991.

PR #411, opened 2024-09-11
send & check sequence numbers on sync reqs

### LINKED ISSUE #389: message sequence numbers

cpacejo:
(accidentally merged #402, this is that re-opened!)

Imbues SyncReqState with a notion of sequence numbers, which are handed out with each window permit granted.  These sequence numbers are plumbed through the sync request->reply loop and matched up on reception.

I don't love this code, and it should be replaced long-term once we figure out our broader story for sequence numbers.  But it's here for now.

Some follow-on TODOs --

* The two halves of SyncReqState, the window state, and the listener state, are mostly independent and can/should be split out.  The window state can then move up a layer and be used for all management messages and/or implemented as part of a transport layer. This is #398.
* We need to generate sequence numbers for async requests also.  This is #401.
* On receipt, sequence numbers of requests should be validated that they are nondecreasing, to prevent replays and out-of-order execution.  This is #400.
* The synchronous request handlers should be async functions, and the logic to send a reply can then be centralized (rather than stuck in each handler function).  This is #399.
* It would be nice if sequence numbers served as nonces.  This is #358.

### DIFF
--- a/adapter/ph/src/zpr.rs
+++ b/adapter/ph/src/zpr.rs
@@ -16,6 +16,10 @@ pub const ZPI_0: Zpi = 0;
 /// Proposed value to allow easily distinguishing packets with plaintext payloads.
 pub const ZPI_ENCRYPTED_HEADER_FLAG: Zpi = 0x80;
 
+/// Message sequence numbers.  Note this is the _abstract_ sequence number.
+/// The physical sequence number included in message headers is a suffix of this.
+pub type SeqNum = u64;
+
 /// The Security Association ID must fit no more than 8 bits.  Note that it shares
 /// space with the ZPI.
 pub type SaId = u8;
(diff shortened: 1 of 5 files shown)


adapter/ph/src/zpr.rs
```diff
@@ -16,6 +16,10 @@ pub const ZPI_0: Zpi = 0;
 /// Proposed value to allow easily distinguishing packets with plaintext payloads.
 pub const ZPI_ENCRYPTED_HEADER_FLAG: Zpi = 0x80;
 
+/// Message sequence numbers.  Note this is the _abstract_ sequence number.
+/// The physical sequence number included in message headers is a suffix of this.
```
svjaffer, line 20: I'm curious why this is necessary

svjaffer (approved):

PR #412, opened 2024-09-11
allow set binary loc in env

mkolehmainen:
Minor change to makefile to make my life easier.


cpacejo (commented):
I don't think this is needed?  Makefile variables by default can be overridden; `make PLANTUML=foo` or `PLANTUML=foo make -e` does what you want

mkolehmainen:
OMG I had no idea!

PR #420, opened 2024-09-12
use atomic counter instead of pulling rand bytes for nonce

### LINKED ISSUE #375: KM_Noise - use a counter instead of rand for nonce
Chris suggests SharedAtomicCounter.

mkolehmainen:
Also tidy up output of the transport sa.

### DIFF
--- a/adapter/ph/src/km_noise.rs
+++ b/adapter/ph/src/km_noise.rs
@@ -37,7 +37,7 @@ use crate::zpr;
 use bytes::{Bytes, BytesMut};
 use curve25519_dalek::montgomery::MontgomeryPoint;
 use openssl::rand::rand_bytes;
-use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::AtomicU64;
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


adapter/ph/src/km_noise.rs
```diff
@@ -37,6 +37,7 @@ use crate::zpr;
 use bytes::{Bytes, BytesMut};
 use curve25519_dalek::montgomery::MontgomeryPoint;
 use openssl::rand::rand_bytes;
+use std::sync::atomic::AtomicUsize;
```
cpacejo: I'd use `AtomicU64` here rather than `AtomicUsize` (just to eliminate any hypothetical architecture-dependence)

cpacejo (dismissed):

adapter/ph/src/km_noise.rs
mkolehmainen: Oh good idea!

cpacejo (approved):
Perfect thanks!!

PR #425, opened 2024-09-13
more better test script logging

mkolehmainen:
This routes the stdout/stderr from the PH instances in the tests into their own files.  If you want to see the output (which happens at the end) you must pass ZPR_TEST_VERBOSE=1 when you run the script.

Eg,
`sudo ZPR_TEST_VERBOSE=1 ./basic-test.sh`

I think this is far clearer than just mushing all the output together.

### DIFF
--- a/.github/workflows/rust-adapter.yml
+++ b/.github/workflows/rust-adapter.yml
@@ -153,7 +153,7 @@ jobs:
       run: |
         cd adapter/ph
         cargo build
-        test/basic-test.sh
+        ZPR_TEST_VERBOSE=1 test/basic-test.sh
 
   capture-integration-test:
     needs: [ph-build, ph-debug-build]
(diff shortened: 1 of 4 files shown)


cpacejo:
To the contrary, I find it often (almost always?) helpful to see the logs interspersed with each other (to track concurrency issues), and also live (to correlate with real time).

As a compromise, instead of hiding the logs, we can print them live _and_ stow them into files by using `tee`:
```
./ph command line |& tee log-a.txt
```

Would this serve your purposes?

mkolehmainen:
Yes @cpacejo love that idea!  (I don't know how you can make sense of the combined output though!)  I'll make that change presently.

cpacejo (approved):
Perfect thanks!!!

PR #427, opened 2024-09-13
remove protocol buffer visa from thrift apis

### LINKED ISSUE #382: visa serialization format - switch from protocol buffers to thrift
Visa service is using protocol buffer to encode visa -- we prefer to use thrift in the ref impl so redo it in thrift.  That is all.

mkolehmainen:
Protocol buffer still used internall by visaservice and by the zplc (which isn't even part of ref impl yet). But at the interface now converts to a thrift version of the visa datastructure.

To make things easier, I also moved all the thrift files into a single source file (vs.thrift).


svjaffer (changes requested):
Sorry, Mathias, but I can't review this, though I did try.

Having code that is generated and doesn't need to be reviewed intermingled with code that was written and does need to be reviewed places a burden on the reviewer to tell them apart, especially in a review this overwhelmingly large.

The comments of a PR is probably not an appropriate place to have this discussion, but I think it's a bad idea to check in generated code in the first place, partly for this reason.  But if we're going to continue doing this, a suggestion: break apart the generated code from the non-generated code, either in separate PRs or clearly delineated commits.  That makes it visually clear what needs to be reviewed and what can just be rubber-stamped.

mkolehmainen:
Ok, I'll try adding some comments in here to see if that makes things clearer...

node/src/vsapi/mod.rs
mkolehmainen, line 1: AUTO GENERATED!!

visaservice/cli/src/vsapi/mod.rs
mkolehmainen, line 1: AUTO GENERATED!!

visaservice/core/pkg/libvisa/transcode.go
mkolehmainen, line 1: This is the significant part of this whole PR.  In which a native visa struct is converted to a thrift one for public consumption.

visaservice/core/pkg/vsapi/vs.go
mkolehmainen, line 1: AUTO GENERATED!!

visaservice/thrift/vs.thrift
mkolehmainen, line 1: The other work for the PR is in here, where I just created a VISA struct in thrift based on the prototype one.

I also combined all the separate thrift files into one because using multiple ones was causing issues which made the sharing of types difficult.

visaservice/core/pkg/libvisa/transcode.go
```diff
@@ -0,0 +1,95 @@
+package libvisa
+
+import (
+	"fmt"
+
+	"github.com/apache/thrift/lib/go/thrift"
+	"google.golang.org/protobuf/proto"
```
svjaffer, line 7: Is this protobuf import just leftover code?

visaservice/thrift/vs.thrift
```diff
@@ -114,24 +180,83 @@ struct ConnectResponse {
...
+//     2. Start the visa service's adatper.  This adapter will present a
+//        certificate to the node that is (a) signed by the ZPR authority and
+//        (b) has a well known CN that tells the node that it is the visa
+//        service's adapter.
+//
+//     3. The node allows this adapter to connect -- even though the node has
+//        has no policy yet.  The pre-built visa includes the hard-coded visa
+//        service adapter's ZPR address.
+//
+//     ~~ Now this visa sercice API kicks in ~~
```
svjaffer: Just a typo: sercice => service

visaservice/core/pkg/libvisa/transcode.go
mkolehmainen: There is still protocol buffer stuff in the visaservice.  For example, the policy (output from the zpl compiler) is contained in a protobuf structure.  The visa service itself still creates protocol buffer visas which are then converted to thrift visas as they leave through the thrift API.

svjaffer (approved):

PR #431, opened 2024-09-16
Put together some initial packet-walk documentation

svjaffer:


packet_walk.md
```diff
@@ -0,0 +1,53 @@
...
+This document is intended to follow the lifetime of packets as they flow through the ZPR network and lay out noteworthy decision points along the way.  It is based on the content of the ZPR RFCs and intended as an implementation guide.
+
+The walk is presented chronologically over the lifetime of a flow and is mainly for exploring expected cases.  This document will not exhaustively cover error cases.
+
+For the purposes of this document specifically, an IP substrate is assumed, but fragmentation is ignored.  A later version may tackle fragmentation in more depth.
+
+## A Packet Arrives
+
+An agent initiates a flow by sending a packet to the Ingress Client Adapter, the gateway into the ZPR network.  At this point, the Adapter has no entry for the flow in its Agent Lookup Table (keyed by the 5-tuple) and the failure of that lookup will indicate that it needs to issue a ZDP Bind Request to the dock.  The Bind Request will be inserted into the packet and the packet will be forwarded to the dock.
+
```
mkolehmainen, line 14: Is the bind request inserted into the packet? Or is the packet inserted into a Bind Request?

packet_walk.md
```diff
@@ -0,0 +1,53 @@
...
+If no matching visa is found, a negative visa is issued instead, only sent to the requesting Dock.
+
+### Dock Receives A Visa Response
+
+The Dock, upon receiving the visa response, will install the visa in its local table.  It will also allocate a Stream ID and send a Bind Reponse to the Adapter.  If the full packet was included in the Visa Response, it will be forwarded on with a Visa Herald message to the next Node.
+
+### Ingress Adapter Receives A Bind Response
+
+The Adapter's outstanding Bind Request finally gets a response.  If the response is successful, the Adapter will map the 5-tuple of the packet to the Stream ID registered in the Bind Response in its Agent Lookup Table.  If the full packet had been too large to forward to the Dock in the Bind Request, the cached packet is now forwarded to the Dock.
+
```
mkolehmainen, line 40: Not all protocols are nice and wait like TCP does for the SYN-ACK.  What does the ingress adapter do if the client application is sending a bunch of packets in the stream assuming they will get through?  

I would think it just drops until it gets a response to its bind request-- -but does mean that the adapter needs to keep track of bind-requests that are in flight.

packet_walk.md
```diff
@@ -0,0 +1,53 @@
...
+## A Packet Arrives And Matches A Flow
+
+When the next packet arrives from the same agent, it now has an entry in the Agent Lookup Table to match.  HMAC is calculated over the packet and appended to it.  The Adapter will compress the packet according to the rules established during the Bind exchange, placing a ZDP transit header in the compressed space.  That packet will then be forwarded on to the Dock.
+
+The Dock should now also have an entry in its own Forwarding Table, which is indexed by the Stream ID in the ZDP header.  It verifies that the packet received matches the visa that authorizes the flow.  If the verification fails, the packet is counted and dropped.  If the verification succeeds, the packet is forwarded on.
+
+If the next ZPR hop is another Node, it does the same, except that it does not need to re-verify the packet.  If, at any point, one of the Nodes does not recognize the Stream ID, but has a visa for it, that is likely the result of a network failover event.  If the Node has neither an entry in its Forwarding Table nor a visa matching the traffic, it is likely that the visa has been revoked and the packet will be dropped.
+
```
mkolehmainen, line 52: Since intermediate nodes (not the nodes attached to ingress or egress adapters) do not look at the packet except to see the stream ID, I don't think there is ever a case where a node would "not recognize stream ID but has a visa".  I think not recognizing stream id means drop packet.  Not sure if there is a management signal that runs backwards to the sending node.

mkolehmainen:
I did a read though and besides my niddly comments, it looks great to me!

packet_walk.md
svjaffer: Good point that I don't mention that.  Yes, we drop subsequent packets, at least for now.

packet_walk.md
svjaffer: Ha, six of one...

Yeah, probably better to phrase it the other way around.

packet_walk.md
svjaffer: We will need a message exchange for failover events, but your point that the intermediary node can't validate the visa stands.

mkolehmainen (dismissed):
So great!

packet_walk.md
```diff
@@ -0,0 +1,55 @@
...
+### Dock Receives A Bind Request
+
+When the Dock receives the bind request it checks its visas to see if one of those captures the request it just received.  If it already has a matching visa, it allocates a new Stream ID for this flow and responds with a Bind Response.
+
+If the dock does not have a visa, it issues a Visa Request to the visa service, packet included.  If the Adapter has sent too many Bind Requests, they may be rate-limited instead.
+
+### Visa Service Receives A Visa Request
+
+The Visa Service fully classifies the packet and checks for a matching visa.  If it finds one, it responds with the visa, again piggy-backed on the triggering packet.
```
nonrenewable: petty - (i imagine this was Mathias' point above) - this is a Visa response that has the packet in it? (rather than the packet with a visa attached to it?) i also may be using incorrect vocab, apologies if so!

packet_walk.md
```diff
@@ -0,0 +1,55 @@
...
+
+The Visa Service also sends the visa (sans packet) to each Node along the route.
+
+If no matching visa is found, a negative visa is issued instead, only sent to the requesting Dock.
+
+### Dock Receives A Visa Response
+
+The Dock, upon receiving the visa response, will install the visa in its local table.  It will also allocate a Stream ID and send a Bind Reponse to the Adapter.  If the full packet was included in the Visa Response, it will be forwarded on with a Visa Herald message to the next Node.
```
nonrenewable: Question here - i've been holding onto it since row 15 😅 - at this point, if we've truncated the packet on the agent side, we don't know here, i imagine? Or we do? I guess I feel like i'm missing something around the value of a partial packet send at that state at the agent, i'm sure it's something obvious to others!

packet_walk.md
nonrenewable: (agreed, this is a good mentioning, i like that it's included!! )

packet_walk.md
```diff
@@ -0,0 +1,55 @@
...
+
+When the Dock receives the bind request it checks its visas to see if one of those captures the request it just received.  If it already has a matching visa, it allocates a new Stream ID for this flow and responds with a Bind Response.
+
+If the dock does not have a visa, it issues a Visa Request to the visa service, packet included.  If the Adapter has sent too many Bind Requests, they may be rate-limited instead.
+
+### Visa Service Receives A Visa Request
+
+The Visa Service fully classifies the packet and checks for a matching visa.  If it finds one, it responds with the visa, again piggy-backed on the triggering packet.
+
+The Visa Service also sends the visa (sans packet) to each Node along the route.
```
nonrenewable, line 27: Mostly out loud musings, not really anything to do about it - Implied but not called out explicitly (i imagine that it's because it's a packet walk, rather than an overview of the system) is that the path decisions are in the Visa server. Maybe one day we have a similar simple README about the Visa service this point could link to re: the route describing how it's generated...

packet_walk.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Ingress Adapter Receives A Bind Response
+
+The Adapter's outstanding Bind Request finally gets a response.  If the response is successful, the Adapter will map the 5-tuple of the packet to the Stream ID registered in the Bind Response in its Agent Lookup Table.  If the full packet had been too large to forward to the Dock in the Bind Request, the cached packet is now forwarded to the Dock.
+
+Any packets that were received on the stream between the issuance of the Bind Request and receiving the Response have been dropped.
+
+### Egress Adapter Receives a Bind Request
+
+When the heralding process reaches the last Node in the path, a Bind Request is issued to the Egress Adapter, containing all the information needed to uncompress packets on this stream and requesting a Stream ID.  The Adapter installs this information into its Dock Lookup Table.  As with previous hops, the resulting Stream ID is installed in the Node's forwarding table.
```
nonrenewable, line 43: Maybe i'm reading this wrong but this sounds like the opposite of the behavior for the ingress adapter is that intentional? I imagined they'd be symmetrical (i.e. the Node decides the stream id to the dock) but maybe they're either not or i'm reading one of these two parts wrong?

packet_walk.md
```diff
@@ -0,0 +1,55 @@
...
+
+## A Packet Arrives And Matches A Flow
+
+When the next packet arrives from the same agent, it now has an entry in the Agent Lookup Table to match.  HMAC is calculated over the packet and appended to it.  The Adapter will compress the packet according to the rules established during the Bind exchange, placing a ZDP transit header in the compressed space.  That packet will then be forwarded on to the Dock.
+
+The Dock should now also have an entry in its own Forwarding Table, which is indexed by the Stream ID in the ZDP header.  It verifies that the packet received matches the visa that authorizes the flow.  If the verification fails, the packet is counted and dropped.  If the verification succeeds, the packet is forwarded on.
+
+If the next ZPR hop is another Node, it does the same, except that it does not need to re-verify the packet.  If, at any point, one of the Nodes does not recognize the Stream ID, the packet is dropped.
```
nonrenewable, line 51: Naive question - the node<->node connectivity isn't described here (and that might be a TODO and address this question i have!) but i'm not sure based on this section if it's the same stream id across all nodes or different ones? (i was going to guess different BUT i could be off and maybe the same is kept intentionally to simplify things)

nonrenewable (dismissed):
(Sorry for the delay, per usual...)

I agree with Mathias, this is awesome!! I really appreciate how you took so many complex pieces/information/decisions and chunked them into digestible statements 🙇 

Of course because I tend to struggle with info in my head, I had questions but I'm guessing they're just because I have not been in all recent conversations/never read the RFCs. Apologies if this is repeating the same info you've done before and happy to talk in person tomorrow if easier to address/discuss, especially if what i'm asking doesn't quite make sense (which is totally possible)

packet_walk.md
```diff
@@ -0,0 +1,55 @@
...
+
+This document is intended to follow the lifetime of packets as they flow through the ZPR network and lay out noteworthy decision points along the way.  It is based on the content of the ZPR RFCs and intended as an implementation guide.
+
+The walk is presented chronologically over the lifetime of a flow and is mainly for exploring expected cases.  This document will not exhaustively cover error cases.
+
+For the purposes of this document specifically, an IP substrate is assumed, but fragmentation is ignored.  A later version may tackle fragmentation in more depth.
+
+## A Packet Arrives
+
+An agent initiates a flow by sending a packet to the Ingress Client Adapter, the gateway into the ZPR network.  At this point, the Adapter has no entry for the flow in its Agent Lookup Table (keyed by the 5-tuple) and the failure of that lookup will indicate that it needs to issue a ZDP Bind Request to the dock.  A Bind Request is created and the packet is inserted into the Bind Request and it is forwarded to the Dock.
```
cpacejo: "A Bind Request is created and the packet is inserted into the Bind Request" I had to read a few times to parse – "A Bind Request is created which includes the packet body" maybe?

packet_walk.md
```diff
@@ -0,0 +1,55 @@
...
+
+If the packet is too big to attach the Bind Request to, it is instead cached and as much of it as can fit is copied onto a new ZDP Bind Request packet.
```
cpacejo: "If the packet is too big to attach the Bind Request to" -> "If the packet is too big to include in the Bind Request"

"as much of it as can fit is copied onto a new ZDP Bind Request packet." -> "as much of it as can fit is included in the Bind Request"

i.e. replacing "inserted", "attach", "copied onto" which all sound like different mechanisms with just "included".

(I feel like you are trying to hint at the fact that the implementation _could_ save a copy by sticking the bind request header onto the front of the packet but I feel that's dipping into implementation detail territory.)

packet_walk.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Dock Receives A Bind Request
+
+When the Dock receives the bind request it checks its visas to see if one of those captures the request it just received.  If it already has a matching visa, it allocates a new Stream ID for this flow and responds with a Bind Response.
```
cpacejo: "bind request" -> "Bind Request" for consistency

"captures" -> "matches"

"Stream ID" -> "Tether ID"

I think here we should mention that the Bind Response also indicates how to compress the packet.  (Expansion is mentioned below.)

packet_walk.md
cpacejo: Visa service matches against policy and creates new visas, no?

packet_walk.md
cpacejo: "visa response" -> "Visa Response" for consistency

"local table" -> specifically, the Dock Forwarding Table

"Bind Reponse" -> "Bind Response" (typo)

"Stream ID" -> "Tether ID"

"If the full packet was included [...]" -> "A Visa Herald message will be sent to the next Node.  This message will include the full packet if it was included in the Visa Response."

Aside – I'm not sure the purpose of forwarding the packet in the Herald message.  If we got the packet in our Visa Response, presumably the egress dock will get the packet in its Visa Response also.  (We can discuss this offline)

packet_walk.md
```diff
@@ -0,0 +1,55 @@
...
+
+If no matching visa is found, a negative visa is issued instead, only sent to the requesting Dock.
+
+### Dock Receives A Visa Response
+
+The Dock, upon receiving the visa response, will install the visa in its local table.  It will also allocate a Stream ID and send a Bind Reponse to the Adapter.  If the full packet was included in the Visa Response, it will be forwarded on with a Visa Herald message to the next Node.
+
+### Ingress Adapter Receives A Bind Response
+
+The Adapter's outstanding Bind Request finally gets a response.  If the response is successful, the Adapter will map the 5-tuple of the packet to the Stream ID registered in the Bind Response in its Agent Lookup Table.  If the full packet had been too large to forward to the Dock in the Bind Request, the cached packet is now forwarded to the Dock.
```
cpacejo: "response is successful" -> "response indicates success"

"Stream ID" -> "Tether ID"

packet_walk.md
cpacejo, line 43: "uncompress" -> "expand"  (to match terminology used in the RFC)

"Stream ID" -> "Tether ID"

"the heralding process" -> I don't see this discussed above.  Presumably this means, each node which gets a Visa Herald message, in turn issues one to the next node in the path.

Aside – I'm curious how we will resolve the race between the visa installation process and the visa heralding process!  (I.e. if a herald outraces the installation, what do we do with the herald request?)

packet_walk.md
cpacejo, line 27: What message is used to send the visa?

packet_walk.md
```diff
@@ -0,0 +1,55 @@
...
+
+Any packets that were received on the stream between the issuance of the Bind Request and receiving the Response have been dropped.
+
+### Egress Adapter Receives a Bind Request
+
+When the heralding process reaches the last Node in the path, a Bind Request is issued to the Egress Adapter, containing all the information needed to uncompress packets on this stream and requesting a Stream ID.  The Adapter installs this information into its Dock Lookup Table.  As with previous hops, the resulting Stream ID is installed in the Node's forwarding table.
+
+## A Packet Arrives And Matches A Flow
+
+When the next packet arrives from the same agent, it now has an entry in the Agent Lookup Table to match.  HMAC is calculated over the packet and appended to it.  The Adapter will compress the packet according to the rules established during the Bind exchange, placing a ZDP transit header in the compressed space.  That packet will then be forwarded on to the Dock.
```
cpacejo: "HMAC" -> "A2A HMAC"

I think we should also note above when the A2A HMAC is calculated for the first packet.

packet_walk.md
```diff
@@ -0,0 +1,55 @@
...
+
+The Dock should now also have an entry in its own Forwarding Table, which is indexed by the Stream ID in the ZDP header.  It verifies that the packet received matches the visa that authorizes the flow.  If the verification fails, the packet is counted and dropped.  If the verification succeeds, the packet is forwarded on.
```
cpacejo: "Forwarding Table" -> "Dock Forwarding Table", to match RFC terminology

"Stream ID" -> "Tether ID"

"the packet is forwarded on" -> "... using the Link ID and Stream ID indicated by the Dock Forwarding Table"

packet_walk.md
cpacejo, line 51: "it does not need to re-verify the packet" -> I thought each Node did in fact re-verify packets against the visa?

packet_walk.md
```diff
@@ -0,0 +1,55 @@
...
+
+If the next ZPR hop is another Node, it does the same, except that it does not need to re-verify the packet.  If, at any point, one of the Nodes does not recognize the Stream ID, the packet is dropped.
+
+### Packet Reaches The Egress Adapter
+
+When the packet arrives at its last ZPR hop, the Egress Client Adapter, it is uncompressed according to the rules installed during the Bind exchange.  After that, its HMAC is checked for packet integrity.  Should the integrity check fail, the packet will be dropped.  Should it succeed, the packet is forwarded on to the Agent.  A reverse packet should flow seamlessly through all of the same stages and checks as this non-first forward packet.
```
cpacejo: "HMAC" -> "A2A HMAC"

"A reverse packet should flow [...]" -> I feel like this is carrying a lot of weight!  It might be useful to at least distinguish between the reverse Binding process (which I believe needs to be explicit) and the fact that Visas are stateful and thus do not need to be rehearalded.

cpacejo (dismissed):
Overall looks good and matches my understanding!  Mostly terminology nits on my part.

packet_walk.md
svjaffer: Well, so, stream IDs are directional and the receiver always decides the stream ID for what it's receiving.  Based on our conversations last week, these could all be stream ID exchanges instead of just a request/response pair in each direction, but that hasn't been officially decided yet, even though that's probably where we will end up.

packet_walk.md
svjaffer: It is not the same stream ID.  Each set of peers exchanges stream IDs with each other.  Any given stream ID is only meaningful between those two peers in a specific direction.

packet_walk.md
svjaffer: Much better wording, thanks!

packet_walk.md
svjaffer: Yeah, that was where I was coming from, but you are correct that that's too much of an implementation detail.

packet_walk.md
svjaffer: I think that's something we need to have a new message type for.  I'll leave this as-is and we can fill it in later once we've had that discussion.

packet_walk.md
svjaffer: I thought they couldn't re-verify because they don't speak the compression?

packet_walk.md
svjaffer: Yes, path decisions are part of the visa creation and that is outside the scope of this document.  I agree that having a similar document for the Visa Service would be good when we get to that point :)

packet_walk.md
svjaffer: I've put in a clarification earlier on that the Visa Service will only include the packet in the Visa Response if it is a full packet, which it should know because it's already doing packet inspection/classification.  You're right that the Dock doesn't make that decision.

packet_walk.md
svjaffer: Corrected the language here

packet_walk.md
nonrenewable: Awesome, makes sense!

packet_walk.md
nonrenewable: oh that consistency point makes way more sense 😂 thank you!

packet_walk.md
nonrenewable: Gotcha! and as i think about it what Stream ID is as a concept is outside the scope of this doc. Thank you!

nonrenewable:
Thank you for answering all my questions! I don't feel qualified to approve it but ✅ from me

packet_walk.md
cpacejo: For rate-limiting and the reverse-stateful stuff at least I think each hop re-verifies.

cpacejo (approved):
Looks good thanks!

PR #437, opened 2024-09-16
automatically flush pcap file when queue is empty

### LINKED ISSUE #436: pcap: flush automatically if pcap queue empty
This will allow viewing live PCAPs in Wireshark.  We'll still keep the manual FLUSH command around to ensure reliable integration testing.

cpacejo:
This allows FIFOs to be used smoothly for packet capture.  All-but eliminates the need for an explicit flush without impacting capture performance.  (Explicit flush is still needed to eliminate race conditions in testing.)


mkolehmainen (approved):
Neato!

PR #438, opened 2024-09-16
Better integrate key management state into peer table

### LINKED ISSUE #388: peer_table - put the security association state directly into the slab
This is currently hangin as a DashMap on the PeerTable struct.

mkolehmainen:
Remove the parallel dashmap thing that was tracking link_id->KmState.  Instead, the KmState is directly the PeerState which is stored as a peer table entry.

Note it is private since it is manipulated exclusively though methods on the table.

The TransportSA, which is used in fastpath is now stored in a RcuBox which presumably is better performing than a mutex.  Not sure I've used that correctly.

The KmHandle -- which is state that the KeyManager uses for management activities is behind a regular mutex but is only used in management path I believe.

### DIFF
--- a/adapter/ph/src/peer_table.rs
+++ b/adapter/ph/src/peer_table.rs
@@ -1,6 +1,6 @@
 #![allow(dead_code)]
 use crate::dock_tables::DockForwardingTable;
-use crate::km::{KeyManager, KmTransportSA, UnimplCodec, ZPIPair};
+use crate::km::{KeyManager, KmTransportSA};
 use crate::queues;
 use crate::rcu::{RcuBox, RcuGuard};
 use crate::sync_req;
@@ -8,9 +8,6 @@ use crate::zpr::{LinkId, SubstrateAddr};
 use cslab::{RcuCslab, RcuCslabReader};
 use dashmap::DashMap;
 use std::future::Future;
-use std::sync::atomic::AtomicBool;
-use std::sync::atomic::Ordering;
-use std::sync::Arc;
 use std::sync::Mutex;
 use std::sync::MutexGuard;
 use tokio::sync::mpsc;
@@ -44,8 +41,7 @@ pub struct PeerState<'pktbuf> {
 // Key Management state per peer.
 struct PeerKmState {
     handle: Mutex<Option<KmHandle>>, // Once the KM state machine starts, it's join handle and related info is stashed here.
-    sa_established: AtomicBool,      // if TRUE then `transport_sa` is valid
-    transport_sa: RcuBox<KmTransportSA>,
+    transport_sa: RcuBox<Option<KmTransportSA>>,
 }
 
 /// The Key Management "handle" is used by the km_multiplexor to hold per-link
@@ -61,15 +57,7 @@ impl PeerKmState {
     fn new() -> Self {
         Self {
             handle: Mutex::new(None),
-            sa_established: AtomicBool::new(false),
-            transport_sa: RcuBox::new(KmTransportSA {
-                sa_id: 0,
-                recv_zpis: ZPIPair::new_zero(),
-                send_zpis: ZPIPair::new_zero(),
-                send_hmac_key: [0; 32],
-                recv_hmac_key: [0; 32],
-                codec: Arc::new(UnimplCodec::new()),
-            }),
+            transport_sa: RcuBox::new(None),
         }
     }
 }
@@ -219,8 +207,7 @@ impl<'pktbuf> PeerTable<'pktbuf> {
         let entry = self
             .get(link_id)
             .ok_or(SecurityAssocaitionStateError::NoAssociationForLink)?;
-        entry.km_state.transport_sa.write(sa);
-        entry.km_state.sa_established.store(true, Ordering::Release);
+        entry.km_state.transport_sa.write(Some(sa));
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


adapter/ph/src/peer_table.rs
```diff
@@ -254,51 +246,66 @@ impl<'pktbuf> PeerTable<'pktbuf> {
...
-        } else {
-            None
+        let entry = self.get(link_id)?;
+
+        // If SA is set, un-set it.
+        if entry.km_state.sa_established.load(Ordering::Acquire) {
+            entry
+                .km_state
+                .sa_established
+                .store(false, Ordering::Release);
```
cpacejo: `Release` does nothing here – there are no operations prior to this which it is releasing.

To order this store and the following `take`, you could use a release fence operation between the two (which causes the `take` to act like a release operation), but that won't fix the race – see my comment above.

adapter/ph/src/peer_table.rs
```diff
@@ -254,51 +246,66 @@ impl<'pktbuf> PeerTable<'pktbuf> {
...
         link_id: LinkId,
     ) -> Option<KmTransportSA> {
-        match self.link_to_km_state.get(&link_id) {
-            Some(km_state) if km_state.sa_established.load(Ordering::Acquire) => {
-                Some(km_state.transport_sa.clone())
-            }
-            _ => None, // either not found or not established
+        let entry = self.get(link_id)?;
+        if entry.km_state.sa_established.load(Ordering::Acquire) {
+            return Some(entry.km_state.transport_sa.get().clone());
```
cpacejo: This races `remove_km_state()` – you could read the flag as true, and then much time passes (during which `transport_sa` is cleared), and then try to read the transport.

There's no way to "solve" this with just loads and stores.  I don't know how to formally classify this problem but I think of it as the difference between a "positive" information flow (here is a thing, it's now safe to act on it forevermore) and "negative" information flow (there is no longer a thing here, stop acting on it).  The latter requires a synchronization path _back_ from the consumer.

To solve this, you need some sort of synchronization structure (e.g. mutex or channel) to link these two variables.  Or, make them a single variable – see my comment on the structure itself.

adapter/ph/src/peer_table.rs
```diff
@@ -55,19 +57,19 @@ pub struct KmHandle {
...
+            handle: Mutex::new(None),
             sa_established: AtomicBool::new(false),
-            transport_sa: KmTransportSA {
+            transport_sa: RcuBox::new(KmTransportSA {
                 sa_id: 0,
                 recv_zpis: ZPIPair::new_zero(),
                 send_zpis: ZPIPair::new_zero(),
                 send_hmac_key: [0; 32],
                 recv_hmac_key: [0; 32],
                 codec: Arc::new(UnimplCodec::new()),
```
cpacejo: I think the `Arc` here should be unnecessary now that this is in an `RcuBox`.  The value in the `RcuBox` (returned via `.get()`) can be held as long as the caller needs without preventing newer versions of that thing from being placed there by `.write()`.

adapter/ph/src/peer_table.rs
```diff
@@ -38,12 +38,14 @@ pub struct PeerState<'pktbuf> {
...
 }
 
+// Key Management state per peer.
 struct PeerKmState {
-    handle: Option<KmHandle>,
-    sa_established: AtomicBool, // if TRUE then `transport_sa` is valid
-    transport_sa: KmTransportSA,
+    handle: Mutex<Option<KmHandle>>, // Once the KM state machine starts, it's join handle and related info is stashed here.
+    sa_established: AtomicBool,      // if TRUE then `transport_sa` is valid
+    transport_sa: RcuBox<KmTransportSA>,
```
cpacejo: This mechanism is inherently/inescapably racy (see my explanation below).

Instead, possible to use an `RcuBox<Option<KmTransportSA>>` here?

cpacejo (changes requested):
The broad strokes are correct!  But I think there are some issues with the synchronization though (which can be easily remedied).

Also the inner `Arc` I think can safely be removed.

adapter/ph/src/peer_table.rs
mkolehmainen: I understand the issue here, I just wonder if it matters so long as trasnport_sa is wholly written (which I think the RCUBOX does?) -- like there is no danger to finding a half written state in there, right?

If established is read as true, then the caller thinks that the state is established.  If it turns false, and the rcubox is updated, before the caller actually fetches the state, then they get an unestablished state which won't work.  Maybe enough for users of the state to check that it's valid before using, eg check if sa_id>0. 

If established is false, well that's fine.

But also, I do think option<rcu<transport>> is good idea.  So i'm going to make that change.

adapter/ph/src/peer_table.rs
mkolehmainen: That sound right, but the pointer makes it possible to easily support clone which is required in the way that we use the transport later in  fastpath: clone it, use it, chuck it.  Also codec is just a trait, so there are other odd errors that come up if I try to remove the pointer, like the fact that it doesn't have a known size.

cpacejo: Oh I forgot it's a trait!  Yes this is fine but I might sniff around if there's a way to avoid the Arc in my own time.

cpacejo (approved):
Looks good thanks!!

PR #443, opened 2024-09-17
"fix?" for recent timing issues in tests

mkolehmainen:
So... I think the "timing" issue is really inside the KM mechanism.

1. adapter sends handshake message - adapter-state=HS
2. node receives handshake message - causes node to switch to state=TRANSPORT
3. node sends handshake response - under transport state -- or not -- depending on some timings.
4. adapter gets handshake response - adapter state = HS but message might arrive under transport = error!

So what I have done (for now, will add a follow up task) is to always send key management messages under zpi-0.


cpacejo (approved):
Ah good find!  Interested to see the final solution.  Is there a diagram of this state machine anywhere?  We might be able to apply formal modeling to it to discover any other potential race conditions.

PR #445, opened 2024-09-18
disable KM in basic test, it does not work reliably right now

cpacejo:
`basic-test.sh` only works maybe 1 out of 5 retries with the KM enabled.  Disabling for now.


cpacejo:
Ah nevermind!!!  I forgot to build after fetching.  Too early in the morning!

PR #448, opened 2024-09-18
reserve stream ID 0

### LINKED ISSUE #432: reserve stream ID 0

cpacejo:
Since the DLT is the source of all stream IDs (at the moment), that's the only code I needed to touch.

The `wrapping_sub` is used to map stream ID 0 to usize(-1), which is a guaranteed lookup fail (not possible to allocate such a high stream ID).


mkolehmainen (dismissed):
Looks good to me!

cpacejo:
Oops, I need to do this for the DFT as well!!

mkolehmainen (approved):
great!

PR #449, opened 2024-09-18
certificate exchange procedure

mkolehmainen:
This is part 1 of a task to add certificate exchange into the key management.  Ultimate goal is that after KM is done, the node will be able to examine a certificate from the adapter.  (or from another node).

I have not attempted to share the node cert with the adapter.  I'm assuming that the adapter can fetch a node cert from somewhere out of band.

This code change here is just a stand alone "exchange" module.  This has not been integrated into key management.

### DIFF
--- a/adapter/ph/Cargo.toml
+++ b/adapter/ph/Cargo.toml
@@ -30,6 +30,7 @@ tracing = "0.1.40"
 tracing-subscriber = "0.3.18"
 zerocopy = { version = "0.7.34", features = ["byteorder", "derive"] }
 zpr-ext = { path = "../../zpr-ext", features = ["bytes", "tokio", "tokio-tun", "zerocopy"] }
+flate2 = { version = "1.0.33", features = ["zlib-ng"] }
 
 [dev-dependencies]
 
--- a/adapter/ph/src/km_cert_exchange.rs
+++ b/adapter/ph/src/km_cert_exchange.rs
@@ -56,14 +56,11 @@ use tracing::error;
 use zerocopy::byteorder::network_endian::*;
 use zerocopy::{AsBytes, FromBytes, FromZeroes, Unaligned};
 
-
-
 const CERT_FINGERPRINT_LEN: usize = 20; // sha-1
 
 // A note on signature length. A 2048 bit RSA key will have a 256 byte signature.
 // Since we always put the signatures last on the message, we don't need to
 // specify a length.  So the code should work fine with various key sizes.
-
 
 const MSG_BUF_SIZE: usize = 4096;
 
@@ -104,7 +101,7 @@ struct CertHelloHdr {
 struct CertResponseHdr {
     pub nonce: U64,
     pub cert_fingerprint: [u8; CERT_FINGERPRINT_LEN], // SHA-1 certificate fingerprint
-    // Followed by the SHA256 RSA signature (256 bytes)
+                                                      // Followed by the SHA256 RSA signature (256 bytes)
 }
 
 #[derive(Debug, PartialEq)]
--- a/adapter/ph/src/km_cert_exchange.rs
+++ b/adapter/ph/src/km_cert_exchange.rs
@@ -1,0 +1,155 @@
+
+///! Key Management Certificate Exchange
+///
+///
+///! This is structured so that it could be broken out into an interface if we need
+///! to.  The general idea is that this is how the payloads are created and processed
+///! during key exchange.  Our noise handshake as just two messages, so this assumes
+///! two messages.
+///!
+///! Message 1 is from initiator to responder which includes:
+///! - The initiator's certificate
+///! - A timestamp (unix epoch millis)
+///! - A signature over the above using the initiator's private key
+///!
+///! Message 2 is a reply from responder to initiator which includes:
+///! - The fingerprint of the responders certificate.  We assume that the responders
+///!   certificate is known to the initiator and can be retrieved out of band.
+///! - The initiator timestamp (copies from initator payload)
+///! - A signature over the above using the responder's private key
+///!
+
+
+
+use bytes::{Bytes, BytesMut, BufMut};
+use zerocopy::{AsBytes, FromBytes, FromZeroes, Unaligned};
+use zerocopy::byteorder::network_endian::*;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+
+#[derive(Debug)]
+pub enum CertExchangeError {
+    CertificateError,
+    InvalidPayloadError,
+    ShortPayloadError,
+    SignatureError,
+}
+
+#[derive(FromZeroes, FromBytes, AsBytes, Unaligned)]
+#[repr(packed)]
+struct CertHelloHdr {
+    pub timestamp: U64,
+    pub cert_len: U16,
+    // Followed by the cert bytes (PEM, compressed?)
+    // Followed by the PKCS1_1.5_SHA256_signature (32 bytes)
+}
+
+#[derive(FromZeroes, FromBytes, AsBytes, Unaligned)]
+#[repr(packed)]
+struct CertResponseHdr {
+    pub timestamp: U64,
+    pub cert_fingerprint: [u8; 20], // SHA-1 certificate fingerprint
+    // Followed by the PKCS1_1.5_SHA256_signature (32 bytes)
+}
+
+
+pub struct KmCertExchange;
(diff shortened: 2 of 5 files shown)


adapter/ph/Cargo.toml
```diff
@@ -30,6 +30,7 @@ tracing = "0.1.40"
 tracing-subscriber = "0.3.18"
 zerocopy = { version = "0.7.34", features = ["byteorder", "derive"] }
 zpr-ext = { path = "../../zpr-ext", features = ["bytes", "tokio", "tokio-tun", "zerocopy"] }
+flate2 = { version = "1.0.33", features = ["zlib-ng"] }
```
mkolehmainen, line 34: Curious what you think about this dependency.   Works ok on my linux machine.  If we compress those certs we save about 20% in my very limited testing.

cpacejo: I think the native Rust backend would be fine/preferable.  This shouldn't be a performance-sensitive path, and zlib isn't a stranger to memory safety issues, e.g. this buffer overflow in `inflate` a couple years back: https://nvd.nist.gov/vuln/detail/CVE-2022-37434

I wonder though about the safety of using this as part of key exchange though.  If we're using this to decompress certificates, by definition we haven't yet established trust with the peer.  Decompression is the sort of code path that I feel has a large surface area for vulnerabilities (as evidenced by the CVE above).  So maybe best that we don't do this?  I'll think a bit.

adapter/ph/src/km_cert_exchange.rs
```diff
@@ -0,0 +1,799 @@
...
+//!
+//!  1. Initiator creates a message.  This includes the initiator's
+//!     signed certificate.  The message is signed with the initiator's
+//!     private RSA key.
+//!
+//!  2. Responder processes the initial message, checking that the
+//!     certificate furnished is signed by the proper authority and
+//!     that the message signature is valid.  If all goes well the
+//!     responder creates a response for the initiator which includes
+//!     the _fingerprint_ of the responders signed certificate. The
```
cpacejo, line 27: Why not include the full certificate?  Then no need to predistribute the responder's cert.

(Initiator should expect a given Common Name in the cert, to determine we're talking to whom we expect.)

adapter/ph/src/km_cert_exchange.rs
```diff
@@ -0,0 +1,799 @@
...
+//!
+//! The _responder_ has no information about the _initiator_ except that
+//! it will expect the _initiators_ certificate to be signed by a
+//! certificate authority known to the _responder_.  The certificate for
+//! the certificate authority is accessible to the _responder_ in some
+//! out of band fashion.
+//!
+//!  1. Initiator creates a message.  This includes the initiator's
+//!     signed certificate.  The message is signed with the initiator's
+//!     private RSA key.
```
cpacejo, line 21: This feels dangerous to me.  A rogue adapter could copy the plaintext of this message (nonce + cert + signature) and replay it to another adapter to impersonate a node.  (Note the nonce is likely just some small integer.)

I think there needs to be some mechanism which links the crypto material of the certificates to that of the session itself.  E.g. TLS [does this](https://www.ibm.com/docs/en/ibm-mq/9.3?topic=tls-how-provides-identification-authentication-confidentiality-integrity) by communicating the session's symmetric key using the asymmetric keys.  I think a challenge/response mechanism is acceptable also ("here's some random bytes I encrypted with your public key, decrypt it and send it back!").

cpacejo (commented):
I feel like we're inventing new crypto here, which seems dangerous.

IKEv2 and DTLS both provide for certificate exchange.  Does Noise also?  At the very least, Noise provides for public key exchange, which I think is sufficient – the Noise public key can be matched against the public key in the provided certificate to demonstrate that they're referring to the same thing.  (It is possible to create certificates using Noise's public keys, I did so as an exercise some weeks ago!)

adapter/ph/Cargo.toml
mkolehmainen: I agree.  The compression isn't essential.

adapter/ph/src/km_cert_exchange.rs
mkolehmainen: Maybe it's not clear that these payloads are all in an encrypted channel the trust of which is based on the holding of a public (noise) key by the initiator (adapter).  The goal of this exchange is just to deliver the certs to each side, and I added the signing as sort of "proof that I have the private key".  You shouldn't be able to replay since the messages are encrypted and part of  handshake state.  Which begs the question of why do I need the nonce stuff in there at all?  Probably it is overkill.

mkolehmainen: Maybe what I was thinking is that a node gets compromised, still has the cert but no longer has the private key that maybe was stored in ram??

mkolehmainen: And to your point about delivering certs from node to adapter. Could do.

cpacejo: > Maybe it's not clear that these payloads are all in an encrypted channel the trust of which is based on the holding of a public (noise) key by the initiator (adapter).

Before certs are exchanged though there's no trust, only privacy.  Anyone can reach this stage of the key exchange, since anyone can have a Noise keypair.

> The goal of this exchange is just to deliver the certs to each side

If the certs carry the Noise public key that the SA is using, then this in and of itself is sufficient and correct.  Since now we use our trust in the CA (who signed the cert) to link the info in the cert (i.e. the Common Name) to the Noise public key which the session is based on.  So now we know who's talking to us.

> You shouldn't be able to replay since the messages are encrypted and part of handshake state

Say we have Good Node, Good Adapter, and Rogue.

Rogue connects to Good Node.  They establish an SA (with Good Node's Noise keypair and Rogue's Noise keypair) and reach the cert exchange phase without problem (which Rogue can do, all Rogue needs is a Noise keypair which they can generate at will).  Good Node now sends their certificate signed with Good Node's private key.  Rogue can decrypt/read this because the SA was established correctly.  Rogue makes a copy.

Rogue now impersonates Good Node (MitM between Good Node and Good Adapter… say DNS hijacking in a café's public wifi).  Good Adapter connects to Rogue.  They establish an SA (with Good Node's Noise keypair and Rogue's Noise keypair) and reach the cert exchange phase.  Rogue now sends the copy it made above, of Good Node's certificate signed with Good Node's private key.  Rogue doesn't actually needs to sign it, the copy it made before validates just fine because the thing it's signing (Good Node's certificate) is unchanged.  Good Adapter now trusts Rogue is Good Node.

Nonces don't help because they are typically small numbers allocated sequentially.

There are two correct ways around this I believe:

1) Rather than Good Node signing the certificate with its private key, Good Node should sign a _challenge_ sent by its peer (Rogue or Good Adapter) with its private key, which the peer then validates with the public key from Good Node's certificate.  The challenge is a cryptographically random value (e.g. 256 bytes of random data).  There's no way for Rogue to guess what challenge Good Adapter will eventually generate when it tries to connect, so Rogue can't trick Good Node into giving a valid signature for this challenge.  (It can't even do this as an active MitM, because of the privacy of the channel established by Noise.)

2) Rather than exchanging _arbitrary_ certificates, the certificate a node/adapter presents _must be_ a certificate for its Noise public key.  The certificates can just be sent, no need to sign or anything.  We already know we're talking across a channel to someone who owns the corresponding Noise private key.  The certificate then merely makes the connection between the Noise public key and the Common Name.  This connection (the certificate) is signed by the CA which we trust.  Rogue can't impersonate this because it can't establish a Noise SA using the corresponding Noise private key.

mkolehmainen: Hmm.  There are valid points in here for sure.  One minor thing is that Rogue cannot just impersonate the node unless it also tricks the adapter into using a different noise key (which, if the public key is fetched from DNS record or something is possible).  So I don't think that invalidates what you are saying.

I think I like #2 up there.  How do we get a noise key into a certificate?  Some sort x509 property?

The implication I think is that in order to connect an adapter to the ZPR, you need to first generate a noise keypair, then request a cert signed by our authority that includes the public component.  All that means for us is that your adapter/node has been blessed.   

To be clear, identity of agents is a separate thing performed by visa service that may involve passwords, certificates, ubikeys, etc.

mkolehmainen:
I'm going to close this and come back with it later with new noise-key certs.

PR #452, opened 2024-09-19
replace SA Arc clone with Rcu get in fastpath

### LINKED ISSUE #451: Eliminate need for Arc in KmTransportSA

cpacejo:
Because the fastpath uses Rcu to get at the SA, clone isn't needed, we can instead rely on Rcu keeping the SA around without holding a lock.  This is a small refactor which allows that to happen.

(I haven't tackled the broader change of replacing the `Arc<dyn Codec>` with a `Box<dyn Codec>` as there are some other dependencies, possibly only in test-land, which rely on it being `Arc`, but that matters less to me than reducing the synchronization overhead in the fastpath.)

### DIFF
--- a/adapter/ph/src/fastpath.rs
+++ b/adapter/ph/src/fastpath.rs
@@ -318,9 +318,7 @@ fn substrate_egress_common<'pktbuf>(
         );
         transport_sa = None;
     } else {
-        transport_sa = asm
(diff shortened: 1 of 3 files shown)


mkolehmainen (approved):
Excellent!

PR #453, opened 2024-09-19
diagrams for fun

mkolehmainen:
Two diagrams:
- the KM state machine
- the noise message exchange 

Definitely overkill but fun to play with plantuml!


cpacejo (approved):
Nice thanks for making these!

PR #454, opened 2024-09-20
implement basic PKI tool

### LINKED ISSUE #45: tool to create a CSR (and keys)
New zpr tool we can use to create a keypair and a csr.  For now we just use a CN in the CSR subject (TODO: that's fine, right?).

The command should be flexible enough to generate a new keypair or use an existing key.  So for example:

For example,
```bash
$ ./zprauth create
Usage:
   create csr [--key private key] CN 
   create key
```

```bash
$ ./zprauth create csr adapter1.blah
wrote private key to adapter1.blah.private
wrote public key to adapter1.blah.public
wrote csr to adapter1.blah.csr
```

```bash
$ ./zprauth create csr --key mykey.private adapter1.blah
wrote csr to adapter1.blah.csr
``` 

Maybe even just `create key` should work too:

```bash
$ ./zprauth create key 
wrote private key to adapter.private
wrote public key to adapter.public
```

Bonus points for coming up with a better name than "zprauth". 

This could probably be expanded to run CA duties too, which would take some of the pain out of setup and testing for those who are not running a full-fledges key management system already.  But that's another story.

cpacejo:
PKI tool
PKI tool
does whatever a PKI tool does

### DIFF
--- a/tools/zpr-pki
+++ b/tools/zpr-pki
@@ -112,7 +112,7 @@ function do_gensignedcert() {
 	SUBJ=$3
 	DAYS=$4
 
-	openssl x509 -req -CA "$CA" -CAkey "$CAKEY" -copy_extensions copyall -days "$DAYS" -force_pubkey /dev/stdin \
+	exec openssl x509 -req -CA "$CA" -CAkey "$CAKEY" -copy_extensions copyall -days "$DAYS" -force_pubkey /dev/stdin \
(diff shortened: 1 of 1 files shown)


mkolehmainen (dismissed):
Looks good to me. Can't wait to try it!

mkolehmainen (approved):
Excellent!

PR #455, opened 2024-09-20
Mk km certs

### LINKED ISSUE #457: KM - decide cert message format
Before we can pass certs ( #419 ) this step just comes up with the cert message format and checking procedure.

mkolehmainen:
Certificate exchange routines -- not actually hooked up to KM.  Just functions to create and parse payload.

### DIFF
--- a/adapter/ph/src/km_cert_exchange.rs
+++ b/adapter/ph/src/km_cert_exchange.rs
@@ -56,14 +56,11 @@ use tracing::error;
 use zerocopy::byteorder::network_endian::*;
 use zerocopy::{AsBytes, FromBytes, FromZeroes, Unaligned};
 
-
-
 const CERT_FINGERPRINT_LEN: usize = 20; // sha-1
 
 // A note on signature length. A 2048 bit RSA key will have a 256 byte signature.
 // Since we always put the signatures last on the message, we don't need to
 // specify a length.  So the code should work fine with various key sizes.
-
 
 const MSG_BUF_SIZE: usize = 4096;
 
@@ -104,7 +101,7 @@ struct CertHelloHdr {
 struct CertResponseHdr {
     pub nonce: U64,
     pub cert_fingerprint: [u8; CERT_FINGERPRINT_LEN], // SHA-1 certificate fingerprint
-    // Followed by the SHA256 RSA signature (256 bytes)
+                                                      // Followed by the SHA256 RSA signature (256 bytes)
 }
 
 #[derive(Debug, PartialEq)]
--- a/adapter/ph/src/km_cert_exchange.rs
+++ b/adapter/ph/src/km_cert_exchange.rs
@@ -1,0 +1,155 @@
+
+///! Key Management Certificate Exchange
+///
+///
+///! This is structured so that it could be broken out into an interface if we need
+///! to.  The general idea is that this is how the payloads are created and processed
+///! during key exchange.  Our noise handshake as just two messages, so this assumes
+///! two messages.
+///!
+///! Message 1 is from initiator to responder which includes:
+///! - The initiator's certificate
+///! - A timestamp (unix epoch millis)
+///! - A signature over the above using the initiator's private key
+///!
+///! Message 2 is a reply from responder to initiator which includes:
+///! - The fingerprint of the responders certificate.  We assume that the responders
+///!   certificate is known to the initiator and can be retrieved out of band.
+///! - The initiator timestamp (copies from initator payload)
+///! - A signature over the above using the responder's private key
+///!
+
+
+
+use bytes::{Bytes, BytesMut, BufMut};
+use zerocopy::{AsBytes, FromBytes, FromZeroes, Unaligned};
+use zerocopy::byteorder::network_endian::*;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+
+#[derive(Debug)]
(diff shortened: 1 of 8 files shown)


adapter/ph/src/km_cert_exchange.rs
```diff
@@ -0,0 +1,396 @@
...
+pub enum ParseError {
+    PEMCertNotFound,
+    PEMFormatError,
+    IOError(std::io::Error),
+}
+
+#[derive(FromZeroes, FromBytes, AsBytes, Unaligned)]
+#[repr(packed)]
+struct CertExchgHdr {
+    pub cert_len: U16,
```
cpacejo, line 53: Eventually we might want to allow for a certificate chain to be transferred (this is necessary if an intermediate CA is used) but not necessary for the time being.

adapter/ph/src/km_cert_exchange.rs
```diff
@@ -0,0 +1,396 @@
...
+    pub fn new_from_paths(
+        cert_file: &Path,
+        authority_cert_file: &Path,
+    ) -> Result<Self, ParseError> {
+        let cert = load_cert(cert_file)?;
+        let authority_cert = load_cert(authority_cert_file)?;
+        Ok(KmCertExchange::new(cert, authority_cert))
+    }
+
+    /// Like [KmCertExchange::new] but takes the paths to the various PEM files.
```
cpacejo: "paths" -> "contents"

adapter/ph/src/km_cert_exchange.rs
```diff
@@ -0,0 +1,396 @@
...
+        Ok(KmCertExchange::new(cert, authority_cert))
+    }
+
+    /// Write a cert exhange payload into the supplied buffer, returns number of bytes written.
+    ///
+    /// ## Errors
+    /// - [CertExchangeError::ShortPayloadError] - the buffer is too short to hold the payload.
+    pub fn write_payload(&mut self, buf: &mut [u8]) -> Result<usize, CertExchangeError> {
+        let cert_der = self.local_cert.to_der().unwrap();
+        let sz = cert_der.len() as u16;
```
cpacejo, line 113: May need a check here that the length doesn't overflow to avoid weird validation errors in the unlikely (but possible) case that a cert is > 64 KiB.

adapter/ph/src/km_cert_exchange.rs
```diff
@@ -0,0 +1,396 @@
...
+const PEM_END_CERTIFICATE: &str = "-----END CERTIFICATE-----";
+
+#[derive(Debug)]
+pub enum CertExchangeError {
+    CertificateError,
+    CertificateFormatError,
+    KeyError,
+    CertificateParseError,
+    InvalidPayloadError,
+    ShortPayloadError,
```
cpacejo, line 37: "ShortPayloadError" name seems backward?  (It looks like it indicates that the payload is too long?)  Maybe "PayloadTooLongError"?

adapter/ph/src/km_cert_exchange.rs
```diff
@@ -0,0 +1,396 @@
...
+            }
+        };
+        Ok(KmCertExchange::new(cert, authority_cert))
+    }
+
+    /// Write a cert exhange payload into the supplied buffer, returns number of bytes written.
+    ///
+    /// ## Errors
+    /// - [CertExchangeError::ShortPayloadError] - the buffer is too short to hold the payload.
+    pub fn write_payload(&mut self, buf: &mut [u8]) -> Result<usize, CertExchangeError> {
```
cpacejo: `&self` shouldn't need to be mutable here, right?

cpacejo: Could use `&mut impl bytes::BufMut` instead of `&mut [u8]` – would then allow e.g. directly writing into a `Packet`.  Then no need to track `n` or return the `usize`.

(The easy translation then is `buf.len()` -> `buf.remaining_mut()` and `buf[..].copy_from_slice()` -> `buf.put()`.)

adapter/ph/src/km_cert_exchange.rs
```diff
@@ -0,0 +1,396 @@
...
+    /// - [CertExchangeError::InvalidPayloadError] - unable to parse our header from the payload.
+    /// - [CertExchangeError::CertificateParseError] - unable to parse the DER encoded certificate from the payload.
+    /// - [CertExchangeError::CertificateVerificationError] - the certificate failed signature verification (not signed by expected authority).
+    /// - [CertExchangeError::KeyMismatchError] - the public key in the certificate does not match the `expected_peer_key`.
+    /// - [CertExchangeError::CertificateFormatError] - unable to get a public key from the certificate.
+    /// - [CertExchangeError::KeyError] - OpenSSL unable to get the raw public key form.
+    pub fn process_payload(
+        &self,
+        payload: &[u8],
+        expected_peer_public_key: &Vec<u8>,
```
cpacejo: `&[u8]` instead of `&Vec<u8>` to make the interface more generic?

(`.as_slice()` gets you a `&[]` from a `Vec`.)

cpacejo (dismissed):
Nice!  Just a few suggestions but looks good as-is!

mkolehmainen:
Made tweaks. Please re-review!

cpacejo (approved):
Looks good thanks!

PR #462, opened 2024-09-26
Add cert exchange to the key manager handshake

### LINKED ISSUE #419: Key Manager to pass certs
During the noise handshake sequence the adapter should include its cert and proof of holding private key.

The node could also send a cert, but I do also think that the adapter can obtain a node cert out of band.  In that case the adapter should send something encrypted with node's public key so the node need only respond with proof that it holds the private key. 

Blocked by:
- ( #457 )

mkolehmainen:
Adds cert exchange to the key manager handshake.  When KM is done, the transport state object now includes the certificate from the peer.

- Also patched up the node and cd binaries so that they work with new code. They also needed addition configurations.
- Patched the basic and capture tests so that they pass in new command line args to main, and also create the correct certs.


tools/zpr-pki
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/usr/bin/env bash
```
cpacejo, line 1: Is this a macOS thing?  (I've only ever seen this used for Python.)

mkolehmainen: I don't know... MacOS does not have bash as /usr/bin/bash which I found out while working on my laptop in office.

adapter/cd/src/cd/config.rs
```diff
@@ -39,18 +41,29 @@ struct Dock {
 
 #[derive(Debug, Clone, Deserialize)]
 struct Adapter {
-    certificate: String, // path to PEM file
-    private_key: String, // path to PEM file
+    noise_certificate: String, // path to PEM file
```
cpacejo: `PathBuf` (or `Box<Path>`) instead of `String` maybe?

adapter/cd/src/cd/zpr.rs
```diff
@@ -136,6 +136,14 @@ impl Zpr {
         Some(*key)
     }
 
+    pub fn get_crypto_config(&self, name: &str) -> Option<CryptoConfig> {
+        let state = self.shared.state.lock().unwrap();
+        let cfg = state.configs.get(name);
+        cfg?;
+        let (conf, _) = cfg.unwrap();
+        Some(conf.get_crypto_particulars())
```
cpacejo: `let (conf, _) = cfg?;` should do the trick, no need for a separate line and `unwrap()`.

adapter/ph/src/main.rs
```diff
@@ -85,13 +91,16 @@ struct CmdLine {
...
+    #[arg(long)]
+    certificate_file: Option<String>, // noise public key signed by authority
 
     #[arg(long)]
-    certificate_file: String,
+    private_key_file: Option<String>, // noise private key
 
     #[arg(long)]
-    private_key_file: String,
+    node_public_key_file: Option<String>, // noise public key for node (only specified when starting an adapter)
```
cpacejo, line 104: You said there's a way to configure Noise not to require this, right?  (So instead we can allow the node to use whatever key material it wants, which we later check is the same as that in the cert the node gives us, which we then check has the CN we expect to connect to.)

(Not asking for a change here – just wondering if this is possible in the future.)

adapter/ph/test/capture-test.sh
```diff
@@ -4,6 +4,8 @@ set -euo pipefail
 PH_BIN=$(realpath "$(dirname $0)/../target/debug/ph")
 PH_DEBUG_BIN=$(realpath "$(dirname $0)/../../ph-debug/target/debug/ph-debug")
 
+ZPR_PKI_BIN=$(realpath "$(dirname $0)/../../../tools/zpr-pki")
```
cpacejo: Hm maybe stick this in `common_funcs.sh` since it only seems to be used there?

adapter/ph/test/common_funcs.sh
```diff
@@ -85,16 +85,22 @@ function destroy_network() {
 
 function create_ca_key_and_cert() {
   CA_NAME=$1
-  openssl genrsa -out "$CA_NAME.key"
-  openssl x509 -new -subj /CN="$CA_NAME" -key "$CA_NAME.key" -extfile /etc/ssl/openssl.cnf -extensions v3_ca -days 1 -out "$CA_NAME.crt"
+  $ZPR_PKI_BIN gencakey >"$CA_NAME.key"
```
cpacejo: Double-quotes around `"$ZPR_PKI_BIN"` are best in case we are being executed from a directory with a space in its name.

adapter/cd/src/cd/config.rs
mkolehmainen: That's a serde struct that reads a toml file.  Maybe it can map a string to a Path.  Not sure.  I'll try it.

mkolehmainen: does work!

adapter/cd/src/cd/zpr.rs
mkolehmainen: great

adapter/ph/src/main.rs
mkolehmainen: yes.  Though I'm not so sure we want to change to an open ended pattern.  We are using the pattern "IK". See this interesting note: https://www.latacora.com/blog/2018/07/18/factoring-the-noise/

But yes, we can switch to any of the noise patterns with their varying trade offs.

adapter/ph/test/common_funcs.sh
mkolehmainen: kk

adapter/ph/src/km_cert_exchange.rs
```diff
@@ -249,89 +264,71 @@ fn load_cert(path: &Path) -> Result<X509, ParseError> {
...
+        Err(e) => {
+            error!("error reading key from PEM data: {}", e);
+            return Err(ParseError::PEMFormatError);
+        }
+    };
+
+    match pk.raw_public_key() {
+        Ok(k) => {
+            let mut key = [0u8; NOISE_KEY_LEN];
+            key.copy_from_slice(&k);
```
cpacejo: Another way to write this is:
```
let key = <u8; NOISE_KEY_LEN>::try_from(&k).unwrap();
```
(I don't think there's any semantic or performance difference, just slightly fewer characters for you to type!)

Though actually this begs the question, do we know at this point that the key is of the correct length?  I.e. if the certificate has a bad key, should we return `Err` instead of panicking?

adapter/ph/src/km_noise.rs
```diff
@@ -69,14 +75,86 @@ pub struct KmNoise {
...
+            Ok(k) => k,
+            Err(e) => {
+                panic!("error generating keypair: {:?}", e);
+            }
+        };
+        skp.into()
+    }
+}
+
+impl Into<snow::Keypair> for NoiseKeypair {
```
cpacejo: I _think_ you can actually implement `impl From<NoiseKeypair> for snow::Keypair` here instead, and that doing so is preferred.  ([This](https://doc.rust-lang.org/std/convert/trait.From.html) seems to indicate doing so is possible since Rust 1.41.)

adapter/ph/src/km_noise.rs
```diff
@@ -152,8 +229,18 @@ impl KmNoise {
...
+        let mut payload_buf = BytesMut::with_capacity(MSG_BUF_SIZE);
+        payload_buf.put_slice(payload.as_bytes());
+        match self.certx.write_payload(&mut payload_buf) {
+            Ok(_) => {}
+            Err(e) => {
+                error!("noise: error writing certificate exchange payload: {:?}", e);
+                return Err(KmError::CertExchangeError);
+            }
+        };
+        let mut buf = BytesMut::zeroed(MSG_BUF_SIZE); // TODO: using with_capacity does not work when passed into hs. Why?
```
cpacejo: `zeroed()` creates a buffer of the specified length, with capacity equal (or greater); `with_capacity()` creates a buffer of length *zero*, with capacity as specified (or greater).

`write_message()` accepts a slice, not a `BufMut`.  A `BytesMut`/`BufMut`, when coerced to a slice (automatically via `Deref`), produces a slice of length equal to the _length_ of the buffer, not its capacity.

So using `with_capacity()` results in a slice of length 0 passed to `write_message()`, which `write_message()` is unable to do anything with.  (Slices cannot have their length modified -- that is the problem which `BufMut` aims to solve.)

Unfortunately I don't actually know whether there's a "clean" way to adapt a slice-writing API (like `write_message()`) into one which writes to an uninitialized `BytesMut` (i.e., avoiding any excess copying or zero-initialization).  So I think what you're doing here is both correct and also best-practice.

cpacejo (dismissed):
This all looks good!  Thanks for your patience with under-the-weather me this week!

adapter/ph/src/km_cert_exchange.rs
mkolehmainen: Yes good idea. Added a check.

adapter/ph/src/km_noise.rs
mkolehmainen: Ah ok. Good one. I thought I'd seen errors previously when I tried to add traits to structs I didn't own.

adapter/ph/src/km_noise.rs
mkolehmainen: Thanks! I left that comment in there hoping you would clarify it for me!

mkolehmainen:
Ok @cpacejo -- I've made changes based on your comments.  Also added some more targets to makefile for fun.

adapter/ph/src/km_cert_exchange.rs
```diff
@@ -249,89 +264,78 @@ fn load_cert(path: &Path) -> Result<X509, ParseError> {
...
+        Ok(k) => {
+            if k.len() != NOISE_KEY_LEN {
+                error!(
+                    "public key in cert is incorrect length (got {} bytes, expected {})",
+                    k.len(),
+                    NOISE_KEY_LEN
+                );
+                return Err(ParseError::KeyError);
+            }
+            let key = <[u8; NOISE_KEY_LEN]>::try_from(k).unwrap();
```
cpacejo, line 319: Can be simpler still! :) & avoid the explicit check and unwrap.

```
let Some(key) = <[u8; NOISE_KEY_LEN]>::try_from(k) else {
   error!(...);
   return Err(...);
}
```

cpacejo (approved):
Perfect thanks!  One nit comment only :)

adapter/ph/src/km_cert_exchange.rs
mkolehmainen: I'll make that change next time I touch the file!

PR #463, opened 2024-09-26
remove auto generated thrift files from the source

### LINKED ISSUE #458: Remove auto generated code, add makefile steps
Remove the auto generated code.  Add makefile steps to build the dependencies.

mkolehmainen:
You will now have to build the thrift files yourself.  Makefiles should do it for you.

There are still protocol buffer generated files in the visa service which came along with the code from prototype.  A separate task will address that.

The relevant workflow files now make use of a new github worker docker image that has the thrift compiler (and rust and go) baked in.  The dockerfile is stored in `tools/devenv`.

There was also an issue with the worflow files where the paths were not enclosed in quotes.  That is fixed too.


svjaffer (approved):
Thank you, Mathias!  This is great :D

PR #467, opened 2024-10-02
visa service interface code

### LINKED ISSUE #146: Node visa service interface code
Add code layer over the VS API that will allow the node to "register" with the visa service.  Registration is the HELLO and AUTH calls.  This is closely paired with the ( #466 ) story that spins up the node's Visa Support Service.

For this to actually work, we need these:
- org-zpr/zpr-core#17
- org-zpr/zpr-core#27
- org-zpr/zpr-core#30

mkolehmainen:
* New crate: 'libnode'
* Little runloop for talking to the visa service, uses channels.
* Also fires up a visa support service, which also uses channels.


visaservice/thrift/vs.thrift
mkolehmainen, line 1: I found out that thrift will generate "nice" api documentation if you use the correct style of comments.  The only changes in here are:
1- changes to comment format and
2- removed a bunch of requires since those were overly restrictive

node/src/core.rs
mkolehmainen, line 1: A lot of the new libnode code was actually already in this crate in the vs/ dir.  The code in here has just been tweaked to work with the new libnode.

libnode/src/vscli.rs
```diff
@@ -0,0 +1,256 @@
...
+use thrift::transport::{ReadHalf, WriteHalf};
+use thrift::transport::{TFramedReadTransport, TFramedWriteTransport};
+use thrift::transport::{TIoChannel, TTcpChannel};
+use tracing::debug;
+
+use crate::vsapi;
+use ph::zpr;
+use vsapi::{TVisaServiceSyncClient, VisaServiceSyncClient};
+
+// ugh!!
```
svjaffer, line 20: Out of curiosity, why ugh?  Is it just because of the ugliness of the nested <>s?  Is this a thing you're planning to clean up later?

svjaffer (approved):

libnode/src/vscli.rs
mkolehmainen: Yes, I'm just just disappointed in rust for not providing me with a less cumbersome way to describe the client interface.  It's very thorough though.

PR #470, opened 2024-10-07
Link State Part 1: Refactor link bringup

svjaffer:
- Move most of link bringup out of main (some pending)
- Initial add of LinkState data structures (mostly unused in this PR)
- Dynamic link bringup based on key management messages
- Remove now-extraneous peer IPs from node command line args

### DIFF
--- a/adapter/ph/src/assembly.rs
+++ b/adapter/ph/src/assembly.rs
@@ -5,16 +5,26 @@ use crate::config;
 use crate::counter::*;
 use crate::counters_enum::*;
 use crate::flow_control::FlowControl;
+use crate::km::ZPIPair;
+use crate::km_cert_exchange::KmCertExchange;
+use crate::km_multiplexor;
 use crate::km_multiplexor::KmState;
+use crate::km_noise;
+use crate::link_state::LinkType;
 use crate::mgmt_processor_worker;
 use crate::peer_table;
+use crate::peer_table::PeerInsertError;
 use crate::queues::*;
 use crate::tun_ctl::TunCtl;
 use crate::zpr;
+use crate::zpr::ZPI_ENCRYPTED_HEADER_FLAG;
+use crate::zpr::{LinkId, SubstrateAddr};
 
 use enum_map::EnumMap;
+use km_noise::NoiseKeypair;
 use std::default::Default;
 use std::result::Result;
+use tracing::info;
 
 #[derive(Clone, Copy, PartialEq)]
 pub enum PhMode {
@@ -70,6 +80,10 @@ pub struct Assembly<'pktbuf> {
     pub mgmt_dispatch: MgmtDispatch<'pktbuf>,
     pub adapter_manager: AdapterManager<'pktbuf>,
     pub km_state: KmState,
+
+    pub self_noise_keypair: Option<NoiseKeypair>,
+    pub peer_noise_keypair: Option<NoiseKeypair>,
+    pub certx: Option<KmCertExchange>,
 }
 
 pub struct PhFlags {
@@ -89,29 +103,98 @@ impl Default for PhFlags {
 }
 
 impl Assembly<'_> {
-    pub fn hack_get_adapter_docking_session_id(&self) -> zpr::LinkId {
-        assert!(matches!(self.ph_mode, PhMode::Adapter));
-        let peer_ids = self.peer_ids.lock().unwrap();
-        assert_eq!(peer_ids.len(), 1);
-        peer_ids[0]
+    pub fn is_node(&self) -> bool {
+        self.ph_mode == PhMode::Node
     }
 
-    pub fn hack_add_peer(
+    pub fn is_adapter(&self) -> bool {
+        self.ph_mode == PhMode::Adapter
+    }
+
+    fn add_peer(
         &'static self,
-        peer_type: peer_table::PeerType,
-        substrate_addr: zpr::SubstrateAddr,
-    ) -> Result<zpr::LinkId, peer_table::PeerInsertError> {
+        link_type: LinkType,
+        peer_addr: &SubstrateAddr,
+    ) -> Result<LinkId, PeerInsertError> {
         let entry = self.peer_table.vacant_entry()?;
 
         let worker_config = mgmt_processor_worker::Config {
             link_id: entry.key(),
         };
 
-        let peer_state = peer_table::PeerState::new(peer_type, substrate_addr, |q| {
+        let peer_state = peer_table::PeerState::new(link_type, *peer_addr, |q| {
             mgmt_processor_worker::launch(&worker_config, self, q)
         });
 
         Ok(entry.insert(peer_state))
+    }
+
+    /// Add an adapter to the peer table
+    pub fn accept_tether(
+        &'static self,
+        adapter_addr: &SubstrateAddr,
+    ) -> Result<LinkId, PeerInsertError> {
+        assert!(self.is_node());
+        info!(
+            "{}: Accepting tether from {}",
+            self.system_name, adapter_addr
+        );
+        let peer_id = self.add_peer(LinkType::NodeToAdapter, adapter_addr)?;
+        self.peer_ids.lock().unwrap().push(peer_id);
+
+        km_multiplexor::add_node_link(
+            &self,
+            peer_id,
+            ZPIPair::new(ZPI_ENCRYPTED_HEADER_FLAG | 3, 4),
+            self.self_noise_keypair.clone().unwrap(),
+            self.certx.clone().unwrap(),
+        )
+        .unwrap();
+
+        info!(
+            "{}: Successfully accepted tether from {}.  Assigned ID {}",
+            self.system_name, adapter_addr, peer_id
+        );
+
+        return Ok(peer_id);
+    }
+
+    /// Add a node to the peer table as an adapter
+    pub fn initiate_tether(
+        &'static self,
+        node_addr: &SubstrateAddr,
+    ) -> Result<LinkId, PeerInsertError> {
+        assert!(self.is_adapter());
+        info!(
+            "{}: Initiating tether towards {}",
+            self.system_name, node_addr
+        );
+        let peer_id = self.add_peer(LinkType::AdapterToNode, node_addr)?;
+        self.peer_ids.lock().unwrap().push(peer_id);
+
+        km_multiplexor::add_adapter_link(
+            &self,
+            peer_id,
+            ZPIPair::new(zpr::ZPI_ENCRYPTED_HEADER_FLAG | 5, 6),
+            self.self_noise_keypair.clone().unwrap(),
+            self.peer_noise_keypair.clone().unwrap().public,
+            self.certx.clone().unwrap(),
+        )
+        .unwrap();
+
+        info!(
+            "{}: Successfully initiated tether to {}.  Assigned ID {}",
+            self.system_name, node_addr, peer_id
+        );
+
+        return Ok(peer_id);
+    }
+
+    pub fn hack_get_adapter_docking_session_id(&self) -> zpr::LinkId {
+        assert!(matches!(self.ph_mode, PhMode::Adapter));
+        let peer_ids = self.peer_ids.lock().unwrap();
+        assert_eq!(peer_ids.len(), 1);
+        peer_ids[0]
     }
 }
 
@@ -231,6 +314,9 @@ pub mod test {
             mgmt_dispatch,
             adapter_manager,
             km_state,
+            self_noise_keypair: None,
+            peer_noise_keypair: None,
+            certx: None,
         }
     }
 }
--- a/adapter/ph/src/assembly.rs
+++ b/adapter/ph/src/assembly.rs
@@ -142,14 +142,16 @@ impl Assembly<'_> {
         let peer_id = self.add_peer(LinkType::NodeToAdapter, adapter_addr)?;
         self.peer_ids.lock().unwrap().push(peer_id);
 
-        km_multiplexor::add_node_link(
-            &self,
-            peer_id,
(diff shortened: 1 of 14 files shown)


adapter/ph/src/mgmt/dispatch.rs
```diff
@@ -19,27 +19,45 @@ use zpr_ext::zerocopy::FromBytesExt;
...
+    ingress_link_id: Option<zpr::LinkId>,
+    peer_sa: zpr::SubstrateAddr,
     mut pkt: Packet<'pktbuf>,
 ) {
     match zdp::ZdpBaseHeader::ref_from_prefix(pkt.body()) {
         Some(base_hdr) if base_hdr.packet_type == zdp::ZdpPacketType::KeyManagement => {
             pkt.advance(std::mem::size_of::<zdp::ZdpBaseHeader>());
+
+            // TODO: once we have multi-node, how do we know whether this is a link or a
+            // tether?
```
mkolehmainen, line 31: In prototype I solved this by having links use a different UDP port from tethers.

adapter/ph/src/link_state.rs
```diff
@@ -0,0 +1,53 @@
...
+#[allow(dead_code)]
+#[derive(Copy, Clone)]
+pub enum LinkType {
+    AdapterToNode,
+    NodeToNode,
+    NodeToAdapter,
+}
+
+#[allow(dead_code)]
+#[derive(Copy, Clone, PartialEq, Debug)]
```
mkolehmainen, line 14: It would be nice to have a little ascii diagram here of the state transitions at some point. Or you could even put on in the diagrams directory.

adapter/ph/src/queues.rs
```diff
@@ -270,19 +270,21 @@ impl<'pktbuf> MgmtDispatch<'pktbuf> {
...
         packet: Packet<'pktbuf>,
     ) -> Result<(), TryEnqueueError<Packet<'pktbuf>>> {
-        match self
-            .sender
-            .try_send(MgmtDispatchMessage::Packet(ingress_link_id, packet))
-        {
+        match self.sender.try_send(MgmtDispatchMessage::Packet(
+            ingress_link_id,
+            *peer_sa,
+            packet,
```
mkolehmainen: Really question for @cpacejo  -- there are a number of places we pass the packet, and it's inbound link ID and the peer_substrate_addr.  Maybe we should put the peer substrate addr into the packet metadata as soon as we create it? And add in the inbound link ID once we know it. ??

mkolehmainen: Is peer substrate addr always sending substrate addr?  Outbound packets may have need for the details going the other way.

mkolehmainen (dismissed):
The capture test is showing some sort of error with command line args that should be an easy fix.

I made comments none of which are critical.

Looks good to me!!

adapter/ph/src/mgmt/dispatch.rs
svjaffer: That's a definite possibility.  If we do that, we definitely need to write it into the spec, though.

adapter/ph/src/link_state.rs
svjaffer: I can copy the diagrams from the RFC in here

adapter/ph/src/queues.rs
cpacejo: Inbound link ID I agree!  I have a mental TODO for this, just added an item to our project.

Substrate address is only useful immediately before resolving inbound link ID, and for new links.  I agree it might be a little cleaner to pass around in the packet metadata but it also will just take up space for 99% of packets flowing through.  I'll give some thought…

adapter/ph/src/assembly.rs
```diff
@@ -89,30 +103,99 @@ impl Default for PhFlags {
...
+            self.system_name, node_addr
+        );
+        let peer_id = self.add_peer(LinkType::AdapterToNode, node_addr)?;
+        self.peer_ids.lock().unwrap().push(peer_id);
+
+        km_multiplexor::add_adapter_link(
+            &self,
+            peer_id,
+            ZPIPair::new(zpr::ZPI_ENCRYPTED_HEADER_FLAG | 5, 6),
+            self.self_noise_keypair.clone().unwrap(),
```
svjaffer: Actually, I think I need to check if these are valid before I use them here, either by checking that they exist or checking !disable_km like we do other places.  Since I have more code forthcoming, I hope it's okay if I fix this in the follow-on (unless there are other things I need to fix).

adapter/ph/src/queues.rs
```diff
@@ -256,7 +256,7 @@ impl<'pktbuf> Capture<'pktbuf> {
 }
 
 pub enum MgmtDispatchMessage<'pktbuf> {
-    Packet(zpr::LinkId, Packet<'pktbuf>), // FIXME: reserve link ID 0, store link ID in packet
+    Packet(Option<zpr::LinkId>, zpr::SubstrateAddr, Packet<'pktbuf>), // FIXME: reserve link ID 0, store link ID in packet
```
cpacejo: Let's make this two different kinds of messages: `LinkMsg(zpr::LinkId, Packet)` and `NonLinkMsg(zpr::SubstrateAddr, Packet)`.  (Substrate address is not relevant once we've resolved the link ID.)  Handling each of these messages I suspect is very different from each other.

adapter/ph/src/queues.rs
```diff
@@ -270,19 +270,21 @@ impl<'pktbuf> MgmtDispatch<'pktbuf> {
 
     pub fn try_dispatch_mgmt_packet(
```
cpacejo: And then split this function into two, one to dispatch link messages, and one to dispatch non-link messages.

adapter/ph/src/mgmt/dispatch.rs
```diff
@@ -19,27 +19,45 @@ use zpr_ext::zerocopy::FromBytesExt;
 /// This function does not block, and does not perform significant processing.
 /// It merely dispatches the management packet to the correct queue.
 pub fn dispatch_mgmt_packet<'pktbuf>(
```
cpacejo: And this function can then also split into two, one for link messages, one for non-link messages.

adapter/ph/src/link_state.rs
```diff
@@ -0,0 +1,53 @@
...
+    link_type: LinkType,
+    link_state: LinkState,
+    link_status: LinkStatus,
+    silent: bool,
+}
+
+impl LinkStateMachine {
+    pub fn new(new_link_type: LinkType) -> Self {
+        Self {
+            link_id: 0, // Invalid link ID
```
cpacejo: Maybe add a `INVALID_LINK_ID` constant to zpr.rs?

EDIT: OH WAIT!  I literally have code which does this in a stash???  Why didn't I ever put it up for review... ignore this comment & I'll take care of it with a forthcoming PR!

adapter/ph/src/assembly.rs
cpacejo: That's fine!

adapter/ph/src/assembly.rs
```diff
@@ -89,30 +103,99 @@ impl Default for PhFlags {
...
+        self.peer_ids.lock().unwrap().push(peer_id);
+
+        km_multiplexor::add_node_link(
+            &self,
+            peer_id,
+            ZPIPair::new(ZPI_ENCRYPTED_HEADER_FLAG | 3, 4),
+            self.self_noise_keypair.clone().unwrap(),
+            self.certx.clone().unwrap(),
+        )
+        .unwrap();
```
cpacejo: What are the ways in which `add_node_link()` can fail, is it always a logic error, or can it fail dynamically in response to external stimuli?  If the former, maybe instead of `.unwrap()` it would be more helpful to `.expect("logic error: blablabla")`.  If the latter, let's test for the error and map it to a `PeerInsertError` which we can then return (and don't forget to remove the peer from the peer table.

adapter/ph/src/assembly.rs
```diff
@@ -89,30 +103,99 @@ impl Default for PhFlags {
...
+
+        km_multiplexor::add_adapter_link(
+            &self,
+            peer_id,
+            ZPIPair::new(zpr::ZPI_ENCRYPTED_HEADER_FLAG | 5, 6),
+            self.self_noise_keypair.clone().unwrap(),
+            self.peer_noise_keypair.clone().unwrap().public,
+            self.certx.clone().unwrap(),
+        )
+        .unwrap();
```
cpacejo: Same here re: `expect()`.  I feel like there's an assumption here why `add_adapter_link()` ought not to fail which is immediately clear; `expect()` can be used to state that reason in the backtrace in the event that it does fail.

adapter/ph/src/fastpath.rs
```diff
@@ -498,8 +505,10 @@ pub fn substrate_ingress<'pktbuf>(
         // Not under a security assocation  which means only ZPI 0 is allowed.
         if zpi_hdr.zpi != zpr::ZPI_0 {
             info!(
```
cpacejo, line 507: So now at this point we no longer have the invariant that the packet is associated with a link (and therefore a known entity).  So our attack surface here is wide open.  It may be best to only log in the case that `ingress_link_id.is_some()`.

adapter/ph/src/fastpath.rs
```diff
@@ -440,56 +444,59 @@ pub fn substrate_ingress<'pktbuf>(
 ) {
     asm.counters[CounterType::InPacksRec].increment();
 
-    let Some(ingress_link_id) = asm.peer_table.lookup_peer(peer_sa) else {
-        drop_and_count(asm, pkt, CounterType::UnknownPeer);
-        return;
-    };
+    let ingress_link_id = asm.peer_table.lookup_peer(peer_sa);
```
cpacejo, line 447: OH!!!!!!!!! I thought I checked in code to reserve link ID 0 but… it's just sitting in my stash????  Let me get a review up for that........

adapter/ph/src/fastpath.rs
```diff
@@ -535,7 +544,9 @@ pub fn substrate_ingress<'pktbuf>(
         if !asm.flags.allow_insecure_zpi_zero {
             warn!(
```
cpacejo: Same here re: logging only when `ingress_link_id.is_some()`.

adapter/ph/src/fastpath.rs
```diff
@@ -559,6 +570,10 @@ pub fn substrate_ingress<'pktbuf>(
         return;
     }
 
+    let Some(ingress_link_id) = ingress_link_id else {
```
cpacejo: If you do split `try_dispatch_mgmt_packet()` into two (one for link messages, one for non-link messages), then you could then insert this check and the call to dispatch non-link messages up above immediately after the call to `maybe_capture()`.  So all the code between there and here then knows we're on an established link.

(Just thinking how to extend the span of that invariant / have fewer classes of packets which flow through this logic.)

adapter/ph/src/peer_table.rs
```diff
@@ -199,6 +195,7 @@ impl<'pktbuf> PeerTable<'pktbuf> {
             .get(link_id)
             .ok_or(SecurityAssocaitionStateError::NoAssociationForLink)?;
         entry.km_state.transport_sa.write(Some(sa));
+        //entry.link_state_machine.lock().unwrap().keying_done();
```
cpacejo, line 199: Is this waiting on some future code?  (& if it's not part of this patch series can you add a comment stating what it's waiting on?)

adapter/ph/src/main.rs
```diff
@@ -349,95 +345,39 @@ fn main() -> ExitCode {
...
-                .unwrap();
-            }
+            asm.certx = Some(KmCertExchange::new_from_paths(
+                &Path::new(&cert_file.as_ref().unwrap()),
+                &Path::new(&ca_file.as_ref().unwrap())
+                ).unwrap());
         }
 
-        let peer_id = asm
-            .hack_add_peer(
```
cpacejo, line 404: So good to see this mess disappear!!

adapter/ph/src/main.rs
```diff
@@ -514,20 +451,23 @@ fn main() -> ExitCode {
                 socket,
             ));
         }
+        //if asm.is_node() {
```
cpacejo, line 454: Is this just waiting on something else in the patch series?

cpacejo (commented):
Looks good!  The main thing I'd do differently is split the mgmt dispatch path into two (one for link messages, one for non-link messages) as detailed in comments -- since those two paths don't really share any common code I think it will simplify the logic and avoid us having to reason about both states in a single code path.  I think it's fine for now tho so feel free to revisit that as a later PR.

cpacejo (dismissed):
Oops meant to hit approve!

adapter/ph/src/main.rs
svjaffer: Yes, sorry, should have taken it out, but it's going in the next PR

adapter/ph/src/peer_table.rs
svjaffer: Yes, sorry, should have fully taken this stuff out.  This is coming in the next PR

adapter/ph/src/queues.rs
svjaffer: That makes sense!

adapter/ph/src/link_state.rs
svjaffer: :rofl: :rofl: :rofl:

adapter/ph/src/fastpath.rs
svjaffer: I figured it was coming eventually, since you mentioned doing it lol

cpacejo (approved):
Perfect!!!!!!!!!!!!!!

PR #476, opened 2024-10-09
Mk addr message

mkolehmainen:
I believe we need an extra message from adapter to node following the key exchange.  I have doctored up the README diagram to illustrate.

Purpose of the message is to tell the node the ZPR address that the adapter is using.  

Normally this would get sorted out during a real authentication of the agent with the visa service.  Either the adapter would be permitted to use a ZPR address in its configuration or it would be given a dynamic one -- depending on policy.

We are not doing real authentication for m2.  Instead, just using the CN in the noise cert and having the node send a connect message to the visa service that says, "adapter with CN=x is at ZPR addr y".

Not sure what message type to hijack.  I would say either HelloRequest/Response or ConfigurationRequest/Response.  The response can just be a simple "ACK".  The request just needs to contain an IP address.

I'll update the diagram here when we figure it out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Core ZPR components
 Here is the general "startup" sequence for milestone 2.
 
  * NODE - the only node
- * V/S ADAPTER - the adapter in front fo the visa service.
+ * V/S ADAPTER - the adapter in front of the visa service.
  * VISASVC - the visa service
  * ADAPTER - one of the "regular" (non-visa service) adapters.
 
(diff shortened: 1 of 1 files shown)


cpacejo:
How does this differ from Register Agent Address?

mkolehmainen:
> How does this differ from Register Agent Address?

yes, that would be a good one to use.

mkolehmainen:
Updated diagram, now using register-agent-address request/response.

cpacejo (dismissed):
This matches my understanding now -- which FWIW I'm just following what Frank has laid out in RFC 6.5 draft 2 § 5.6.  Register Agent Address is step 6 (though Frank lists is as optional there, but mandatory elsewhere, I left a Q in the doc a while back for him).

(Aside -- once we implement Hello, that will occur directly between key exchange and Register Agent Address.)

svjaffer (commented):
Technically, the diagram is also missing the hello message, which comes after the key management but before register agent address

mkolehmainen:
> Technically, the diagram is also missing the hello message, which comes after the key management but before register agent address

@svjaffer -- are we doing the hello/hello-response messages for milestone2? 

(I heard offline that YES we will have hello in there, so will update diagram!)

PR #477, opened 2024-10-09
milestone 2 configurations and visa service tweaks

### LINKED ISSUE #469: milestone 2 demo configuration
Get ready for milestone 2 we need:
- node configuration
- visa service configuration
- visa service adapter configuration
- client adapter configuration
- service adapter configuration
- a policy (maybe two of them)

mkolehmainen:
New directory `examples/milestone2` where I have put the various configuration bits i _think_ we will need for milestone 2.

A bunch of hacky changes to the visa service.  Also had to alter the compiler. The correct compiler to use is in the zpr-prototype repo in the `refimpl-m2` branch.

This is really just a stage one of getting this stuff ready.  An assoicated task is the one to create a visa service test tool.  Note that the README in milestone2 dir has steps one can use to try out the visa service without any ZPR setup at all.  Probably should turn that into some sort of handy integration test.  But I note there is a follow on task to create the visa service conformance tester which I think will pretty much do that.


mkolehmainen:
Note that this includes the old node RSA keys that we talked about getting rid of.  That will be in followup task.

svjaffer (dismissed):

PR #482, opened 2024-10-10
move link ID and stream ID into packet metadata

### LINKED ISSUE #481: let's move link ID etc into packet metadata!

cpacejo:
also fix a copy+pasta in queues

### DIFF
--- a/adapter/ph/src/queues.rs
+++ b/adapter/ph/src/queues.rs
@@ -256,7 +256,7 @@ impl<'pktbuf> Capture<'pktbuf> {
 }
 
 pub enum MgmtDispatchMessage<'pktbuf> {
-    WithLink(zpr::LinkId, Packet<'pktbuf>), // FIXME: store link ID in packet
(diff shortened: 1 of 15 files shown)


adapter/ph/src/queues.rs
```diff
@@ -256,7 +256,7 @@ impl<'pktbuf> Capture<'pktbuf> {
 }
 
 pub enum MgmtDispatchMessage<'pktbuf> {
-    WithLink(zpr::LinkId, Packet<'pktbuf>), // FIXME: store link ID in packet
+    WithLink(Packet<'pktbuf>),
```
svjaffer: Instead of removing the comment entirely, maybe replace with `// Link ID stored in packet metadata` to be clear, especially since the name is still `WithLink`

svjaffer (dismissed):

cpacejo:
The `NonZero` stuff was proving to be annoying to work with and didn't really provide any benefit.  So I reworked this code to include reverting that patch.

svjaffer (approved):

PR #488, opened 2024-10-15
rework KM startup to avoid race

### LINKED ISSUE #485: Fix KM startup jitters
I believe the "invalid state" errors from KM have to do with the delay between starting the KM state machine and it actually starting its run loop.

mkolehmainen:
This gets rid of annoying race condition between starting a KM machine and having its queues setup.

### DIFF
--- a/adapter/ph/src/km_multiplexor.rs
+++ b/adapter/ph/src/km_multiplexor.rs
@@ -20,7 +20,6 @@ const CHECK_ERROR_INTERVAL: time::Duration = time::Duration::from_millis(500);
 
 #[derive(Debug)]
 pub enum KmMsgProcessingError {
-    InvalidState,
     LinkUnconfigured,
     EnqueueFailed,
 }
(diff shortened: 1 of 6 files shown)


adapter/ph/src/km_multiplexor.rs
```diff
@@ -294,8 +293,13 @@ fn add_noise_link(
     let spawn_km_tx = asm.km_state.km_tx.clone();
     let spawn_sig_tx = asm.km_state.km_sig_tx.clone();
 
+    let (km_tx, km_rx) = mpsc::channel(16);
```
cpacejo, line 296: Let's make a `const` specifying this channel depth… in my [main cleanup PR](https://github.com/org-zpr/zpr-core/pull/490) I've added some `DEFAULT_xxx_QUEUE_SIZE` constants to `config.rs` which this could be one of.  If you commit first I can take care of doing this.

cpacejo (approved):
Fantastic!!  Excited to see stuff just "work" :)

PR #489, opened 2024-10-15
fix race condition in BufferStack

cpacejo:
`get_buffers()` suffered from a potential lost notification and hang. This fixes that by registering for notifications before dropping the mutex.


mkolehmainen (approved):
Can't say I fully understand what was broken here, but looks good!  Like the tests!

PR #491, opened 2024-10-18
remove RSA keys from node/vs interaction

### LINKED ISSUE #139: visa service to confirm authority on certificates offered by nodes
The first port of the visa service does not include the checking of the certificates from nodes.

During the `hello -> authenticate` phase of the node-visaservice connection protocol, the node presents a certificate (see `NodeAuthRequest.node_cert` in the thrift protocol file).  The visa service must ensure that this certificate is signed by the ZPRnet authority.

Also means that the visa service needs the CA cert as a config input.

### LINKED ISSUE #487: remove rsa key usage from node/visa-service interaction
The prototype visa made use of an RSA key to create an HMAC when doing an authentication with a node.  Especially with the first, bootstrap node.  For now (maybe forever) remove this.  If looks easy, just use the noise keys.  This is done when the node no longer needs rsa keys are part of its config.

This impacts:
* libnode - that is using the RSA key to do an HAMC.
* visaservice - checking HMAC, has key expectations, etc.
* milestone 2 configurations - has an RSA keypair for the node.

mkolehmainen:
The visa service was expecting RSA HMAC and key from a node during initial auth.  I have removed this and updated M2 configs.

Also reworked use of the thrift VSAPI code in rust -- new idea is to just build that once for the libnode and then everybody else use that.  Otherwise you get into a bad situation where types don't match.  This required some changes to makefiles.


libnode/src/errors.rs
mkolehmainen, line 1: just moved these into their own file.

examples/milestone2/node-config.toml
```diff
@@ -4,14 +4,12 @@ node_name = "n0"
 
 [creds]
 ca_certificate = "ca-cert.pem"
-rsa_certificate = "node-rsa-cert.pem"
-rsa_private_key = "node-rsa-key.pem"
+noise_certificate = "node-cert.pem"
+noise_private_key = "UFIB5WPx5eMEtn6XY55//vGn+HXRVLdcwQC0BtDZhkU="
```
cpacejo, line 8: I know this isn't new code so feel free to address separately – I think we should support (or even require) loading the private key from a file.  Admins may want looser permissions on config files (so that e.g. any admin can edit them) without making the private key more visible than it need be (which is only by the user the PH is executing as).  Particularly, config files are often source-controlled or copy+pasted in Slack.  I'll open a separate task to make sure we don't forget this.

cpacejo (dismissed):
This all looks good, but it was hard to find what actually changed, since it looks like there was a `cargo fmt` mixed in with this which touched many files.  It would be helpful for future reviews if you can publish pure formatting changes as their own PR.  Thanks!

examples/milestone2/node-config.toml
mkolehmainen: Yes all good ideas. Not trying to solve the configuration issue at the moment, but just illustrate all the crap you need to get a node going.

PR #500, opened 2024-10-23
basic visa service admin tool

### LINKED ISSUE #483: visa service basic admin to add new policy
Using the new admin http api into visa service allow for a new policy to be loaded.

Not sure of current (as of 10/15/2024) state of the visa service admin API.  We had decided to go with an HTTP based admin interface (vs the old gRPC one).

When I initially wrote this story I thought that the admin service was still using gRPC for the API, but turns out that the HTTP on is already there.  So this story is now to just see up installing a new policy works (and if not, make it work!)

mkolehmainen:
Initial visa service admin tool that only exercises two of the admin calls: list-policy and install-policy.  This tool can be used to install a new (compiled) policy into a running visa service.


visaservice/admin/src/main.rs
```diff
@@ -0,0 +1,199 @@
...
+use std::path::{Path, PathBuf};
+use std::time::Duration;
+
+mod apitypes;
+use apitypes::{PolicyBundle, PolicyListEntry, PolicyVersion};
+
+// Somewhat inconveniently, this must match the setting in:
+// - visaservice/mods/polio/const.go (used by the "new" visa service)
+// - zpr-prototype/pkg/snet/policy/const.go (used by the old compiler)
+const POLICY_SERIAL_VERSION: u32 = 41;
```
svjaffer, line 21: Is there any way to make this a more global value that can be read in by all 3 files?  Or is the purpose to force you to modify all 3 files if you modify 1?

Do we have a long term story for dealing with this versioning?

svjaffer (approved):

visaservice/admin/src/main.rs
mkolehmainen: We will clean this up when we create the new compiler for sure.  I'll add a story to note this so that we don't forget it.

PR #502, opened 2024-10-25
abstract the tokio tun into linux-only (for now) ZprTun

### LINKED ISSUE #494: abstract tunnel interface
In order to support platform specific implementations, abstract current tunnel interface usage and make the current version the one that compiles/runs on linux.

mkolehmainen:
Attempt to create a reasonable structure for platform specific code.  Added ZprTun which is a TUN implementation that right now is just the tokio-tun and is only available for linux.

This is a baby step to adding in TUN support for macos.

Not sure what the best practice was for platform specific code in rust, I used this for inspiration:
https://crosvm.dev/book/contributing/style_guide_platform_specific_code.html


adapter/ph/src/sys/linux/zprtun.rs
```diff
@@ -0,0 +1,61 @@
...
+use std::os::fd::AsRawFd;
+use tokio_tun::{Tun, TunBuilder};
+use zpr_ext::std::mem::slice_assume_init_mut;
+use zpr_ext::tokio_tun::*;
+
+pub struct LinuxZprTun(Tun);
+
+impl LinuxZprTun {
+    #[allow(dead_code)]
+    pub fn new(tun: Tun) -> Self {
```
cpacejo: Let's make this a `From<Tun>` impl instead of a named method – that way we can use `new()` as the canonical name to create a new TUN device.

adapter/ph/src/sys/linux/zprtun.rs
```diff
@@ -0,0 +1,61 @@
...
+        LinuxZprTun(tun)
+    }
+
+    pub fn new_mq(ifname: &str, concurrency: usize) -> Vec<Self> {
```
cpacejo: ~Let's make `ifname` an option, or remove it (later we can add a "builder"-style interface for setting various bringup options like name).  In many/most cases we just want the OS to pick a free name anyway.~

EDIT: nevermind!!  I remembered that `tokio_tun` accepts an empty name to indicate "OS chooses" and that that's what we do in main.  I think this is fine as-is!

adapter/ph/src/sys/linux/zprtun.rs
```diff
@@ -0,0 +1,61 @@
...
+
+        tok_tun_devs.into_iter().map(LinuxZprTun).collect()
+    }
+
+    pub fn try_send(&self, buf: &[u8]) -> io::Result<usize> {
+        self.0.try_send(buf)
+    }
+}
+
+impl TunExt for LinuxZprTun {
```
cpacejo: These can just be direct methods of `LinuxZprTun`; there is nothing special about `TunExt`.  It only existed so I could add these methods to the existing `Tun` type.

adapter/ph/src/sys.rs
```diff
@@ -0,0 +1,19 @@
...
+
+#[cfg(target_os = "macos")]
+mod macos;
+
+mod platform {
+    #[cfg(target_os = "linux")]
+    pub use super::linux::LinuxZprTun as ZprTun;
+
+    #[cfg(target_os = "macos")]
+    pub use super::macos::MacosZprTun as ZprTun;
```
cpacejo: Alternatively you can just name this `ZprTun` in both the OS-specific crates.  No need to rename it then.

adapter/ph/src/sys/linux/zprtun.rs
cpacejo: Let's return a `Result<>`, rather than panicking if the call to `try_build_mq()` fails.  (Caller should be in control of panicking due to I/O errors.)

adapter/ph/src/sys/linux/zprtun.rs
cpacejo, line 1: I think this would be best in its own crate, as it's fully decoupled from the rest of the PH.  (The name `ZprTun` is fine!  I'm just working slowly to break the PH into smaller decoupled+cohesive parts…)

cpacejo (dismissed):
Great thanks for doing this!  Just a few structural comments/suggestions.

cpacejo (dismissed):
Looks great thanks!!

adapter/ph/src/sys/linux/zprtun.rs
mkolehmainen: AH ok.  I'll make a crate when I add the mac version.  Would be nice if PH was just a lib too and did not contain a main.

mkolehmainen:
Ok I think I've addressed the issues brought up.

adapter/ph/Cargo.toml
```diff
@@ -32,6 +32,7 @@ zerocopy = { version = "0.8.3", features = ["derive", "std"] }
 zpr = { path = "../../zpr" }
 zpr-ext = { path = "../../zpr-ext", features = ["bytes", "tokio", "tokio-tun", "zerocopy"] }
 nix = { version = "0.29.0", features = ["ioctl", "net"] }
+thiserror = "1.0.65"
```
cpacejo, line 35: Oh, neat find!

cpacejo (approved):
Perfect!!

PR #503, opened 2024-10-25
factor out shared ZPR concepts into their own crate

cpacejo:
This avoids a circular dependency between ph and libnode.


mkolehmainen (approved):
Great!

PR #504, opened 2024-10-29
Link State 2: The State Machine

svjaffer:
This commit introduces most of the link state machine internals and hooks up the events that make it run.  Handshakes now proceed in order as expected and some hacks have been removed.

Some of the other random stuff in here is about getting Register Agent Address to work as part of the bringup process.


adapter/ph/src/assembly.rs
```diff
@@ -107,6 +106,7 @@ impl Assembly<'_> {
         self.ph_mode == PhMode::Node
     }
 
+    #[allow(dead_code)]
     pub fn is_adapter(&self) -> bool {
         self.ph_mode == PhMode::Adapter
     }
```
cpacejo: If this is unused I'd just as soon delete it!

adapter/ph/src/assembly.rs
```diff
@@ -122,74 +122,43 @@ impl Assembly<'_> {
...
-            .unwrap();
-        }
-
-        info!(
-            "{}: Successfully initiated tether to {}.  Assigned ID {}",
-            self.system_name, node_addr, peer_id
-        );
+        self.peer_table.inspect(peer_id, |peer_state| {
+            if peer_state.link_state_machine.configure(self).is_err()
+                || peer_state.link_state_machine.start(self).is_err()
```
cpacejo: Can we capture what these errors are and include them in the error message?

adapter/ph/src/assembly.rs
```diff
@@ -122,74 +122,43 @@ impl Assembly<'_> {
...
-                self.certx.clone().unwrap(),
-            )
-            .unwrap();
-        }
-
-        info!(
-            "{}: Successfully initiated tether to {}.  Assigned ID {}",
-            self.system_name, node_addr, peer_id
-        );
+        self.peer_table.inspect(peer_id, |peer_state| {
```
cpacejo: Using `.get()` here instead of `.inspect()` might make the code flow a bit more smoothly.  (Also will make explicit handling of the case that peer state got deleted somehow.)

adapter/ph/src/peer_table.rs
```diff
@@ -191,14 +193,15 @@ impl<'pktbuf> PeerTable<'pktbuf> {
...
+        asm: &'static Assembly<'pktbuf>,
         link_id: LinkId,
         sa: KmTransportSA,
     ) -> Result<(), SecurityAssocaitionStateError> {
         let entry = self
             .get(link_id)
             .ok_or(SecurityAssocaitionStateError::NoAssociationForLink)?;
         entry.km_state.transport_sa.write(Some(sa));
-        //entry.link_state_machine.lock().unwrap().keying_done();
+        let _ = entry.link_state_machine.keying_done(asm);
```
cpacejo: Do we know here this method can never return an error?

adapter/ph/src/mgmt/handlers.rs
```diff
@@ -80,22 +83,142 @@ pub async fn handle_hello_request<'pktbuf>(
...
-        asm,
-        ingress_link_id,
-        zdp::ZdpPacketType::HelloResponse,
-        seq_num,
-        rsp_pkt,
-    )
-    .await;
+    if let Some(Ok(_)) = asm.peer_table.inspect(ingress_link_id, |peer_state| {
+        peer_state.link_state_machine.process_hello_request()
+    }) {
```
cpacejo: Alternative less nesty way to write this:

```
let Some(Ok(_)) = inspect(..) else { return Err(...) };
let mut rsp_pkt = ...
...
Ok(())
```

cpacejo (commented):
Reviewed all except the changes to `link_state` so far!

adapter/ph/src/link_state.rs
```diff
@@ -1,29 +1,142 @@
...
     Keying,
     Helloing,
     Closing,
     Active,
-    Listening,
     RegisterAA,
     Error,
 }
 
+pub enum LinkStateError {
```
cpacejo, line 128: For errors ideally they `#[derive(Debug)]` and also `impl Display` and `impl std::error::Error`.  Mathias though found this crate `thiserror` and added a dependency on it, which automates the process of implementing the latter two traits: https://docs.rs/thiserror/latest/thiserror/

adapter/ph/src/link_state.rs
```diff
@@ -33,21 +146,274 @@ pub enum LinkStatus {
...
+    fn get_state(&self) -> LinkState {
+        self.locked_fsm.lock().unwrap().state
+    }
+
+    fn set_state(&self, state: LinkState) {
+        self.locked_fsm.lock().unwrap().state = state
+    }
+
+    #[allow(dead_code)]
+    pub fn set_silent(&mut self) {
```
cpacejo: No need for `mut` here

adapter/ph/src/peer_table.rs
svjaffer: Oh, I just missed this spot when adding the result handling.  I'll fix that!

adapter/ph/src/assembly.rs
svjaffer: Done

adapter/ph/src/link_state.rs
```diff
@@ -33,21 +146,274 @@ pub enum LinkStatus {
...
+            "{}: Configured link {}.  State: {:?}, status: {:?}",
+            asm.system_name, self.id, locked_fsm.state, locked_fsm.status
+        );
+        Ok(())
+    }
+
+    /// Start an inactive link/tether
+    /// Transitions from Inactive -> Keying
+    pub fn start<'pktbuf>(&self, asm: &'static Assembly<'pktbuf>) -> Result<(), LinkStateError> {
+        assert!(self.id != 0);
```
cpacejo: 0 -> `zpr::LINK_ID_UNKNOWN`

adapter/ph/src/mgmt/handlers.rs
svjaffer: Done

adapter/ph/src/link_state.rs
```diff
@@ -33,21 +146,274 @@ pub enum LinkStatus {
...
+        if self.get_state() != LinkState::Inactive {
+            return Err(LinkStateError::UnexpectedTransition);
+        }
+
+        self.set_state(LinkState::Keying);
```
cpacejo: Here and elsewhere we do a lot of taking and dropping the lock, making this method not have a definite linearization point (and opens the potential for race conditions under concurrent access).  If the bodies of these methods were all implemented on `LinkStateMachine` directly, and then these methods just took the lock and forwarded to `LinkStateMachine`, they'd all have a linearization point and avoid race conditions.

adapter/ph/src/mgmt/requests.rs
```diff
@@ -64,7 +65,14 @@ pub async fn send_hello_request<'a, 'pktbuf>(
...
             asm.buffer_stack.put_buffer(hello_res.destroy());
-            Ok(())
+
+            if let Some(Ok(_)) = asm.peer_table.inspect(link_id, |peer_state| {
+                peer_state.link_state_machine.process_hello_response(asm)
+            }) {
+                Ok(())
+            } else {
+                Err(())
+            }
```
cpacejo, line 76: Can you add a TODO here that this is temporary until we break apart the async sync request stuff?  (Want to make sure we don't forget, as in the `async` design as-is, calling directly back specifically to the link state machine breaks the return-to-caller model.)

adapter/ph/src/link_state.rs
```diff
@@ -33,21 +146,274 @@ pub enum LinkStatus {
...
+        if self.get_state() != LinkState::Keying {
+            return Err(LinkStateError::UnexpectedTransition);
+        }
+
+        self.set_state(LinkState::Helloing);
+        self.send_hello(asm);
+        Ok(())
+    }
+
+    fn send_hello<'pktbuf>(&self, asm: &'static Assembly<'pktbuf>) {
```
cpacejo: Maybe name `maybe_send_hello()` since this only actually does that conditionally?

adapter/ph/src/link_state.rs
```diff
@@ -33,21 +146,274 @@ pub enum LinkStatus {
...
+                    asm, self.id,
+                ));
+            }
+            (LinkType::NodeToNode, LinkState::Helloing) => {
+                self.set_state(LinkState::Active);
+            }
+            (LinkType::NodeToAdapter, _) => {
+                // Nodes should not be receiving these messages from adapters
+                warn!("{}: Discarded unsolicited Hello Response", asm.system_name);
+            }
```
cpacejo: Should we return an error here like above?

adapter/ph/src/link_state.rs
```diff
@@ -33,21 +146,274 @@ pub enum LinkStatus {
...
+    pub fn process_register_agent_address_request(
+        &self,
+        asm: &Assembly,
+        addr: IpAddress,
+    ) -> Result<(), LinkStateError> {
+        match (self.link_type, self.get_state()) {
+            (LinkType::NodeToAdapter, LinkState::RegisterAA) => {
+                self.locked_fsm.lock().unwrap().agent_addresses.push(addr);
+                self.set_state(LinkState::Active);
+                self.run_active(asm)
```
cpacejo: Just marking here an example of a sequence that I think ought all to be performed under a single lock, so the transition is atomic.

adapter/ph/src/link_state.rs
```diff
@@ -33,21 +146,274 @@ pub enum LinkStatus {
...
+        let mut locked_fsm = self.locked_fsm.lock().unwrap();
+        locked_fsm.state = LinkState::Initial;
+        locked_fsm.silent = false;
+        // TODO: Send terminate
+    }
+
+    #[allow(dead_code)]
+    /// Initiate the shutdown of the link
+    pub fn shutdown(&mut self) {
+        self.set_state(LinkState::Closing);
```
cpacejo: Will stuff be happening in here in a later commit?

Also -- I think here we should bring the TUN carrier down?

adapter/ph/src/link_state.rs
```diff
@@ -33,21 +146,274 @@ pub enum LinkStatus {
...
+    fn set_state(&self, state: LinkState) {
+        self.locked_fsm.lock().unwrap().state = state
+    }
+
+    #[allow(dead_code)]
+    pub fn set_silent(&mut self) {
+        self.locked_fsm.lock().unwrap().silent = true
+    }
+
+    pub fn reset(&self) {
```
cpacejo: Is `reset()` a state transition or is it "special" (a reset of the data structure)?  If it's a state transition, how does it differ from `shutdown()`?

Either case here I think we should bring the TUN carrier down?

cpacejo (changes requested):
Nice I like it!!

Only major change I think we should make though is making the state transitions atomic.

adapter/ph/src/link_state.rs
svjaffer: Done

adapter/ph/src/link_state.rs
svjaffer: Done

adapter/ph/src/link_state.rs
svjaffer: Done

adapter/ph/src/link_state.rs
svjaffer: Sure, done

adapter/ph/src/link_state.rs
svjaffer: Yes, I'm missing some TODOs here.  Link shutdown isn't called currently, but I'll be fleshing this operation out in a subsequent commit.

adapter/ph/src/link_state.rs
svjaffer: Reset and close have different behavior in how they bring the link down, but yes, the tun carrier should come down.

adapter/ph/src/link_state.rs
svjaffer: Yes, fixed

adapter/ph/src/mgmt/requests.rs
svjaffer: Done

adapter/ph/src/link_state.rs
svjaffer: Nice!

cpacejo (approved):
Looks good go go go!

PR #507, opened 2024-10-31
ability to build on mac, run an adapter

### LINKED ISSUE #495: create mac implementation of tun interface
Create a mac implementation of the tun interface.  Has to work with tokio too.

I would look at the rust WG code for inspiration.

mkolehmainen:
Permits compiling on a mac, and you can bring up an adapter and successfully connect to a node on linux.

There is still some work to do around the TUN interface.  This simply brings up an interface and sets an address (and mtu).  But we still need to add the correct routing.  

I'll make a story for that and for the "set MTU" thing for the UdpSocket stuff in zpr-ext.


adapter/ph/Cargo.toml
```diff
@@ -23,16 +23,24 @@ openssl-sys = { version = "0.9.102" }
...
 thiserror = "1.0.65"
+zpr-ext = { path = "../../zpr-ext", features = ["bytes", "tokio", "zerocopy"] }
+
+[target.'cfg(target_os = "linux")'.dependencies]
+tokio-tun = { version = "0.11.5" }
+
+[target.'cfg(target_os = "macos")'.dependencies]
+# Locked verision of rust-tun as the maintainers may be suspect.
+# This is used for its ability to manage a TUN device on mac.
+tun = { version = "=0.7.1", features = ["async"] }
```
cpacejo, line 44: 0.6.1 is the last "unadulterated" version.

Aside -- because we'll be open-source (as will our commit history), I'd reword the comment to be in neutral terms.  E.g. "Locked version, because rust-tun 0.7 was substantially reworked by a new maintainer, removing several features we depend on."

zpr-ext/src/tokio.rs
```diff
@@ -2,12 +2,17 @@ pub mod net {
...
     use std::io::{self, IoSlice, IoSliceMut};
-    use std::os::fd::{AsFd, AsRawFd};
+    use std::os::fd::AsFd;
     use tokio::io::Interest;
     use tokio::net::{UdpSocket, UnixStream};
 
+    #[cfg(target_os = "linux")]
+    use nix::sys::socket;
+
+    #[cfg(target_os = "linux")]
```
cpacejo: Both `linux` and `android` here to match the use sites below.

(For context I've been specifying both of those to match what I've seen in some other libraries.  Since Android essential "is" Linux.)

zpr-ext/src/std.rs
```diff
@@ -255,6 +255,7 @@ pub mod os {
         }
     }
 
+    #[cfg(unix)]
```
cpacejo: `cfg(any(doc, unix))` ensures this shows up in documentation also.  (Which I believe gets built without any "target" options set.)

zpr-ext/src/std.rs
```diff
@@ -265,7 +266,12 @@ pub mod os {
             use std::ptr;
             use std::vec::Vec;
 
-            #[cfg(any(doc, target_os = "android", target_os = "linux"))]
+            #[cfg(any(target_os = "android", target_os = "linux"))]
```
cpacejo: And then also here add `doc` (but not to the "macos" variant).

adapter/ph/src/sys/macos/zprtun.rs
```diff
@@ -1 +1,80 @@
...
+        // SAFETY: we've now initialized this much of the slice
+        unsafe {
+            buf.advance_mut(size);
+        }
+        Ok(size)
+    }
+
+    pub fn try_send(&self, _buf: &[u8]) -> std::io::Result<usize> {
+        let idev = &*(self.0);
+        idev.send(_buf)
```
cpacejo: `buf` instead of `_buf` since it's used now :)

cpacejo (changes requested):
Looks good!  Just a couple minor suggestions to change

adapter/ph/Cargo.toml
mkolehmainen: Noted.  Altered comment, but keeping at 0.7.1 since that one works for us for now.

adapter/ph/src/sys/macos/zprtun.rs
mkolehmainen: oh good catch!

adapter/ph/Cargo.toml
cpacejo: I'm OK to check this in with 0.7.1 for just macOS but I do want to unify across OSes and we can't do that with 0.7.1.  0.6.1 is the last version that supports multiqueue and vectored I/O.

cpacejo (approved):
Looks good thanks!

PR #512, opened 2024-10-31
rework the arguments and parsing for ph/main

mkolehmainen:
Redo the arg structure a bit. Add more help.  Fix up the tests.

### DIFF
--- a/adapter/ph/src/main.rs
+++ b/adapter/ph/src/main.rs
@@ -1,18 +1,18 @@
 #![cfg_attr(feature = "ci", deny(warnings))]
 
-use clap::Parser;
+use clap::{Parser, Subcommand};
 use km_cert_exchange::KmCertExchange;
 use std::default::Default;
 use std::fs;
 use std::io::ErrorKind;
 use std::net::{IpAddr, SocketAddr};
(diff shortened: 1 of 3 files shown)


adapter/ph/src/main.rs
```diff
@@ -69,71 +69,138 @@ use queues::*;
...
-    #[arg(long)]
-    private_key_file: Option<String>, // noise private key
+    /// Enable debug logging
+    #[arg(long, short, default_value_t = false)]
+    debug: bool,
 
-    #[arg(long)]
-    node_public_key_file: Option<String>, // noise public key for node (only specified when starting an adapter)
+    #[command(subcommand)]
+    command: Option<Command>,
```
cpacejo, line 116: What would happen if you made this a plain `Command` (not an `Option`)?  (Asking b/c we reject the `None` case below.)

cpacejo (approved):
Perfect I like this approach!

adapter/ph/src/main.rs
mkolehmainen: Will try this on my next pass, which will add in config file.

PR #520, opened 2024-11-05
Fix integration tests to dump stats again

### LINKED ISSUE #519: Basic-test doesn't dump stats

svjaffer:
Resolves #519

### DIFF
--- a/adapter/ph/test/basic-test.sh
+++ b/adapter/ph/test/basic-test.sh
@@ -35,7 +35,6 @@ function counters() {
 # Set up automatic cleanup
 #
 
-CHILDREN=()
 
 trap cleanup EXIT
 
@@ -72,7 +71,6 @@ sudo -E ip netns exec zpr-node sudo -E -u "$ZPR_USER" "$PH_BIN" \
      --certificate-file node.crt \
      --private-key-file node.key \
      --tun-if tun0 node 2>&1 |tee node.log &
-CHILDREN=(${CHILDREN[@]} "$!")
 
 sleep 2
 
@@ -88,7 +86,6 @@ sudo -E ip netns exec zpr-a sudo -E -u "$ZPR_USER" "$PH_BIN" \
   --node-addr "$NODE_SUBSTRATE_ADDR_A":12345 \
   --node-public-key-file node.pubkey \
   --agent-addr "$A_ZPR_ADDR" 2>&1 |tee adapter1.log &
-CHILDREN=(${CHILDREN[@]} "$!")
 
 
 sudo -E ip netns exec zpr-b sudo -E -u "$ZPR_USER" "$PH_BIN" \
@@ -103,7 +100,6 @@ sudo -E ip netns exec zpr-b sudo -E -u "$ZPR_USER" "$PH_BIN" \
   --node-addr "$NODE_SUBSTRATE_ADDR_B":12345 \
   --node-public-key-file node.pubkey \
   --agent-addr "$B_ZPR_ADDR" 2>&1 |tee adapter2.log &
-CHILDREN=(${CHILDREN[@]} "$!")
 
 
 #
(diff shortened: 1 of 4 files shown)


adapter/ph/test/basic-test.sh
```diff
@@ -72,7 +71,6 @@ sudo -E ip netns exec zpr-node sudo -E -u "$ZPR_USER" "$PH_BIN" \
      --certificate-file node.crt \
      --private-key-file node.key \
      --tun-if tun0 node 2>&1 |tee node.log &
-CHILDREN=(${CHILDREN[@]} "$!")
```
svjaffer, line 75: So, the basic problem was that this was getting the PID for an ephemeral process that was going away almost immediately, so issuing signals to it wasn't doing anything.

adapter/ph/test/basic-test.sh
```diff
@@ -115,25 +111,23 @@ wait_for 5 check_carrier zpr-a tun0 || { echo "FAILURE"; exit 1; }
 wait_for 5 check_carrier zpr-b tun0 || { echo "FAILURE"; exit 1; }
 wait_for 5 check_carrier zpr-node tun0 || { echo "FAILURE"; exit 1; }
 echo "Carrier has arrived."
-
-stty sane || true
-
-
-echo "PAUSING FOR KEY MANAGEMENT EXCHANGE ..."
-countdown 10
```
svjaffer, line 123: Just got rid of this because it wasn't necessary anymore.  Unrelated to the main purpose of the PR

adapter/ph/test/basic-test.sh
```diff
@@ -155,8 +149,15 @@ done
...
 #
 
-sudo kill -SIGUSR1 "${CHILDREN[@]}" 2> /dev/null || true
-sleep 1  # FIXME: let's do something better here
+for ph in $(pgrep ph)
+do
+	echo
+	echo "Terminating $ph"
+	sleep 1
+	sudo kill -SIGTERM "$ph"
```
svjaffer: I switched it to SIGTERM since SIGUSR1 *only* prints stats and doesn't terminate the process and since this is in the cleanup block, I figured we should be actually tearing down the processes :)

adapter/ph/test/basic-test.sh
cpacejo: OH BECAUSE OF THE TEE!!!!!!!!!!!!!!!!!!!!!!!!

adapter/ph/test/basic-test.sh
cpacejo: NICE!!!!  Was gonna ask about that haha

adapter/ph/test/basic-test.sh
cpacejo: It used to be SIGTERM, I think it got changed to SIGUSR1 while we were debugging this issue ha!

adapter/ph/test/basic-test.sh
```diff
@@ -155,8 +149,17 @@ done
 # Cleanup
 #
 
-sudo kill -SIGUSR1 "${CHILDREN[@]}" 2> /dev/null || true
-sleep 1  # FIXME: let's do something better here
+rgid=$(($(ps -o rgid= -p "$$")))
+echo $rgid
```
cpacejo: `echo $rgid` leftover from debugging?

cpacejo (dismissed):
Nice work!!!!!!!!!

adapter/ph/test/basic-test.sh
svjaffer: I needed to see if it was actually different on the different github runners and it wasn't.  Basically, the first time I pushed it up, basic-test and capture-test got to this point at the same time and sniped each-others' processes and got stuck and had to be force-killed.  So I'm trying to find a way to limit it to processes in this process tree

svjaffer: Well, I tried about a dozen more elegant ways to avoid that issue, but none of them worked so in the end I brute forced it with regex-parsing pstree.  Eh, it works.

adapter/ph/test/common_funcs.sh
```diff
@@ -134,6 +134,13 @@ function countdown() {
     echo
 }
 
+# Get all descendant PIDs whose name matches a specific list
+function get_descendants() {
+    exenames="(ph|node|adapter)"
```
svjaffer, line 139: Only ph exists currently, but I wanted to make it very obvious how to modify this to add another process name to the list.  If someone wanted to be very fancy, they could make this function take a list of names and format them, removing the need for anyone to muck with regex strings at all.  But that's more effort than I'm willing to throw at this for now.

cpacejo: Ya if we find a better way we can fix it later!

cpacejo (approved):

PR #521, opened 2024-11-05
boxed packet buffers

cpacejo:
In preparation for removing `'pktbuf` from `Assembly`, this commit makes `buffer_stack` own the buffers it manages.  It owns them as `Box`es, so there is no performance hit (i.e. no reference counting).  The difference though is that this prevents us from allocating a single contiguous slab of memory and divvying it up in order to avoid malloc header overhead.  However it's possible we can leverage the [`yoke`](https://docs.rs/yoke/latest/yoke/) crate to give us this capability in the future.

`Packet`s now become parameterized over the backing buffer type, rather than the lifetime of the buffer reference.  Backing buffers must meet various requirements which are captured by the `PacketBuffer` trait.

Since almost all of our code relies on `Packets` being sourced from the buffer stack and being of a specified size, rather than write `Packet<Buffer<{ config::PACKET_BUFFER_SIZE }>>` everywhere, I've also introduced the `BufferPacket` alias, which is used through most of the packet processing path.

Therefore most of the changes here are either:
* replacing `Packet<'pktbuf>` with `Packet<impl PacketBuffer>` in buffer-agnostic code
* replacing `Packet<'pktbuf>` with `BufferPacket` in code which is part of the packet pipeline
* deleting `<'pktbuf>` arguments in many places

The only real changes are in `buffer_stack` and the top of `packet`.

### DIFF
--- a/adapter/ph/src/adapter_manager_worker.rs
+++ b/adapter/ph/src/adapter_manager_worker.rs
@@ -3,7 +3,7 @@ use crate::assembly::{Assembly, PhMode};
 use crate::counters::CounterType;
 use crate::fastpath;
 use crate::mgmt::requests;
-use crate::packet::Packet;
+use crate::packet::BufferPacket;
 use crate::queues::AdapterManagerMessage;
 use std::future::Future;
 use tokio::sync::mpsc;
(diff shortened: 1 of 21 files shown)


mkolehmainen (approved):
Great to remove all those lifetimes.  Looks like that was a painful refactoring!  

Looks good to me!

PR #522, opened 2024-11-05
add ph ability to load a config file

### LINKED ISSUE #516: add config-file option to PH
Add config file option to the PH so that all args can be loaded from a configuration file (or be overridden on command line).

mkolehmainen:
- Reworked the command line args again.
- Added ability to use a config file which can replace or work with command line args.
- Updated the test scripts.
- Added examples of the config file format.


adapter/ph/src/main_args.rs
```diff
@@ -0,0 +1,971 @@
...
+        }
+        if let Some(tun_if) = &config.tun_if {
+            self.tun_if = Some(tun_if.clone());
+        }
+        if let Some(debug) = config.debug {
+            self.debug = debug;
+        }
+    }
+
+    fn merge_adapter(&mut self, config: &AdapterConfigSection) {
```
svjaffer: Merge as a name is a bit misleading here.  This function is just loading config from the file and then it gets overwritten by command line stuff at the call site.

adapter/ph/src/main_args.rs
```diff
@@ -0,0 +1,971 @@
...
+// This describes the node configuration file TOML format.
+#[derive(Deserialize, Debug, Clone)]
+struct NodeConfig {
+    global: GlobalConfigSection,
+}
+
+// Global section is shared by nodes and adapters.
+#[derive(Deserialize, Debug, Clone)]
+struct GlobalConfigSection {
+    name: Option<String>,
```
svjaffer, line 360: Is it a problem to allow Command to be configured in the file?  I'm imagining a scenario where the only command line arg passed in is the config file, in which case it would have to be able to specify what PhMode the ph is operating in.

adapter/ph/src/main.rs
```diff
@@ -70,139 +69,18 @@ use queues::*;
...
-                tun_if: control.tun_if,
-                debug: control.debug,
-                node_addr: None,
-                agent_addr: None,
-                node_public_key_file: None,
-            };
-        }
-        None => {
-            println!("command required: either 'adapter' or 'node'");
+    let (ph_mode, config) = match main_args::argparse(None) {
```
svjaffer, line 77: I love moving all this stuff out of main.  Good change!

adapter/ph/src/main_args.rs
```diff
@@ -0,0 +1,971 @@
...
+            };
+            config = Config::default();
+            config.name = "adapter".to_string();
+            // fold in anything from the config file:
+            if let Some(config_file) = config_file {
+                config.merge_globals(&config_file.global);
+                config.merge_adapter(&config_file.adapter);
+            }
+            // fold in anything from the command line:
+            config.merge_common(&common);
```
svjaffer: I think trying to keep this all in the top level match makes this code awkward.  I think it's okay to ask at multiple points if this is a node or adapter and pass that info around.  Not sure how well I can communicate this via text, but I think this would make more sense to me as a top level control flow here:

1. Parse args into their data structure
2. Grab config path arg for the PH mode if it exists, validate the file path, yadda yadda.  If that all checks out...
    2a. Parse config file
    2b. Apply config file -> in here we can branch on adapter/node and call, for instance, merge_adapter
3. Apply command line args, blindly overwriting whatever did or did not get loaded from the file and again, we can make another Node/Adapter branch at this point

I know branch repetition feels icky, but I think it would flow better and have less code repetition.

adapter/ph/src/main_args.rs
svjaffer: Oh, as a subcommand, I'm guessing it's a required arg?

adapter/ph/src/main_args.rs
mkolehmainen: I was interpreting it as "merge AdapterConfigSection into self".  But I'll think of a better name.

adapter/ph/src/main_args.rs
mkolehmainen: Yes, I think it's better if you explicitly write what mode you want on command line.  So, the typical case is just:

`ph adapter -c myadapter.toml` or `ph node -c mynode.toml`

adapter/ph/src/main_args.rs
mkolehmainen: I reworked this with a couple of new's.

mkolehmainen:
Once this is merged I'll convert the example configs to the new format.

svjaffer (approved):

PR #529, opened 2024-11-07
newconfigs for milestone 2

### LINKED ISSUE #525: rework the m2 example configs to use new format
Redo the example configs using the new PH config file format.

Also need to add in missing items.  Off the top of my head:
- listen addr (really port) for the node
- claims (pretty sure node needs these to talk to visaservice, though maybe can be synthesized from addr?)

Also while I'm here let's either load the files before returning the config or fix the paths so that you can use relative paths in the config file.

mkolehmainen:
* Rewrite the m2 config files in the new format, and reorganize them into their own directories.
* Add ability to resolve file paths properly to the config parser.
* Add some additional checks and output tweaks to main.


adapter/ph/src/main.rs
```diff
@@ -186,7 +186,16 @@ fn main() -> ExitCode {
         })
         .unwrap();
 
-    let control_socket = Arc::new(UnixListener::bind(config.control_path).unwrap());
+    // Ensure directory for control socket exists.
+    if let Some(parent) = Path::new(&config.control_path).parent() {
+        fs::create_dir_all(parent).expect("failed to create control socket directory");
+    }
+
+    fs::create_dir_all(Path::new(&config.control_path).parent().unwrap()).unwrap();
```
cpacejo: I think PH creating directories is surprising behavior (in the face of typos) and may have security implications (e.g. directory ownership).  If the directory doesn't exist I think we should just error out.

adapter/ph/src/main_args.rs
```diff
@@ -272,85 +285,163 @@ impl Config {
...
-            self.control_path = PathBuf::from(control_path);
+            let cp = PathBuf::from(control_path);
+            if cp.is_relative() {
+                // Do not use cacnonicalize here since that will fail if the path does not exist.
+                self.control_path = path::absolute(cp).or_else(|e| {
+                    Err(ArgsError::PathError(format!(
+                        "path error for control_path: {:?}",
+                        e
+                    )))
+                })?;
```
cpacejo: (And then we won't need to fail in this case)

cpacejo (changes requested):
This all looks good though I have a reservation about automatically creating parent directories.

adapter/ph/src/main.rs
mkolehmainen: sure that's fine will do.

cpacejo (dismissed):
Looks good thank you!!

PR #532, opened 2024-11-11
Visa adapter prep

cpacejo:
* rename Dock Forwarding Table to Peer Forwarding Table, since it will be instantiated once for each peer (docking sessions and links alike)

* no need to distinguish next-hop types -- without the dichotomy of DFT + FT, there's no need to hop between the two, thus no need for two next-hop types

* represent internal dock in peer table -- this gives up a Peer Forwarding Table for the internal dock, but entails some dummy peer state stuff.  I tried representing it separately but it ended up being way messier.

* enable node forwarding based on the Peer Forwarding Table.  Since we already generate hardcoded PEPs when requesting a visa, this works already.


adapter/ph/src/assembly.rs
```diff
@@ -170,12 +189,9 @@ impl Assembly {
         return Ok(peer_id);
     }
 
-    pub fn hack_get_adapter_docking_session_id(&self) -> zpr::LinkId {
```
svjaffer, line 173: Nice!

zpr/src/lib.rs
```diff
@@ -30,7 +30,10 @@ pub type LinkId = u32;
 pub const LINK_ID_UNKNOWN: LinkId = 0;
 
 /// Link ID used to refer to a node or adapter's local agent.
-pub const AGENT_LINK_ID: LinkId = LinkId::MAX;
+pub const AGENT_LINK_ID: LinkId = 1;
```
svjaffer: I think this is fine for now, but in the future, couldn't an adapter have multiple agents associated?

adapter/ph/src/link_state.rs
```diff
@@ -154,22 +154,20 @@ pub enum LinkStateError {
...
-#[allow(dead_code)]
 #[derive(Copy, Clone, PartialEq)]
 pub enum LinkType {
+    Internal,
     AdapterToNode,
     NodeToNode, // Currently unsupported
     NodeToAdapter,
 }
 
-#[allow(dead_code)]
```
svjaffer, line 165: Didn't the compiler complain at you about this?

cpacejo: No surprisingly!  It's used somewhere I forget where…

zpr/src/lib.rs
cpacejo: `AGENT` is maybe the wrong word here: on a node, this refers to the local adapter (and its associated TUN device), on the adapter, this refers to the TUN device.  I don't know yet what it would mean to have multiple local adapters/TUN devices.  On a node these would correspond to multiple internal links I guess which is fine.  On an adapter they'd have to each correspond to a separate docking session.  I think if/when we support multiple it would be appropriate to generalize this.

svjaffer (dismissed):

svjaffer (approved):

PR #535, opened 2024-11-13
no packet info on mac tun

### LINKED ISSUE #513: Mac - get TUN socket routing working
We need to do more setup for a mac TUN interface.
- routing, gateway, etc.
- maybe also need to set things up with scutil.

mkolehmainen:
No packet info on mac tun so moved the tun_pi stuff into sys and added a NOP version for mac.

### DIFF
--- a/adapter/ph/src/agent_output_worker.rs
+++ b/adapter/ph/src/agent_output_worker.rs
@@ -4,8 +4,9 @@ use crate::counters::*;
 use crate::fastpath;
 use crate::net_defs;
 use crate::packet::Packet;
+use crate::sys::TunPi;
 use crate::sys::ZprTun;
-use crate::zprtun::tun_pi;
+use crate::zprtun;
 use std::sync::Arc;
 
 #[derive(Copy, Clone)]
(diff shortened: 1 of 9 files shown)


adapter/ph/src/agent_output_worker.rs
```diff
@@ -33,15 +34,23 @@ pub async fn launch(config: Config, asm: Arc<Assembly>, tun: Arc<ZprTun>) {
...
+                    if pi.strip || !is_ip(pi) {
+                        // packet was too large or non-IP; drop
+                        asm.counters[CounterType::OutPacksDrop].increment();
+                        // reuse `buf`
+                        buf = pkt.destroy();
+                        continue;
+                    }
+                } else {
+                    // No packet info, permit IP and IPv6 only (for now?)
+                    if pkt.body()[0] >> 4 != 4 && pkt.body()[0] >> 4 != 6 {
```
cpacejo, line 49: The IP checks are important but the main reason I added the PI checks was to check for truncated packets (i.e. they were larger than our receive buffer).  Do you know whether macOS has a means of signaling this?

If not it's probably sufficient to rely on the classifier to catch truncated packets later in the pipeline.

cpacejo (approved):
Nice!

adapter/ph/src/agent_output_worker.rs
mkolehmainen: Hmm.  Looking through the rust-tun code more it reads like there is PI on the mac, but my empirical testing shows that it is not returned up to the caller.  I will update this comment on next PR.

PR #536, opened 2024-11-13
"Special" peers

cpacejo:
Add a notion of "special" peers to the peer table.  The only type of "special" peer right now is the visa service adapter.  There can be at most one of any "special" peer of each type registered.

Also while doing this I noticed that it was possible, both for special peer lookup and substrate lookup, for the lookup to succeed, but a subsequent `get()` of the returned link ID to fail, if memory ordering doesn't behave.  (Since the writes into these "reverse" lookup tables might become visible before writes into the main peer table.)  So also, I added some release/acquire fences to ensure that (so long as the peer is not removed) performing a reverse lookup immediately after an insert will always ensure that the inserted peer is visible.

### DIFF
--- a/adapter/ph/src/peer_table.rs
+++ b/adapter/ph/src/peer_table.rs
@@ -8,6 +8,8 @@ use crate::sync_req;
 use bytes::Bytes;
 use cslab::{RcuCslab, RcuCslabReader};
 use dashmap::DashMap;
+use enum_map::{Enum, EnumMap};
+use enumset::{EnumSet, EnumSetType};
 use std::future::Future;
 use std::sync::atomic::{self, Ordering};
 use std::sync::Mutex;
@@ -20,8 +22,16 @@ use zpr::{LinkId, SubstrateAddr};
 
 const PEER_TABLE_SIZE: usize = 1024;
 
+/// Some peers are "special", e.g. the visa service adapter attached to the initial node.
+/// These names let us identify them.
+#[derive(Debug, Enum, EnumSetType)]
+pub enum SpecialPeerName {
+    VisaServiceAdapter,
+}
+
 pub struct PeerState {
     pub substrate_addr: SubstrateAddr,
+    pub special_names: EnumSet<SpecialPeerName>,
     pub link_state_machine: LinkStateWrapper,
     pub sync_req_state: sync_req::SyncReqState,
     pub pft: PeerForwardingTable,
@@ -76,6 +86,7 @@ impl PeerState {
 
         Self {
             substrate_addr,
+            special_names: EnumSet::<SpecialPeerName>::empty(),
             link_state_machine: LinkStateWrapper::new(link_id, link_type),
             pft: PeerForwardingTable::new(),
             sync_req_state: sync_req::SyncReqState::new(),
@@ -83,6 +94,10 @@ impl PeerState {
             mgmt_processor_worker,
             km_state: PeerKmState::new(),
         }
+    }
+
+    pub fn add_special_name(&mut self, name: SpecialPeerName) {
+        self.special_names |= name;
     }
 
     /// Return a reference to the transport SA if there is an SA on the link, and if it is established.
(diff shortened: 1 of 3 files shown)


adapter/ph/src/peer_table.rs
```diff
@@ -92,10 +108,14 @@ impl PeerState {
     }
 }
 
+type AtomicLinkId = atomic::AtomicU32;
+const _: () = assert!(std::mem::size_of::<AtomicLinkId>() == std::mem::size_of::<LinkId>());
```
mkolehmainen, line 112: Curious what is the reason for this assert here?

adapter/ph/src/peer_table.rs
```diff
@@ -145,6 +166,9 @@ impl PeerTable {
             return;
         };
         self.sa_to_link.remove(&peer_state.substrate_addr);
+        for name in peer_state.special_names {
+            self.special_peers[name].store(0, Ordering::Relaxed); // note; no useful ordering possible here
```
mkolehmainen: So is the SpecialPeerName still sitting in the special_peers map? Just with link_id set to zero?  Is that what you want?

adapter/ph/src/peer_table.rs
```diff
@@ -163,7 +187,27 @@ impl PeerTable {
...
+        id
+    }
+
+    pub fn lookup_special_peer(&self, name: SpecialPeerName) -> Option<LinkId> {
+        // synchronizes with the Release in VacantPeerTableEntry::insert();
+        // ensures anyone who reads from the slab following this sees the peer
+        // (assuming of course it hasn't been removed!)
+        let id = self.special_peers[name].load(Ordering::Acquire);
+
+        if id == 0 {
```
mkolehmainen: Ah. I see the zero check here.

mkolehmainen (dismissed):
Looks good.  I'm trying to get my mind around fencing.  Just some idle questions but otherwise Looks Good To Me!!

adapter/ph/src/peer_table.rs
cpacejo: There's no way in Rust (that I can find) to derive an atomic type corresponding to a given integer type.  So here I just have to declare `AtomicLinkId` as 32-bits which happens to match the declaration of `LinkId`.  So the assert is to make sure that they stay the same (so if e.g. we ever change the definition of `LinkId` we're forced to change it here).

adapter/ph/src/peer_table.rs
cpacejo: Yes, an `EnumMap` is better understood as an array – all the elements are always present, there's no way to delete them.  So I use 0 as the "empty" value.  Oh!  I should use `UNKNOWN_LINK_ID` here though instead of `0`.

cpacejo:
> Looks good. I'm trying to get my mind around fencing. Just some idle questions but otherwise Looks Good To Me!!

It took me a while to understand fencing too until I was able to boil it down to the following.

Release/acquire fencing lets you make the guarantee that, if I issue a release fence before writing to a variable X, and another thread issues an acquire fence after reading the thing I wrote to that variable X, then it's guaranteed to afterward also see everything else I did _prior_ to writing to X.  (It may also see other things which have happened since then, but is not guaranteed to.)

Visually I write this as (threads A + B, variables Foo and X):

[A] write(Foo) -> release -> [A] write(X) => [B] read(X) -> acquire -> [B] read(Foo)

This is one of the properties a mutex provides (and is also where the names come from): everything I do before _releasing_ a mutex is visible to whomever subsequently _acquires_ the mutex.

Without the release->acquire ordering, just because B reads what I wrote from X, has no bearing on whether it will also see what I wrote to Foo.  It might read some old cached value of Foo.  Or my write of X might get scheduled by the CPU and/or memory bus before my write of Foo.

Release and acquire fences can either be issued before/after (respectively) the write/read of the signal variable, or as part of the write/read itself.  The former is useful when you don't have control of the accesses to the signal variable (e.g. it's in some library).

So here, "X" (the signal variable) is "the reverse lookup table", and "Foo" is "the peer table".  So I'm giving the guarantee that, if someone sees a peer in one of the reverse lookup tables, it's guaranteed to see that it's in the peer table.  (Provided that no one has snuck in and removed it meanwhile.)  Without the fences, it's possible that the reader sees an old version of the peer table which doesn't yet have the peer in it, despite having read it from the reverse lookup table.

<https://en.cppreference.com/w/cpp/atomic/memory_order#Release-Acquire_ordering> and <https://en.cppreference.com/w/cpp/atomic/atomic_thread_fence> are my go-to references for this stuff by the way!

svjaffer (approved):

PR #539, opened 2024-11-15
Enhance ph-debug

### LINKED ISSUE #538: Enhance ph-debug
I really just wanted to be able to start and stop links from ph-debug, but once I got in there, there were a bunch of changes to make

svjaffer:
Did a bunch of cleanup in Ph-debug.  Main changes are:
- CLI mode
- Switched over to using the clap Subcommands, which cleaned some things up
- Added an RpcCommands enum to keep commands consistent between PH and PH-debug

resolves #538

### DIFF
--- a/adapter/ph-debug/src/main.rs
+++ b/adapter/ph-debug/src/main.rs
@@ -268,7 +268,7 @@ fn handle_watch(interval: u64, socket: &str) -> std::io::Result<()> {
 
     loop {
         let stream = &mut UnixStream::connect(socket).unwrap();
-        stream.write_all(b"COUNTERS\n")?;
+        stream.write_all(format!("{}\n", RpcCommands::Counters).as_bytes())?;
         stream.flush()?;
         let mut response = String::new();
         stream.read_to_string(&mut response)?;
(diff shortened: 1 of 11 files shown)


adapter/ph-debug/src/main.rs
```diff
@@ -17,92 +18,236 @@ use std::os::unix::net::UnixStream;
...
+use zpr::rpc_commands::RpcCommands;
 use zpr_ext::std::os::unix::net::{SocketAncillary, UnixStreamExt};
 const ANCILLARY_BUFFER_SIZE: usize = 128;
 
+macro_rules! basic_command {
+    ($comm:expr, $socket:ident) => {
+        basic_call_response_0($comm, $socket)
+    };
+    ($comm:expr, $socket:ident, $arg:tt) => {
+        basic_call_response_1(($comm), ($socket), ($arg))
```
cpacejo: Not an error but I think the parentheses should be unnecessary here?  (as with the other two cases)

adapter/ph-debug/src/main.rs
```diff
@@ -17,92 +18,236 @@ use std::os::unix::net::UnixStream;
...
 
-    #[arg(
-        long,
-        help = "",
-        long_help = "This option has no default, if no program is provided for a command that \
-                            needs a program, no program will be set"
-    )]
-    program: Option<String>,
+    /// Path to the Packet Handler's management socket
+    socket: String,
```
cpacejo: Possible to keep this as a flag (i.e. `#[arg(short, long)]`)?  I'm imagining in the near-ish future we can then make the flag optional and fall back to a default `/var/run/zpr/ph.sock` or pull it from env vars or such.

cpacejo (dismissed):
Nice!!!!!!!!  I am excited to use the prompt mode!!!

mkolehmainen (dismissed):
looks great!

adapter/ph-debug/src/main.rs
svjaffer: OK, switched it to that default

mkolehmainen (approved):
looks good to me!

PR #540, opened 2024-11-18
Add ability to get MTU from a tokio UdpSocket on macos

mkolehmainen:
Bit of a hack but appears to work.

### DIFF
--- a/adapter/ph/src/sys/macos/tun_pi.rs
+++ b/adapter/ph/src/sys/macos/tun_pi.rs
@@ -12,6 +12,9 @@ pub struct TunPi {
 impl TunPi {
     /// The size of a per-packet packet info structure.
     /// On macos this is `0`.
+    ///
+    /// TODO: Needs more exploration. The rust-tun code documents that there is PI
+    /// on macos, but our tests reading into buffers do not see it.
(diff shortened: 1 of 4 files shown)


cpacejo:
Does this query the path MTU or interface MTU?  For this API we want to query path MTU but if I'm reading the code correctly it appears to query the interface's MTU.  (Path MTU is essentially how we get from the kernel the information reported back by downstream hosts/routers via ICMP as to what _their_ MTU is.)

cpacejo:
I'm having trouble finding leads how to get path MTU on macOS but this might be a way, by reading the routing table: https://stackoverflow.com/questions/76852711/macos-path-mtu-cache-doesnt-gets-properly-updated-when-pmtu-discovery-is-set

cpacejo:
(That said I don't think we use this yet.  Its purpose is so we can properly reject too-large packets from the agent, rather than having them be blackholed, but that code doesn't exist yet.  So if to get the true PMTU requires a bit more work -- e.g. guess-and-checking by attempting to send our own ECHO packets, that is something we can put off for now.)

mkolehmainen:
OMG I did the wrong thing then.  I failed to pay attention to what that function was supposed to be doing.

PR #543, opened 2024-11-21
vs.zpr special name

cpacejo:
Reworked how the special peer assignment work (so it can happen after bringup), and also assign a special name to any adapter which connects with CN=vs.zpr.


mkolehmainen:
FYI the current set of configs for the demo assume "vs.zpr.org" as the vs CN

zpr/src/lib.rs
```diff
@@ -88,3 +88,25 @@ pub mod compression_mode {
...
+// [cn]
+// cnAtv = SEQ:cnAtv
+// [cnAtv]
+// type = OID:commonName
+// value = UTF8:vs.zpr
+// ```
+//
+// or similar.
+
+/// CN=vs.zpr
```
mkolehmainen: Should be "vs.zpr.org".

Also, do we really need to write this out in bytes?  I'd rather this was a plain old string in a config file and then we can build the asn structure in main or something similar.  This opaque byte array feels error prone.

But it's true that it should change infrequently if ever.

mkolehmainen (changes requested):
We just need to agree on the name.  If we switch to vs.zpr we need to update the relevant demo documentation and configs.

zpr/src/lib.rs
cpacejo: In order to match against it it needs to be a const value – this was the only way I could figure out to build it.  (E.g. the `asn1` crate doesn't provide `const` functions.)  I could rework this as a lookup in a map which is constructed when the system is brought up.

cpacejo:
Ah I missed `vs.zpr.org`.  Note though that [commonName](https://stackoverflow.com/a/5142550) (CN) is not a domain name – it can be any string like "Chris Pacejo" or "Visa Service", so there's no particular need/use for a `.org` suffix.  If we did want to identify special nodes/adapters using domain names, we could use a DN with [domainComponent](https://www.rfc-editor.org/rfc/rfc1279.html) (DC) values like `DC=vs, DC=zpr, DC=org`.  (CN is also fine to stuff a domain name into of course, my only hesitation in treating the CN field as a domain name though is the implication that _all_ nodes in the system have a CN which is a domain name, which feels overly restrictive to me.)

Hence – I don't feel too strongly, but I'd lean toward using CN, and one which _isn't_ reflective of the public DNS hierarchy just to emphasize that we're not relying or implying any relationship there.  Unless you feel strongly about sitting in the DNS hierarchy (and hence `vs.zpr.org`), I can find+replace in the configs as part of this PR.

zpr/src/lib.rs
cpacejo: Figured it out, did it with some const fn + macro magic :)

cpacejo:
For now I've switched to `vs.zpr` -- essentially adhering to the model the `vs.zpr` is not a domain name, just a well-known CN.  Also there was one place `vs.zpr.org` was used in a URL -- I've replaced that with its well-known IPv6 address, `[fd5a:5052::1]`, both to reflect that `vs.zpr.org` is no longer special, `vs.zpr` is not a domain name, and also that DNS may not be available to the admin tool.

zpr/src/lib.rs
mkolehmainen: Super.

mkolehmainen:
Ok looks good!

mkolehmainen (approved):

PR #545, opened 2024-11-25
Add timeout and cancel-in-progress for github workflows

### LINKED ISSUE #544: make sure our CI tasks have a time out

svjaffer:
Resolves #544

### DIFF
--- a/.github/workflows/go-visaservice.yml
+++ b/.github/workflows/go-visaservice.yml
@@ -11,6 +11,11 @@ on:
       - 'visaservice/mods/**'
       - 'visaservice/core/**'
       - 'visaservice/thrift/**'
+      - '.github/workflows/go-visaservice.yml'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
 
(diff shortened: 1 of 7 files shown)


.github/workflows/go-visaservice.yml
```diff
@@ -11,6 +11,11 @@ on:
       - 'visaservice/mods/**'
       - 'visaservice/core/**'
       - 'visaservice/thrift/**'
+      - '.github/workflows/go-visaservice.yml'
+
+concurrency:
```
svjaffer, line 16: This is the part that makes it so a new push to the PR will cancel a previous workflow.  Putting 2 instances in the same concurrency group guarantees that only 1 can run at a time and `cancel-in-progress` means that the previous instance will be terminated when a new one is started.

.github/workflows/go-visaservice.yml
```diff
@@ -11,6 +11,11 @@ on:
       - 'visaservice/mods/**'
       - 'visaservice/core/**'
       - 'visaservice/thrift/**'
+      - '.github/workflows/go-visaservice.yml'
```
svjaffer, line 14: I chose to add each workflow file to its own list of changed files that cause that workflow to be run.  Do double-check that I got all the names right and didn't make a copy-pasta bug!

cpacejo (approved):
Love it!!

PR #548, opened 2024-11-26
first version of a visa service test tool

### LINKED ISSUE #468: visa service test tool
The visa service test tool is part of our general requirement to produce a set of conformance tools for the reference implementation.

This story is about getting a start of this tool.  The idea is to be able to start a visa service with some "known" policy, point this tool at the visa service and have it check the various APIs.  For this story we will check the Visa Service API. So the tool should check:

- hello
- authenticate
- de_register
- authorize_connect
- agent_disconnect
- ping
- request_visa

mkolehmainen:
Possibly a little half baked, but the idea is that this is a start on a conformance tool for a visa service.  This tests very little right now.  

Eventually I think there will be a lot of tests.  I will add some tasks to extend this.

What it can do now is query the admin interface for a policy, use that policy to mock up a node connection, then use the policy to try a valid and invalid visa request.


visaservice/vst/pkg/packets/generator.go
```diff
@@ -0,0 +1,93 @@
...
+const ProtocolTCP = uint32(6)
+
+// For TCP, generates a SYN packet.
+// Returns (PACKET, L3TYPE, ERROR)
+func GeneratePacket(srcAddr, dstAddr netip.Addr, endpoint *polio.Scope) ([]byte, int, error) {
+	if endpoint.Protocol != ProtocolTCP {
+		return nil, 0, fmt.Errorf("only TCP is supported")
+	}
+
+	sport := 1024 + rand.Intn(65535-1024)
```
svjaffer: I don't know Go, but could these be constexprs or something equivalent?

visaservice/vst/pkg/testfw/suite.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+func RunTests(tests []Tester, vsAddr, adminAddr netip.AddrPort, nodeCert *x509.Certificate, log *zap.Logger) (*Scorecard, error) {
+	zlog := log.Sugar()
+	card := NewScorecard(len(tests))
+	state := NewTestState(vsAddr, adminAddr, nodeCert, zlog)
+	defer state.Close()
+	for _, test := range tests {
+		if err := RunTest(test, state, card); err != nil {
+			return card, fmt.Errorf("test %s failed: %w", test, err)
```
svjaffer, line 22: Does this return out on the first failure and not run subsequent tests?

visaservice/vst/pkg/testfw/state.go
```diff
@@ -0,0 +1,132 @@
...
+
+type TestState struct {
+	vsAddr      netip.AddrPort // visa service address for thrift api
+	adminAddr   netip.AddrPort // visa service admin HTTPS api
+	NodeCert    *x509.Certificate
+	Log         *zap.SugaredLogger
+	policy      *polio.Policy   // policy extracted from GetCurrentPolicy, may be nil
+	adminClient *vsadmin.Client // use GetAdminClient
+	node        *mocks.Node     // use GetNode
+	lastOctect  uint32          // last octect of the IP address of the node
```
svjaffer: What is this used for?  Also, no second c: octet

visaservice/vst/pkg/tests/helpers.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func connectAdapter(node *mocks.Node, crec *plc.ConnectRec, dockAddr netip.Addr, octect uint32) (*vsapi.Agent, error) {
+	claims := make(map[string]string)
+	claims["zpr.adapter.cn"] = crec.CN
+	if crec.HasAddr() {
+		claims["zpr.addr"] = crec.Addr.String()
+	} else {
+		// Hmm, just make one up?
+		claims["zpr.addr"] = fmt.Sprintf("fd5a:5052:1::%d", octect)
```
svjaffer: Given the easy overflow of one octet, I'd suggest either passing in a full IP address to this function or incrementing the port instead.

visaservice/vst/pkg/testfw/state.go
svjaffer: OK, I see where it's used.  As mentioned there, I would personally have the function to get the next octet get the IP address instead, since that's the logical type you're producing.  Also, one octet can overflow very easily and while we don't have 256 tests yet, that's not outside the bounds of reality.

But this is test code, so if it's too much of a hassle, it's fine to leave it be.

visaservice/vst/pkg/packets/generator.go
mkolehmainen: No C++ constexpr.  This makes a rand source port for each packet generated.  But I can name those constants.  I expect a future version of the suite to explicitly set up the RNG so that the "randomness" can be repeated.

visaservice/vst/pkg/testfw/suite.go
mkolehmainen: I have added more comments to help clarify.  If error is returned that is considered exceptional and it does halt the suite of tests.  However a test can fail by calling a fail function, which does not prevent subsequent tests from running.

visaservice/vst/pkg/testfw/state.go
mkolehmainen: I will patch this up.

svjaffer (approved):
Thanks for the changes!  Looks good!

PR #550, opened 2024-12-02
Visa adapter part 3

cpacejo:
Refactoring of internal bind logic to allow for "internal" calls for issuing binds to/from the internal adapter.  (Mostly moved code, not new code.)

Sorry for the big messy review!  Importantly now though -- traffic now passes between the VS adapter and the node's local adapter!

### DIFF
--- a/adapter/ph/src/assembly.rs
+++ b/adapter/ph/src/assembly.rs
@@ -3,11 +3,14 @@ use crate::buffer_stack::BufferStack;
 use crate::capture_worker::CaptureWorker;
 use crate::config;
 use crate::counters::*;
+use crate::defs;
 use crate::flow_control::FlowControl;
+use crate::forwarding_tables;
 use crate::km_cert_exchange::KmCertExchange;
 use crate::km_multiplexor::KmState;
 use crate::km_noise;
 use crate::link_state::{LinkEvent, LinkStateError, LinkType};
+use crate::mgmt;
 use crate::mgmt_processor_worker;
 use crate::peer_table;
 use crate::peer_table::PeerInsertError;
@@ -83,6 +86,12 @@ pub struct Assembly {
     pub self_noise_keypair: Option<NoiseKeypair>,
     pub peer_noise_keypair: Option<NoiseKeypair>,
     pub certx: Option<KmCertExchange>,
+}
+
+pub enum AddRouteError {
+    BindFailed(mgmt::requests::BindAgentAddressError),
+    PeerGone,
+    PftFull,
 }
 
 impl Assembly {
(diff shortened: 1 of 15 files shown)


adapter/ph/src/assembly.rs
```diff
@@ -191,6 +204,56 @@ impl Assembly {
...
+
+        let ingress_tether_id = ingress_peer_state
+            .pft
+            .insert(pep)
+            .map_err(|()| AddRouteError::PftFull)?;
+
+        Ok(ingress_tether_id)
+    }
+
     pub fn hack_default_policy(&self, ingress_link_id: LinkId) -> Option<NonZero<LinkId>> {
```
mkolehmainen: I get that this is a hack of some sort, but would you add a comment about what this is doing?  It's not clear to me how this function relates to a policy.

adapter/ph/src/mgmt/handlers.rs
```diff
@@ -305,7 +304,10 @@ pub async fn handle_bind_agent_address_request(
         dst_port,
     );
 
-    let ingress_link_id = pkt.metadata().ingress_link_id;
+    let Some(ingress_link_id) = NonZero::new(pkt.metadata().ingress_link_id) else {
+        // who sent this??
+        return Ok(());
```
mkolehmainen, line 313: Should this log something? Is this not an error?

adapter/ph/src/mgmt/handlers.rs
```diff
@@ -314,97 +316,73 @@ pub async fn handle_bind_agent_address_request(
...
+                    .await;
+
+                    return Ok(());
+                }
+
+                Err(super::dock::BindAgentAddressError::AddRouteError(
+                    assembly::AddRouteError::PftFull,
+                )) => {
+                    // PFT full; respond with error message
+                    // TODO: maybe tick a counter somewhere?
```
mkolehmainen, line 364: Yes! Let's add project tasks to address these counter needs.

mkolehmainen (dismissed):
Looks good. Made a couple minor commens.

adapter/ph/src/mgmt/handlers.rs
cpacejo: I added a log statement -- for context though this is part of my experiment/campaign to use `Option<NonZero<...>>` types in more places to add some type safety around "unknown" 0 integer values.  (The Rust compiler has special support for this construct such that it compiles the same as a plain integer.)  So this check should go away as I push that pattern into more places.

adapter/ph/src/mgmt/handlers.rs
cpacejo: Added a draft for now will flesh out later!

adapter/ph/src/assembly.rs
cpacejo: Done!

mkolehmainen (approved):
Looks great, thanks!

PR #553, opened 2024-12-02
change visa service CN from 'vs.zpr.org' to 'vs.zpr'

### LINKED ISSUE #551: regenerate vs key with "vs.zpr" CN

mkolehmainen:
Visa service is know known to have CN = 'vs.zpr'.  

- Regenerated the certificate holding the noise key.
- Updated the attribute in policy.


cpacejo (approved):
Fantastic thanks for doing this!

PR #554, opened 2024-12-02
Launch visa support service

### LINKED ISSUE #27: node listens for and accepts a visa service connection
See our PKI write up (*). 

The node during startup will accept a connection from the visa service adapter based on a well known value in the x509 certificate CN field.  Once the node has determined that the VS adapter has connected, it needs to access the visa service registration function through the VS thrift API (see org-zpr/zpr-core#146)
Previously Blocked by:
- org-zpr/zpr-core#17
- org-zpr/zpr-core#419
- org-zpr/zpr-core#34

cpacejo:
Launches the visa support service in the node.

Also a couple other changes:

* `--agent-addr` is now required for nodes also, to indicate the address of their local adapter (on which the VSS runs).
* VSS API now accepts `SocketAddr` instead of string for host address
* included some random commit I had kicking around something to do with displaying a TUN error?


mkolehmainen (dismissed):
Looks good to me but best wait for tests to pass before merge.

svjaffer (approved):

PR #555, opened 2024-12-02
agent_addr now a global config & required for nodes

mkolehmainen:
Once @cpacejo #554 goes through this patch makes the config files for milestone2 compatible with the new main_args.


cpacejo (approved):
Thank you!!!!

PR #559, opened 2024-12-03
Libnode request simplifications

cpacejo:
Backlog of some changes I've had kicking around my repo for a while.  The main changes here are to make the request methods return the responses, so we no longer need the request ID mechanism.  I also simplified the return types so they just follow the pattern `Result<ActualResponse, VSClientError>`.  (All in service of simplifying integration into the PH…)


mkolehmainen (approved):
Looks good to me!

PR #568, opened 2024-12-03
add timeout to libnode vss thrift connect

mkolehmainen:
Add 3 sec timeout to VSS connect routine.  Was way to slow to timeout.

Still see MICV errors when running node in debug. Not sure which stream they are on.


cpacejo (approved):
Looks good I'll merge it into my branch!

PR #571, opened 2024-12-04
Set up a new policy file for basic-test and capture-test

svjaffer:
zpr-a needs to be able to ping zpr-b on both of its addresses and vice-versa.


examples/milestone2/policies/basic-test.yaml
```diff
@@ -0,0 +1,84 @@
...
+      - [zpr.adapter.cn, eq, vs.zpr]
+    admin_attrs:
+      - [zpr.adapter.cn, eq, client.zpr.org]
+
+  datasources:
+    zpr.adapter:
+      api: validation/1
+      authority:
+        encoding: pem
+        cert_data: $import[../vs/ca-cert.pem]
```
mkolehmainen, line 35: This import here is resolved during compilation.  Is there actually a ca-cert.pem file in a directory above this named "vs"?  I would add the ca-cert.pem to the policies dir and then you don't need any "../vs/".  Or maybe I'm missing something?

mkolehmainen: Oh wait, you are putting this in examples milestone2?  Why not just add a "data" directory or something under `adapter/ph/test` ?  Anyway, I'm going to approve, but I think it maybe is a little odd to put this and the test related stuff in the milestone2 examples dir.

mkolehmainen (approved):
Psyched that this works! But have some reservations about locations of files.

svjaffer:
See https://github.com/org-zpr/zpr-core/pull/585

PR #578, opened 2024-12-04
update hashbrown to avoid vulnerability CWE-502

cpacejo:
Addresses https://github.com/advisories/GHSA-wwq9-3cpr-mm53

### DIFF
--- a/visaservice/admin/Cargo.lock
+++ b/visaservice/admin/Cargo.lock
@@ -385,9 +385,9 @@ dependencies = [
 
 [[package]]
 name = "hashbrown"
-version = "0.15.0"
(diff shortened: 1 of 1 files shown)


mkolehmainen (approved):
It's odd that this only changes a lock file.  Shouldn't this alter the Cargo.toml file too?  Anyway, great!

cpacejo:
> It's odd that this only changes a lock file. Shouldn't this alter the Cargo.toml file too? Anyway, great!

The .lock file is the ultimate source of "exactly what versions we depend on".  Since we don't depend on it directly, this is the best we can do without adding said dependency.

PR #581, opened 2024-12-04
Fix pre-commit hooks to run from any directory in the repo

svjaffer:
Also cleaned up checks on a directory that no longer exists. Finally, accounted for the movement of some code into libnode

### DIFF
--- a/tools/pre-commit
+++ b/tools/pre-commit
@@ -8,7 +8,7 @@ GIT_PH_DIR="$GIT_BASE_DIR/adapter/ph"
 GIT_LIBNODE_DIR="$GIT_BASE_DIR/libnode"
 
 # Check if there are any files in PH that are changed by this commit (not counting deletes)
-if [ ! git diff HEAD --cached --quiet --diff-filter=d --relative="$GIT_PH_DIR" ]
+if  ! git diff HEAD --cached --quiet --diff-filter=d --relative="$GIT_PH_DIR"
 then
     # Checks to make sure cargo fmt has been run on ph
     cargo fmt --manifest-path "$GIT_PH_DIR/Cargo.toml" --check --quiet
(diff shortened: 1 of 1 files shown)


tools/pre-commit
```diff
@@ -3,69 +3,59 @@
...
-	echo "Please clean up warnings with tests in ph before committing"
-	exit 1
-fi 
-
-# Checks to make sure cargo fmt has been run on cd
-cargo fmt --manifest-path adapter/cd/Cargo.toml --check --quiet
-
-if [ $? -ne 0 ]
+    # Checks to make sure cargo fmt has been run on ph
+    cargo fmt --manifest-path $GIT_PH_DIR/Cargo.toml --check --quiet
```
cpacejo: Should have " " around the variable expansion (i.e. `"$GIT_PH_DIR"`) to avoid issues with word splitting due to whitespace.  (Same comment on other occurrences in this file.)

(Note that this is actually _not_ needed when assigning to a variable as above, since word splitting is not performed in that case.)

tools/pre-commit
```diff
@@ -3,69 +3,59 @@
 # Git pre-commit hooks. Checks to ensure code is formatted and compiles without warnings
 # when a user tries to commit. Can be over-ruled by running git commit with --no-verify
 
-# Checks to make sure cargo fmt has been run on ph
-cargo fmt --manifest-path adapter/ph/Cargo.toml --check --quiet
+GIT_BASE_DIR=`git rev-parse --show-toplevel`
```
cpacejo, line 6: Very nit: backtick syntax is largely superseded by `$( )`.  I don't think it's deprecated or anything, but `$( )` is newer and better.  (Notable use case is nesting.)

cpacejo (dismissed):
Looks good just a couple bash nits!!

cpacejo (approved):
Perfect thanks!

PR #589, opened 2024-12-09
Log categories

cpacejo:
Promulgate the use of `tracing`'s "target" specifiers for log messages.  I've defined a handful of categories (corresponding to "functional areas" or "cohesive things that might need debugging") and applied them to every logging statement.  This way, we can turn on/off logging for each area independently.

Support for actually filtering messages (e.g. via the command line and/or debug tool) will come in a separate PR.

Also, deleted one or two leftover chatty log messages from M2 debugging.

Also, automated Display/Debug impls for a couple KM errors.

No functional changes.

### DIFF
--- a/adapter/ph/src/km.rs
+++ b/adapter/ph/src/km.rs
@@ -18,6 +18,7 @@ use std::fmt::Debug;
 use std::future::Future;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
+use thiserror::Error;
 use tokio::sync::mpsc;
 use tokio::time;
 use tokio_util::sync::CancellationToken;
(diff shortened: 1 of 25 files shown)


adapter/ph/src/km.rs
```diff
@@ -18,142 +19,78 @@ use std::fmt::Debug;
...
+use thiserror::Error;
 use tokio::sync::mpsc;
 use tokio::time;
 use tokio_util::sync::CancellationToken;
 use tracing::*;
 use zerocopy::FromBytes;
 use zpr;
 
-#[derive(Debug)]
+#[derive(Debug, Error)]
```
mkolehmainen, line 30: thanks for cleaning this up!

mkolehmainen (approved):
Great stuff!

PR #590, opened 2024-12-09
Add full link teardown & restart capabilities

### LINKED ISSUE #569: Exiting the PH with ^C should send a ZDP disconnect/goodbye message and close links.

svjaffer:
And RPCs to do those things

### DIFF
--- a/adapter/ph/src/link_state.rs
+++ b/adapter/ph/src/link_state.rs
@@ -667,7 +667,7 @@ impl LinkStateWrapper {
     /// Validate a received shutdown request
     /// Does not transition
     /// Generates no packets
-    fn process_terminate_request<'pktbuf>(
+    fn process_terminate_request(
         &self,
         _asm: &Assembly,
         reason: TerminateReason,
@@ -691,7 +691,7 @@ impl LinkStateWrapper {
     /// Initiate the shutdown of the link
     /// Transitions to Closing from any running state
     /// Generates a Terminate Request packet
-    fn initiate_close<'pktbuf>(
+    fn initiate_close(
         &self,
         asm: &Arc<Assembly>,
         reason: TerminateReason,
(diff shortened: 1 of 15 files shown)


adapter/ph/src/link_state.rs
```diff
@@ -649,46 +664,217 @@ impl LinkStateWrapper {
...
+            Ok(())
+        }
+    }
+
     /// Initiate the shutdown of the link
-    /// Transitions to Closed from any running state
-    pub fn initiate_close(&self) {
+    /// Transitions to Closing from any running state
+    /// Generates a Terminate Request packet
+    fn initiate_close<'pktbuf>(
```
cpacejo: This shouldn't need `'pktbuf` (there are no types referencing it).

adapter/ph/src/link_state.rs
```diff
@@ -649,46 +664,217 @@ impl LinkStateWrapper {
         }
     }
 
-    #[allow(dead_code)]
+    /// Validate a received shutdown request
+    /// Does not transition
+    /// Generates no packets
+    fn process_terminate_request<'pktbuf>(
```
cpacejo: Same here re: `'pktbuf`

adapter/ph/src/link_state.rs
```diff
@@ -649,46 +664,217 @@ impl LinkStateWrapper {
...
+        let link_type = self.link_type;
+        let task_asm = asm.clone();
+        tokio::task::spawn_local(async move {
+            info!("Link {} is clearing its state", link_id);
+
+            task_asm.peer_table.clear_peer_state(link_id);
+
+            if link_type == LinkType::AdapterToNode {
+                task_asm.tun_ctl.set_carrier(false).unwrap();
+            }
```
cpacejo: I'm not sure it actually matters given the context this is called from, but would it make sense to perform the above three operations immediately (rather than in a task), to give the caller the stronger guarantee that "the peer state is gone, and the carrier is down"?

cpacejo (dismissed):
Just a couple super minor nits!  Looks good!!

cpacejo (approved):
Looks good go go go!

PR #592, opened 2024-12-12
finer-grained logging options

cpacejo:
Added `--quiet`, and now both it and `--debug` take a list of logging targets to quiet down (= only FATAL and ERROR), or noisy-up (everything down to DEBUG).  Also `ALL` is accepted, specifying all targets.

### DIFF
--- a/adapter/ph/src/config.rs
+++ b/adapter/ph/src/config.rs
@@ -53,8 +53,11 @@ pub struct Config {
     /// should be left as None so that the kernal can pick a free one.
     pub tun_if: Option<String>,
 
-    /// Enable debug logging.
-    pub debug: bool,
+    /// Enable debug logging for specified targets, or ALL
+    pub debug: Vec<String>,
+
+    /// Disable info & warnings for specified targets, or ALL
+    pub quiet: Vec<String>,
 
     /// Required for adapter - the node dock address on substrate.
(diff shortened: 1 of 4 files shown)


adapter/ph/src/config.rs
```diff
@@ -53,8 +53,11 @@ pub struct Config {
     /// should be left as None so that the kernal can pick a free one.
     pub tun_if: Option<String>,
 
-    /// Enable debug logging.
-    pub debug: bool,
+    /// Enable debug logging for specified targets, or ALL
+    pub debug: Vec<String>,
```
n3wscott, line 57: it would be nice to have a way to know what values are possible here, does rust have a way to add that info to the help message?

mkolehmainen: yes I agree!

mkolehmainen (approved):
Looks great, but would be so awesome to have help tell us what target values are possible.

adapter/ph/src/config.rs
cpacejo: I… intended to do this and forgot????

cpacejo:
Not sure yet the best approach to adding a way to show this list with `clap`.  Will commit this as-is and opened #612 I'll address soon.

Issue #602, opened 2024-12-16
Github actions shouldn't rebuild PH


svjaffer:
Right now, every time we add a new action for PH-related tests, it has to rebuild PH for that action regardless of whether it's already been built or not.  Fix it to save time!

mkolehmainen:
Also see if relevant - https://github.com/earthly/setup-ucacher

PR #605, opened 2024-12-16
Add a rust cache to speed builds up

n3wscott:
Relates to https://github.com/org-zpr/zpr-core/issues/602

Baseline

```
ph-debug-build: 53s
ph-build: 2m 51s

basic int-test: 5m 19s
cap int-test: 4m 51s
```

With `Swatinem/rust-cache`, on a Retry build: 

```
ph-debug-build: 31s
ph-build: 1m 16s

basic int-test: 3m 36s
cap int-test: 3m 24s
```

I attempted to add artifact uploading and caching but we need a better plan for this. My recommendation would be to have each project build their artifacts and then release them, and we separate out the integration tests into its own github action file that might end up running only hourly. The local artifact generation seemed to save a tiny bit amount of time but really not much because the adapter rust action is asked to also build the go visa service and all the rust visa service components as well.

### DIFF
--- a/.github/workflows/rust-adapter.yml
+++ b/.github/workflows/rust-adapter.yml
@@ -25,6 +25,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
+    - uses: actions-rust-lang/setup-rust-toolchain@v1
+      with:
+        cache-workspaces: |
+          adapter/ph
+          adapter/ph-debug
+          zpr-ext
+          cslab
+          cbpf-rs
 
     - name: Install Deps
       timeout-minutes: 2
@@ -62,6 +70,15 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
+    - uses: actions-rust-lang/setup-rust-toolchain@v1
+      with:
+        cache-workspaces: |
+          adapter/ph
+          adapter/ph-debug
+          zpr-ext
+          cslab
+          cbpf-rs
+
     - name: Build + test ph
       timeout-minutes: 5
       run: |
@@ -75,6 +92,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
+    - uses: actions-rust-lang/setup-rust-toolchain@v1
+      with:
+        cache-workspaces: |
+          adapter/ph
+          adapter/ph-debug
+          zpr-ext
+          cslab
+          cbpf-rs
     - name: Check format
       timeout-minutes: 5
       run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/rust-adapter.yml
```diff
@@ -25,6 +25,14 @@ jobs:
...
     steps:
     - uses: actions/checkout@v4
+    - uses: Swatinem/rust-cache@v2
+      with:
+        workspaces: |
+          adapter/ph
+          adapter/ph-debug
+          zpr-ext
+          cslab
+          cbpf-rs
```
cpacejo, line 35: Also `libnode` I think should be in this list?  (Although that has a slightly more complicated build cycle, as it has some code generated by the Thrift compiler.)

.github/workflows/rust-adapter.yml
```diff
@@ -25,6 +25,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
+    - uses: Swatinem/rust-cache@v2
```
cpacejo, line 28: Does this pin this dependency at exactly the `v2` tag, or does it do some sort of semver matching?

(Ideally we could even pin this at a specific commit hash… maybe I'm overly cautious but my worry with "hyperlink" dependencies is they get nefariously, or incompetently, substituted for something else at some point in the future.)

cpacejo (approved):
Nice!!  A couple small nits but I think this is fine as-is!

.github/workflows/rust-adapter.yml
n3wscott: This is an issue across the board in the repo, though I am betting we are not pulling in many 3rd party actions. I do know of some automation to help with this, let me make a followup issue to tackle that.

.github/workflows/rust-adapter.yml
n3wscott: I would recommend checking in the thrift compiled code.

Correct me if I am wrong, it seemed `libnode` was only invoked because of the integration tests? Does the main line `ph` code import/expect `libnode` as well?

n3wscott:
I am going to merge this and follow up to address https://github.com/org-zpr/zpr-core/issues/608 this afternoon.

.github/workflows/rust-adapter.yml
cpacejo: Yes, `ph` uses `libnode` to communicate with the visa server.

Sarah & I are staunchly opposed to checking in generated code :)  Our primary arguments against it are, it can through neglect become out-of-sync with the true source, and it makes a mess of reviews.  However I'm sympathetic to the counterargument that forcing a build everywhere complicates the build environment.

A solution I've seen elsewhere that works well is to give the Thrift source into its own CI/CD pipeline which publishes the generated files as their own versioned Rust crate + Go package + etc.  Then the consumers simply depend on it as any other dependency from a public repo.  I've been meaning to look into that but haven't found the cycles / activation energy yet.  I opened #613 to tackle this, will try to get to it before the years' end.

PR #606, opened 2024-12-17
libnode: eliminate internal Mutex from VSConn

cpacejo:
This is a handful of changes which ultimately restructure the `VSConn` struct and its API such that it no longer needs the internal Mutex or Arc.  Instead, the burden of managing exclusive and shared access is placed on the caller, and enforced using borrow-checking.  This is achieved by separating the various "API" methods of `VSConn` from the `run()` method.  The former are attached to a new `VSConnHandle` type, which is both cloneable, and to which only shared access is needed; the latter remains on the `VSConn` type, which is no longer cloneable, and which needs mutable access.

### DIFF
--- a/adapter/ph/src/assembly.rs
+++ b/adapter/ph/src/assembly.rs
@@ -63,7 +63,7 @@ pub struct Assembly {
     pub agent_input: AgentInput,
     pub substrate_egress: SubstrateEgress,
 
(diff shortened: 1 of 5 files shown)


mkolehmainen (approved):
Looks great! Thanks!

PR #607, opened 2024-12-17
node: get node name from cert

### LINKED ISSUE #575: node agent name should come from certificate
But the agent PROVIDES needs to be separated from the agent name.  Right now I am using the name to fill in the provides field for the node agent.  That is temporary hack.  Actually, that whole "create node agent" is a little hack.  Provides should be separated out at same time as we make the name come from the cert.

cpacejo:
The only code left using the "name" command-line argument is the code that connects to the visa server, and this expects the name to match that found in the certificate subject.  So, let's simply derive the name from the certificate, and ditch the command-line argument.


n3wscott (approved):
seems like a good idea

PR #609, opened 2024-12-17
use a bash selector that works on mac and linux at the same time

n3wscott:
general change is `#!/bin/sh` --> `#!/usr/bin/env bash`

This change asks for the default bash entry binary from the env rather than assume it is in `/bin/` or `/usr/bin/`

This matters to mac because the default is bash3, so I need to install bash5. The only way to launch in bash5 is using the `env`.

### DIFF
--- a/diagrams/README.md
+++ b/diagrams/README.md
@@ -10,7 +10,7 @@ Then, set up a shell script wrapper to execute this jar file named
 `plantuml`.  The contents can be as simple as:
 
 ```sh
-#!/bin/sh
+#!/usr/bin/env bash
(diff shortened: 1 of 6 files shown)


cpacejo:
I don't object to this change, but alternatively I'd ask, what Bash 5-specific features are we using, vs. features present in any POSIX-compatible `sh` (which should always be at `/bin/sh`)?

cpacejo (approved):

n3wscott:
it could be we don't use any bash5 things, but the downside is bash3 is like 15 years old at this point and a lot of the flag names don't work with simple commands.

PR #610, opened 2024-12-17
use the real sha of github actions and configure dependabot to keep them up to date.

### LINKED ISSUE #608: Pin action versions to their SHA
Does this pin this dependency at exactly the `v2` tag, or does it do some sort of semver matching?

(Ideally we could even pin this at a specific commit hash… maybe I'm overly cautious but my worry with "hyperlink" dependencies is they get nefariously, or incompetently, substituted for something else at some point in the future.)

_Originally posted by @cpacejo in https://github.com/org-zpr/zpr-core/pull/605#discussion_r1888950938_

n3wscott:
This will cause github to create PRs to us to keep things in sync with the tag comment on the action.

See https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

Fixes https://github.com/org-zpr/zpr-core/issues/608

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    open-pull-requests-limit: 10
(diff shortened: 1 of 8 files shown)


cpacejo (approved):
That's amazing!!  From a quick Google it looks like dependabot will only trigger for security updates, correct?

n3wscott:
I am not totally sure, but I think dependabot will trigger on all updates in the actions ecosystem too! So we will have to watch it and we might have to tune it to perhaps 1 PR, or only check once a week.

Issue #597, opened 2024-12-16
Visa Service/ZPL Compiler/PH - make it work with ipv4


mkolehmainen:
A few facets here:

1. The PH needs to use IPv4-in-Ipv6 format for all IPv4 addresses (https://github.com/org-zpr/zpr-core/pull/611)
2. The visa service should not care, but maybe should expect all address to be IPv6 (which may include IPv4-in-IPv6)
3. The compiler which may sometimes encounter an address needs to convert IPv4 to the IPv6 format. (https://github.com/org-zpr/zpr-prototype/pull/10 https://github.com/org-zpr/zpr-core/pull/619, https://github.com/org-zpr/zpr-prototype/pull/12)
4. May need to duplicate the `one-node-v6-test.sh` as a `one-node-v4-test.sh`. (https://github.com/org-zpr/zpr-core/pull/621)

cpacejo:
Step 1 is this:
```
diff --git a/adapter/ph/src/link_state.rs b/adapter/ph/src/link_state.rs
index f41ab61..1c02cfe 100644
--- a/adapter/ph/src/link_state.rs
+++ b/adapter/ph/src/link_state.rs
@@ -552,7 +552,7 @@ impl LinkStateWrapper {
                     let connect_req = libnode::vsapi::ConnectRequest {
                         connection_id: Some(123), // unused
                         dock_addr: Some(
-                            IpAddress::new_from_std(&asm.agent_addresses[0]).into_v4_or_v6_octets(),
+                            IpAddress::new_from_std(&asm.agent_addresses[0]).v6.to_vec()
                         ),
                         claims: Some(
                             [
```

n3wscott:
with the changes landed, I changed the integration test to use ipv4 addresses and it does not quite work yet, getting 

```
      [vs] 2024-12-19T11:00:01.157-0800	info	visa denied: failed to resolve source ZPR address	{"error": "agent not found", "source": "10.253.2.1"}
      [vs] 2024-12-19T11:00:01.157-0800	info	===== dumping of agent database of size 4 =====
[...]

PR #611, opened 2024-12-17
Only support IPv6 inside of the adapter.

n3wscott:
Implements https://github.com/org-zpr/zpr-core/issues/597#issuecomment-2546443692 and deletes the method.

### DIFF
--- a/adapter/ph/src/net_defs.rs
+++ b/adapter/ph/src/net_defs.rs
@@ -78,14 +78,6 @@ impl IpAddress {
             IpAddr::V6(v6) => Self::new_from_std_v6(v6),
         }
     }
-
-    pub fn into_v4_or_v6_octets(&self) -> Vec<u8> {
-        if self.is_v4() {
-            self.read_as_v4().into()
-        } else {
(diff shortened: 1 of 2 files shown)


cpacejo (approved):
Looks good thanks!

adapter/ph/src/net_defs.rs
```diff
@@ -78,14 +78,6 @@ impl IpAddress {
             IpAddr::V6(v6) => Self::new_from_std_v6(v6),
         }
     }
-
-    pub fn into_v4_or_v6_octets(&self) -> Vec<u8> {
```
cpacejo, line 82: Rust does also allow squashing unused code warnings by adding an `#[allow(dead_code)]` annotation before the unused item or item group.  (But I agree here this code should go, I only added it as a bit of a hack to support exactly the above use case.)

n3wscott: good to know, I did not even bump into the dead code warning, I just deleted it because it is not used in the project.

PR #615, opened 2024-12-18
Commit the generated code

n3wscott:
Using https://github.com/github-linguist/linguist/blob/main/docs/overrides.md, we can mark which directories are full of generated code and the PR viewer will not expand them by default.

In the future we will add validation that shows you have an up to date generated code in each PR. 

The value in this is to explicitly see what changes each generation tool bump.


n3wscott:
Seems like we need to have a larger discussion. https://appliedinvention.slack.com/archives/C076TGY865R/p1734548758459509

PR #621, opened 2024-12-20
Working IPv4 and IPv6 integration tests

### LINKED ISSUE #597: Visa Service/ZPL Compiler/PH - make it work with ipv4

n3wscott:
Fixes https://github.com/org-zpr/zpr-core/issues/597

Editing the scripts to allow for the agents to be started as IPv4 or IPv6 addresses. Split into two runs because it would change the logic of the test to have more than two agents (but I also don't understand all the things yet) (and it would need a more complicated policy?)

Depends on https://github.com/org-zpr/zpr-prototype/pull/12


integration-test/parse_arguments.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
```
mkolehmainen, line 1: this is a great addition!

mkolehmainen (commented):
Looks good to me!  I'll let @svjaffer do the approve.

svjaffer (approved):
Looks great!

PR #624, opened 2024-12-30
use external vsapi

### LINKED ISSUE #613: Try to figure out a better CI/CD solution for Thrift stuff
My proposal is to set up an independent CI/CD pipeline for Thrift bindings, which publishes as an artifact Rust and Go crates/packages.  Consumers of these bindings then can simply depend on the generated code as they would any other public crate/package.

Probably also, we should separate the concept of crate/package version from protocol version.  A backwards-incompatible change in the protocol should warrant a new package name (or at least a new namespace within the package).  Whereas a major version bump in the package semver should solely indicate a local API change (e.g. the planned sync->async changes for the Rust bindings).

cpacejo:
Uses the vsapi module from <https://github.com/org-zpr/zpr-vsapi-rs>.

Uses a GitHub Personal Access Token (PAT) with read access to our repository.  This PAT is stored in the organization secret `ZPR_CICD_RO_TOKEN` and will need to be re-upped yearly.  (Though ideally, everything is made public in about a years' time anyway and we can do without this mechanism.)

All builds which directly or indirectly depend on other private repos now access those repos via GitHub HTTPS, using the above PAT communicated to `git` via its "askpass" mechanism.

For Rust, had to add `git` to our Dockerfile and set `CARGO_NET_GIT_FETCH_WITH_CLI=true` in order to enable Cargo to use the "askpass" mechanism.

For Golang, had to set `GOPRIVATE=github.com/org-zpr/*` so Go wouldn't try to query its own registry.  Also had to change some `safe.directory` setting to work around a [bug](https://github.com/docker-library/golang/issues/452) in `actions/checkout`.

Developers will have to _either_ run `git config --global url.git@github.com:.insteadOf https://github.com/` (which ends up in ~/.gitconfig), _or_ configure a PAT and use git askpass like the runners now do.  Also, anyone developing Golang will have to set `go env -w GOPRIVATE="github.com/org-zpr/*"` (which ends up in ~/.config/go/env).  Again, once we're public this requirement goes away.

Rather than HTTPS GitHub URLs, I did try to use SSH URLs, in order to allow local development to not need any new configuration.  Rust was happy with this (the runners then used Git URL rewriting to rewrite these as HTTPS URLs), but Golang was not (there is no way to tell Golang to use SSH access to GitHub).

There are a lot of files changed, but the vast majority are just updating "use"/"import" statements.


mkolehmainen (dismissed):
This looks good. Not sure I will fully understand how this works until I try to change the thrift. But I like the general approach.

mkolehmainen (approved):
Great!

svjaffer (approved):

PR #626, opened 2025-01-03
starter ZPL compiler for M3

### LINKED ISSUE #618: initial ZPL compiler framework
Get a cli up and running, get ability to parse a simple define and a simple allow.

- Does not emit a binary policy.
- Does not load config.

This is purposely tiny story to ease into the new compiler development.

mkolehmainen:
Probably throw away code for parsing the subset of ZPL that we think we need for M3.

This doesn't do anything useful yet except print out some diagnostics.


compiler/src/parser.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    // Construct an index that adds entries for all the AKAs.
+    let mut class_index: HashMap<String, String> = HashMap::new();
+    for (name, class) in classes.iter() {
+        class_index.insert(name.clone(), name.clone());
+        class_index.insert(class.aka.clone(), name.clone());
+    }
+
+    // Define statements create classes.
+    for statement in &mut statements {
```
cpacejo, line 50: I'm surprised this needs the `mut`?  (i.e. I don't see `*statement` being modified anywhere?)

Testing now it seems Rust doesn't complain either way which is weird to me!

cpacejo (approved):
This all looks good to me!

compiler/src/parser.rs
mkolehmainen: I'll remove in next PR.  I think I probably needed it at some point and then changed my approach.

PR #633, opened 2025-01-08
tighten up compiler in prep for incorporating config

### LINKED ISSUE #627: put all the example zpl from rfc into a test suite
Just testing parsing syntax, not building policy.

### LINKED ISSUE #630: handle the endpoints/users clause omission
First pass of the parser requires the endpoints AND users clauses.  Rework so that one of them can be omitted.  So basically these are the same:

- `allow users to access services`
- `allow endpoints with users to access services`
- `allow endpoints to access services`

mkolehmainen:
Now can sufficiently parse what we need for m3. Added a github action to run the tests etc on commits.  Next step is loading config so that we can actually build a policy.


compiler/src/parser.rs
```diff
@@ -71,14 +72,15 @@ pub fn parse(tokens: Vec<Token>) -> Result<Policy, CompilationError> {
     resolve_class_flavors(&mut classes)?;
 
     // Next parse all the allows.
-    for statement in &mut statements {
+    for statement in &statements {
         if statement[0].tt == TokenType::Allow {
-            let allow = parse_allow(statement, &class_index)?;
+            let allow = parse_allow(statement, &class_index, &classes)?;
 
             // The allow parser does not check the class flavors.
```
svjaffer: Is this "Remove this" comment for later, or something you meant to come back to here?

svjaffer (dismissed):
I appreciate the added comments!

compiler/src/parser.rs
mkolehmainen: oh whoops

svjaffer (approved):

PR #636, opened 2025-01-16
Send echo request/replies and track latency + loss

svjaffer:
Once the link becomes active, it begins sending echo requests to its partner every 3 seconds (in the future this will be configurable).  For each echo request, we track whether or not we got a response (success/failure counters) and how long it took to get a response (round trip time).  All of those stats, plus time since last link state change are now available in the `link show <id>` command through the debug tool.

### DIFF
--- a/adapter/ph/src/link_state.rs
+++ b/adapter/ph/src/link_state.rs
@@ -3,13 +3,18 @@ use crate::km::ZPIPair;
 use crate::km_multiplexor;
 use crate::logging::targets::LINK_STATE;
 use crate::mgmt;
+use crate::mgmt::core::SyncReqError;
 use crate::net_defs::IpAddress;
+use crate::sample_ring::SampleRing;
 use crate::special_peers;
 use crate::vs_worker;
 use crate::zdp::{ResponseCode, TerminateReason};
 
+use std::fmt::{Display, Formatter};
 use std::sync::{Arc, Mutex};
+use std::time::{Duration, Instant};
 use thiserror::Error;
+use tokio::time::MissedTickBehavior;
 use tracing::*;
 use zpr::LinkId;
 use zpr::ZPI_ENCRYPTED_HEADER_FLAG;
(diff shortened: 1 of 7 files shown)


adapter/ph/src/link_state.rs
```diff
@@ -191,14 +218,21 @@ impl LinkStateMachine {
             status: LinkStatus::Down,
             silent: false,
             agent_addresses: Default::default(),
+            last_state_change: std::time::Instant::now(),
         }
     }
+
+    pub fn set_state(&mut self, new_state: LinkState) {
+        self.state = new_state;
+        self.last_state_change = std::time::Instant::now();
```
mkolehmainen, line 227: If this the one true place where we can see link state changes, might be nice to have a debug statement emitted here.

adapter/ph/src/sample_ring.rs
```diff
@@ -0,0 +1,60 @@
+use std::ops::AddAssign;
+
+/// Simple struct that stores last SIZE samples of some data, overwriting the oldest after that
+pub struct SampleRing<Type, const SIZE: usize> {
+    pub samples: [Type; SIZE],
+    next_sample_idx: usize,
+    num_samples: usize,
+}
+
+impl<Type: std::marker::Copy + std::cmp::Ord + AddAssign, const SIZE: usize>
```
mkolehmainen, line 10: Ah! Here's some proper rust! ;)

Also is "Type" a special value for rust or just regular variable name?

adapter/ph/src/link_state.rs
```diff
@@ -770,14 +800,50 @@ impl LinkStateWrapper {
...
+                            .unwrap()
+                            .echo_timeout += 1
+                    }
+                    Err(_) => {
+                        peer.link_state_machine
+                            .locked_data
+                            .lock()
+                            .unwrap()
+                            .echo_failure += 1
+                    }
```
mkolehmainen, line 843: What happens if we get too many fails (for some value of "too_many")?  Is some other process watching for that? Or is that future work?

mkolehmainen (approved):
Approved, but I do have comments you can choose to address or not. Thanks!

adapter/ph/src/sample_ring.rs
svjaffer: It's a Rust generic.  I could have called it anything, but thought Type was clearest.  Typename would also work.  After the colon are the traits that Type must satisfy.

adapter/ph/src/link_state.rs
svjaffer: For now, nothing.  Future work will build upon this.

ETA: For clarity, there will need to be configuration around keep-alives.  Specifically, how often to send them, how many to send in a batch (for more accurate loss calculations), how long to keep latency/loss data for, and to your point here, how many failed keep-alives before we restart the link.

I'm trying to scope creep less on my tasks and so this is *just* sending keep-alives and gathering RTT measurements.  Another TODO coming out of this is to keep windowed loss metrics in the same way we're doing windowed RTT metrics here.  Sometimes you do care about the lifetime loss metrics of a link, but more often you're wondering what the condition is in the last _n_ minutes.

adapter/ph/src/link_state.rs
svjaffer: I like the idea.  That said, I think I'm going to leave it as-is for now because without the context of the link ID, the log is useless and I'm not sure I want to pass the link ID in just for the log message and moving this API up a level into the wrapper would cause issues with the mutex, so I'm going to think on this a bit and decide how I want to handle it.

Issue #629, opened 2025-01-07
from zpl and configuration, produce a policy binary


mkolehmainen:
Ability to create a policy binary using new style ZPL + configuration and policy for the m2 policies.

Output from the compiler should be able to be installed into the visa service.

PR #638, opened 2025-01-16
Combine the config with the ZPL to create the policy binary

### LINKED ISSUE #628: ability to load "configuration" for an m2 policy
Write a new zpl policy that mirrors the old m2 policies.  Include the configuration.  

Can just successfully parse the policy and load configuration.  Does not yet produce anything useful.

mkolehmainen:
Although this will create a binary that looks pretty good, I have not yet tested that the visa service can read this.

This partially implements a bunch of parsing operations, TODOs are scattered throughout.  There aren't enough tests.  But figured that there was a enough code/work here to make a good stopping point on this task.

Note that I just copied the "policy.proto" file into this project though it is a shared protcol buffer definition file used also in the prototype compiler and in the prototype visa service.   The rust pb library (prost) generates the rust code from the .proto file compile time through some magic in the `build.rs` file.

Next tasks (to be created in project):
- ensure binary policy works in the visa service (#629)
- more unit testing (#639)
- So far only "tested" on one of the M3 policies, so try the other one (ping+http). (#640)


.github/workflows/rust-compiler.yml
```diff
@@ -23,6 +23,8 @@ jobs:
     - name: check warnings etc
       timeout-minutes: 5
       run: |
+        sudo apt update
+        sudo apt install -y protobuf-compiler
```
cpacejo, line 27: Should we add this to the Docker image?

compiler/src/crypto.rs
```diff
@@ -0,0 +1,104 @@
...
+/// Create SHA256 digest of given byte array.
+pub fn sha256_of_bytes(data: &[u8]) -> Digest {
+    digest::digest(&digest::SHA256, data)
+}
+
+/// Convert a `ring::digest::Digest` to a hex string.
+pub fn digest_as_hex(digest: &Digest) -> String {
+    let mut s = String::new();
+    for b in digest.as_ref() {
+        s.push_str(&format!("{:02x}", b));
```
cpacejo, line 38: You can more succinctly (and efficiently, when it matters) write `write!(&s, "{:02x}", b)`.

compiler/src/fabric.rs
```diff
@@ -0,0 +1,317 @@
...
+    pub access_only: bool, // If true, this policy is only for access, not for setting up a connection
+    pub condition: Vec<Attribute>, // List of attributes that must be met for the policy to apply
+                           // TODO: withouts, constraints, etc.
+                           //       Actually, withouts are just attributes, eg (role, ne, marketing)
+}
+
+/// Debugging output
+impl fmt::Display for Fabric {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "revision: {}\n", self.revision)?;
```
cpacejo, line 71: `writeln!()` will give the `'\n'` for free.  (Just a option if you find it more readable -- the two are exactly equivalent on all platforms.)

compiler/src/weaver.rs
```diff
@@ -0,0 +1,416 @@
...
+    fn compute_revision(
+        &mut self,
+        policy_digest: Option<Digest>,
+        config_digest: &Digest,
+    ) -> Result<(), CompilationError> {
+        let mut revhash = Vec::new();
+
+        match policy_digest {
+            Some(d) => revhash.extend_from_slice(d.as_ref()),
+            None => panic!("error - call to new Binp but ZPL digest is not set"), // programming error
```
cpacejo: What's a "Binp"?

cpacejo (commented):
Almost done just the big files left!

compiler/src/weaver.rs
mkolehmainen: haha. old, worse name.  Removed.

.github/workflows/rust-compiler.yml
mkolehmainen: Yes I think that is best.  Created issue for this #651.

compiler/src/config.rs
```diff
@@ -0,0 +1,964 @@
...
+    pub provider: Option<Vec<(String, String)>>, // optional provider attributes
+}
+
+impl Config {
+    /// Attempt to lookup the given hostname in the configurations resolver table.
+    /// If the passed name does not map to an IP address, or if the name is not
+    /// found in the table, then None is returned.
+    pub fn resolve(&self, hostname: &str) -> Option<IpAddr> {
+        if let Some(hosts) = &self.resolver.hosts {
+            if let Some(ip) = hosts.get(hostname) {
```
cpacejo: I think here you could instead use `let xxx = yyy?;` to avoid the deep nesting of ifs.

compiler/src/config.rs
```diff
@@ -0,0 +1,964 @@
...
+                let addr: IpAddr;
+                // TODO: I suppose the resolver table could map a name to another name.
+                //       For now that is not supported.
+                addr = match ip.parse() {
```
cpacejo: Could also be written as `return ip.parse().ok();`.

compiler/src/config.rs
```diff
@@ -0,0 +1,964 @@
...
+    Ok(interaction)
+}
+
+/// Convert a TOML integer to a u8.  Returns None if the integer is out of range.
+fn toml_as_u8(v: &toml::Value) -> Option<u8> {
+    match v {
+        toml::Value::Integer(i) => {
+            if *i < 0 || *i > 255 {
+                return None;
+            }
```
cpacejo: Can also write the pattern as `toml::Value::Integer(&i) if i < 0 || i > 255 =>`.  (Doesn't make a difference here, just a bit more succinct.)

compiler/src/weaver.rs
```diff
@@ -0,0 +1,416 @@
...
+                ));
+            }
+            def_ts = Some(ts);
+        }
+        if def_ts.is_none() {
+            return Err(CompilationError::ConfigError(
+                "no default trusted service found in configuration".to_string(),
+            ));
+        }
+        let def_ts = def_ts.unwrap();
```
cpacejo: Maybe better as:
```
let Some(def_ts) = def_ts else {
    return Err(...);
}
```
so no `.unwrap()` is needed.

compiler/src/weaver.rs
```diff
@@ -0,0 +1,416 @@
...
+
+        // The only thing we care about is cert path.
+        if def_ts.cert_path.is_none() {
+            return Err(CompilationError::ConfigError(
+                "default trusted service must have a cert path".to_string(),
+            ));
+        }
+        self.fabric.default_auth_cert = def_ts.cert_path.clone().unwrap();
```
cpacejo: & here I think you can do:
```
let Some(ref cert_path) = def_ts.cert_path else {
    return Err(...);
}
self.fabric.default_auth_cert = cert_path.clone();
```

compiler/src/policybuilder.rs
```diff
@@ -0,0 +1,500 @@
...
+use crate::fabric::{Fabric, FabricService, ServiceType};
+use crate::polio;
+use crate::ptypes::Attribute;
+use crate::zpl;
+
+/// Updeate this if we change the protobuf. This is checked by visa service during deserialization.
+pub const SERIAL_VERSION: u32 = 41;
+
+/// This value for a PROC in a connect record means NO PROC.
+const NO_PROC: u32 = 0xffffffff;
```
cpacejo: Maybe better as `u32::MAX`?

compiler/src/policybuilder.rs
```diff
@@ -0,0 +1,500 @@
...
+/// That which can create a policy, requires a [Fabric] to do so.
+impl PolicyBuilder {
+    /// Create the builder. This sets some topical info in the policy.
+    ///
+    /// Once created, you should call [PolicyBuilder::with_max_visa_lifetime], then
+    /// [PolicyBuilder::with_fabric] (which does the real work), and finally
+    /// [PolicyBuilder::build] to get the compiled policy.
+    pub fn new() -> PolicyBuilder {
+        let utc: DateTime<Utc> = Utc::now();
+        let policy_date = utc.to_rfc3339();
```
cpacejo: If I'm reading the docs correctly this will output with a timezone `+00:00` I think?

I have a _slight_ preference for using the "Z" format variant (defined by [RFC 9557](https://datatracker.ietf.org/doc/html/rfc9557#name-updating-rfc-3339)), it's a bit more interoperable and is intended to imply nothing about a preferred time zone.  `.to_rfc3339_opts(..., true)` would give us this ([doc](https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.to_rfc3339_opts)). (Also gives us the ability to choose a fixed precision for the seconds field, which might be a good idea for consistency.)

cpacejo (dismissed):
Looks good!  Lots here I had to skim some of it, just mostly a couple nits tho

compiler/src/policybuilder.rs
mkolehmainen: yes that is nicer.

cpacejo (approved):

PR #641, opened 2025-01-17
update readme files

mkolehmainen:
Update main readme by removing M2 stuff and putting it into M2 readme. Add a couple links.


examples/milestone2/README.md
```diff
@@ -1,6 +1,116 @@
...
+    All other messages are ZDP.
+```
+
+```
+NOTES:
+
+(1) The auth_connect message is not fully filled in. We just include a couple of
+    claims: the adapater ZPR address and the CN from the adapter certificate.
+
+(2) The echo/echo-resp stuff is just for keep alive and I'm not sure what actual
```
svjaffer, line 100: I know this was just moved, but I'm not really sure what to make of this note.  The last sentence isn't true anymore.  The ZDP messages are echo request/echo response messages, which are defined in the RFC.

svjaffer (approved):
I think I'd like that footnote 2 to be reworded at the very least, per my comment below, but it's not a huge deal.

examples/milestone2/README.md
mkolehmainen: Huh, I'm not even sure we had those messages at all in M2.  However I'll just leave it alone since it's a piece of history now.

PR #653, opened 2025-01-23
Clean up packet handler logging output

### LINKED ISSUE #652: Clean up noisy logging
Logging during runtime is currently too noisy.  Quiet down per-flow logging now that the demo is over

svjaffer:
Running the system is too darn noisy!

### DIFF
--- a/adapter/ph/src/fastpath.rs
+++ b/adapter/ph/src/fastpath.rs
@@ -21,7 +21,7 @@ use crate::{compress, km};
 use blake3;
 use bytes::{Buf, BufMut};
 use std::time::SystemTime;
-use tracing::{debug, error, info, warn};
+use tracing::{debug, error, warn};
 use zerocopy::FromBytes;
 use zpr;
 use zpr_ext::std::mem::{drop_guard, DropGuard};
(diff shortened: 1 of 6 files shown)


adapter/ph/src/fastpath.rs
```diff
@@ -21,7 +21,7 @@ use crate::{compress, km};
 use blake3;
 use bytes::{Buf, BufMut};
 use std::time::SystemTime;
-use tracing::{debug, error, info, warn};
+use tracing::{debug, error, warn};
```
cpacejo, line 24: On these I've just started doing `tracing::*`, I can't think of a downside.  (No change needed for the PR though.)

adapter/ph/src/net_defs.rs
```diff
@@ -78,6 +78,10 @@ impl IpAddress {
             IpAddr::V6(v6) => Self::new_from_std_v6(v6),
         }
     }
+
+    pub const fn is_v6_unicast_link_local(&self) -> bool {
+        self.v6[0] == 0xfe && self.v6[1] == 0x80
```
cpacejo: Technically it's fe80::/10, though only fe80::/64 has been allocated.  Probably best to ignore the full fe80::/10 for consistency though.  (So anything > 0x80 and < 0xC0 if my math is right.)

cpacejo (approved):

PR #655, opened 2025-01-23
Add entries in the forwarding table based on visa destination

### LINKED ISSUE #637: Node installs visas in forwarding table
Use received visas to populate the forwarding table.

svjaffer:


adapter/ph/src/assembly.rs
```diff
@@ -214,6 +214,26 @@ impl Assembly {
...
+        let Ok(locked_ids) = self.peer_ids.lock() else {
+            return Err(());
+        };
+
+        for id in locked_ids.iter() {
+            let peer = self
+                .peer_table
+                .get(*id)
+                .expect("Peer IDs out of sync with peer table");
+            for addr in peer.link_state_machine.get_agent_addresses() {
```
cpacejo: Could we deadlock here?  If we're holding the peer IDs lock, and take some lock in the link state machine, while the link state machine process is holding its own lock, and then tries to access the peer IDs?  Maybe that doesn't happen, the LSM never accesses the peer ID list.

adapter/ph/src/assembly.rs
```diff
@@ -214,6 +214,26 @@ impl Assembly {
         return Ok(peer_id);
     }
 
+    /// Temporary? function to find a link based on the agent address
+    pub fn find_egress_link(&self, agent_addr: IpAddress) -> Result<NonZero<LinkId>, ()> {
```
cpacejo: Would `Option<NonZero<LinkId>>` may be a better choice?  (Since the error case carries no information…)

adapter/ph/src/mgmt/dock.rs
```diff
@@ -57,13 +67,45 @@ pub async fn bind_agent_address(
             match asm.vsconn.as_ref().unwrap().request_visa(visa_req).await {
                 Ok(vsapi::VisaResponse {
                     status: Some(vsapi::StatusCode::SUCCESS),
+                    visa,
                     ..
                 }) => {
+                    asm.counters[CounterType::VisaRequestSuccess].increment();
+                    // for now, just pull the destination address tether to set up forwarding
+                    let Some(octets) = visa.unwrap().visa.unwrap().dest else {
+                        asm.counters[CounterType::VisaRequestError].increment();
```
cpacejo: Should we only increment either success or error?

adapter/ph/src/mgmt/dock.rs
```diff
@@ -57,13 +67,45 @@ pub async fn bind_agent_address(
...
+                            "Could not parse visa".into(),
+                        ));
+                    };
+                    let addr = match octets.len() {
+                        4 => IpAddr::from(
+                            <[u8; 4]>::try_from(octets.as_slice()).expect("Bad IP length"),
+                        ),
+                        16 => IpAddr::from(
+                            <[u8; 16]>::try_from(octets.as_slice()).expect("Bad IP length"),
+                        ),
```
cpacejo: Could break this out into a separate function (e.g. in zpr-ext?)

adapter/ph/src/mgmt/dock.rs
```diff
@@ -57,13 +67,45 @@ pub async fn bind_agent_address(
...
+                        _ => {
+                            asm.counters[CounterType::VisaRequestError].increment();
+                            return Err(BindAgentAddressError::ParseError(
+                                "Could not parse visa dest address".into(),
+                            ));
+                        }
+                    };
+                    match asm.find_egress_link(addr.into()) {
```
cpacejo: We only use `addr` here right?  What if instead you add a method to `IpAddress` to construct it from a slice (rather than first converting to `IpAddr`)?  I.e. `impl TryFrom<&[u8]> for IpAddress`?

cpacejo (changes requested):

adapter/ph/src/net_defs.rs
```diff
@@ -114,6 +114,22 @@ impl From<IpAddr> for IpAddress {
     }
 }
 
+impl TryFrom<Vec<u8>> for IpAddress {
+    type Error = ();
```
cpacejo: Because we consume the `Vec`, we should return it in the error case.  i.e. `type Error = Vec<u8>`.  (This is the same for the builtin [Vec-to-array conversion](https://doc.rust-lang.org/std/primitive.array.html#impl-TryFrom%3CVec%3CT,+A%3E%3E-for-%5BT;+N%5D).)

integration-test/parse_arguments.sh
```diff
@@ -3,7 +3,7 @@
 # Usage message
 usage() {
     echo "Usage:"
-	echo "    $0 [--agent_protocol <IPv4|IPv6>]"
+	echo "    $0 [--agent_protocol <IPv4|IPv6>] [--num_agents <number of agents>"
```
cpacejo: Missing the closing `]`

cpacejo (changes requested):

cpacejo (approved):
Perfect let's go!!!

PR #657, opened 2025-01-27
update docs w/ respect to the thrift repos

mkolehmainen:
I just copied chris's note into the readme.  Eventually could use more docs in here for how to actually alter and rebuild the thrift.


cpacejo (approved):
Thanks!

PR #659, opened 2025-01-28
add and use two_way_queue

cpacejo:
two_way_queue is a special type of queue used for sending packets from datapath to mgmt.  Packets are send zero-copy (i.e. owned) to mgmt, but only exposed to mgmt as a reference.  Mgmt must copy the packet if it wants to hold it long-term.  When mgmt is done with the original packet, it is destructed into a buffer and returned to datapath.

This means that datapath retains ownership of all datapath packets.  This patch thus also ensures that mgmt is no longer interacting with the buffer stack, in preparation for making the buffer stack local per datapath lane.

This also means that mgmt and datapath now only communicate exclusively through queues, and nonblocking shared-memory operations (e.g. RCU).

### DIFF
--- a/adapter/ph/src/agent_output_worker.rs
+++ b/adapter/ph/src/agent_output_worker.rs
@@ -1,13 +1,18 @@
+use crate::adapter_tables::AltEntry;
 use crate::assembly::Assembly;
 use crate::classifier::{self, ClassifierResult};
+use crate::compress;
 use crate::config;
 use crate::counters::*;
 use crate::fastpath;
 use crate::net_defs;
 use crate::packet::Packet;
+use crate::queues::{AdapterManager, TryEnqueueError};
 use crate::sys::TunPi;
 use crate::sys::ZprTun;
+use crate::zdp;
 use crate::zprtun;
+use bytes::BufMut;
 use std::sync::Arc;
 use tokio::net::UnixDatagram;
 use tokio::select;
@@ -15,6 +20,8 @@ use tokio::select;
 #[derive(Copy, Clone)]
 pub struct Config {
     pub worker_index: usize,
+    pub buffer_count: usize,
+    #[allow(dead_code)]
     pub batch_size: usize,
 }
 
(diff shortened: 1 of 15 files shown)


adapter/ph/src/config.rs
```diff
@@ -68,6 +68,7 @@ pub struct Config {
 
 /// Configuration of data path & control plane topology.
 pub struct TopologyConfig {
+    /// Number of packet buffers to allocate per fastpath worker.
```
cpacejo, line 71: This comment is not quite correct yet (will be in follow-on PR).

adapter/ph/src/agent_output_worker.rs
```diff
@@ -147,6 +170,71 @@ impl Worker {
...
+        self.process_packet_post_classify(pkt, /* allow_bind_request */ true);
+    }
+
+    /// Post-classification portion of `agent_output` function.  Used for
+    /// re-injecting already-classified packets e.g.  which were held
+    /// awaiting bind.  `allow_bind_request` should be `true` for "real"
+    /// packets; `false` for packets re-injected from mgmt plane after
+    /// fulfilling a bind request (so as to prevent the theoretical
+    /// possibility of a packet loop).
+    pub fn process_packet_post_classify(&mut self, mut pkt: Packet, allow_bind_request: bool) {
```
cpacejo, line 182: Not new code; just moved from `fastpath` module.

adapter/ph/src/two_way_queue.rs
```diff
@@ -0,0 +1,222 @@
...
+        }
+    }
+
+    /// Receive a single returned item.  Blocks until there is such an item.
+    ///
+    /// Panics if called when no items are outstanding.
+    pub fn blocking_recv_return(&mut self) -> U {
+        assert!(self.outstanding > 0);
+        let ret = self.incoming_return_q.blocking_recv().unwrap();
+        self.outstanding -= 1;
```
mkolehmainen, line 78: Is there any worry about different threads mucking about with this outstanding counter?  Does rust prevent sender from being shared between threads?

adapter/ph/src/assembly.rs
```diff
@@ -341,7 +342,7 @@ pub mod test {
             .agent_addresses
             .unwrap_or(Vec::from([IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1))]));
         let buffer_stack = builder.buffer_stack.unwrap_or_else(|| {
-            let buf_storage = vec![Box::new([0u8; config::PACKET_BUFFER_SIZE]); 0];
+            let buf_storage = vec![Box::new([0u8; config::PACKET_BUFFER_SIZE]) as Box<[_]>; 0];
```
mkolehmainen, line 345: What is point of the `as Box<[_]>` bit?

mkolehmainen (approved):
Ok the queue item with its own return channel queue thing is neat. I could not see where the management side was coping packets but I'll look more closely when this is in main.

adapter/ph/src/two_way_queue.rs
cpacejo: Because `self` is `&mut` the compiler guarantees no concurrent access :)

adapter/ph/src/assembly.rs
cpacejo: Converts the array-boxes into slice-boxes.  (They still point to the same thing, they just also include the length.)

PR #660, opened 2025-01-28
tweaks to make new compiler work with visa service

### LINKED ISSUE #629: from zpl and configuration, produce a policy binary

### LINKED ISSUE #640: Create a couple of ZPL+Configuration policies and add to test regimen
We have a since policy so far for the M3 examples, add another one and actually add both to our compiler "integration" test.

mkolehmainen:
Includes further changes to the compiler to ensure compatibility with the visa service.

Also had to make some tweaks to the old visa service itself.  I don't think this breaks backwards compatibility with old YAML ZPL.

All the old visa service tests use old policy and old (actually even older) compiler.  Maybe worth adding a story to convert all the tests to new ZPL and make use of new compiler.  Not all features of visa service are supported by new ZPL yet though.


visaservice/core/visa.log
svjaffer, line 1: Is this file supposed to be here?

compiler/src/policybuilder.rs
```diff
@@ -305,27 +337,26 @@ impl PolicyBuilder {
...
     ) -> polio::Proc {
-        // In the prototype compiler we include endpoint information in the
-        // proc, but that is not used anymore so am leaving it out for
-        // now.  We will just use REGISTER but leave endpoints empty.  We
-        // will also set relevant flags.
-
         let mut proc = Vec::new();
 
         // Args for register are (NAME:String, Type:SvcT, ENDPOINTS:String)
+        //
```
svjaffer: Empty comment

visaservice/core/pkg/vsapi/vs.go
svjaffer, line 1: This seems like a duplicate file.  Should this be here?

compiler/src/weaver.rs
```diff
@@ -388,7 +434,9 @@ fn find_defined_service(
 /// the parent classes.  We ignore optional attributes.
 fn attrs_for_class(class_idx: &HashMap<String, &Class>, class_name: &str) -> Vec<Attribute> {
     let mut attrs = Vec::new();
-    let mut cl = class_idx.get(class_name).unwrap();
+    let mut cl = class_idx
+        .get(class_name)
+        .expect(&format!("class {} not found in class index", class_name));
```
svjaffer, line 439: FWIW, this could instead be
```.expect(&format!("class {class_name} not found in class index"));```

visaservice/core/visa.log
mkolehmainen: nope!

visaservice/core/pkg/vsapi/vs.go
mkolehmainen: yes, I'm confused as to how that got in there.  Let me look around.

mkolehmainen: I removed them.

compiler/src/weaver.rs
mkolehmainen: Ah yes, I noticed that form in some other source files.  I will try to use it more going forward.

svjaffer (approved):

PR #664, opened 2025-01-31
initial pass a runners for adapter and node

### LINKED ISSUE #663: create wrappers to handle network cruft required to run node/adapter
This is done when you can just open a terminal and run a node or adapter without having to do any network setup at all.

- Bonus points if this cleans up when it's done.
- Bonus for each non unix platform it runs on.  (I will attempt linux and mac for first version).

mkolehmainen:
New crate in adapter/runners that creates two binaries, one to run a node other to run an adapter.

Linux only for now.


adapter/runners/src/runner.rs
```diff
@@ -0,0 +1,51 @@
...
+
+/// Launch a node or adapter. Caller provides the launching function.
+pub fn runner(
+    alt_ph_bin: Option<PathBuf>,
+    config_file: &Path,
+    dry_run: bool,
+    launcher: LaunchF,
+    verbose: bool,
+) -> Result<(), LaunchErr> {
+    if sys::get_platform().has_root_perms() == false {
```
cpacejo, line 26: I wonder if we can drop root permissions after configuring the TUN devices.  E.g. with `setuid()` and `setgid()` to a value provided on the command line.

Also maybe we want to allow running as non-root if the TUN devices are already configured.

(Not needed in this PR but if you agree I'll open corresponding issues so we don't forget!)

cpacejo (dismissed):
Looks good ship it!

adapter/runners/src/runner.rs
mkolehmainen: I'm going to see about adding the privilege drop.  I was under mistaken belief that we needed root during normal operation.

mkolehmainen:
@cpacejo  - I added the root dropping and it all seems to work (tested only on linux -- there's a separate story to port to mac).

adapter/runners/src/sys/linux.rs
```diff
@@ -19,6 +22,91 @@ impl Platform for LinuxPlatform {
...
+
+        stat::fchmod(
+            dirfd,
+            stat::Mode::S_IRWXU | stat::Mode::S_IRWXG | stat::Mode::S_IROTH | stat::Mode::S_IXOTH,
+        )
+        .map_err(|e| PlatformErr::OsError(format!("fchmod failed: {}", e)))?;
+        Ok(())
+    }
+
+    fn drop_privledges(&self, username: &str, dry_run: bool) -> Result<(), PlatformErr> {
```
cpacejo, line 75: Spelling: `privilege` (I had to look it up lol)

cpacejo (approved):
Looks good!

PR #672, opened 2025-02-05
316 packet lifetime 7

cpacejo:
Two separate but related changes (in two commits for easier reviewing):

First, we're now explicit about packets being copied into egress queues, no more drop guard callback shenanigans.  (This simplifies the following step.)

Second, we made a single FastpathWorker, which is shared by all fastpaths (agent output and substrate ingress).  The main fastpath stages (agent output, substrate ingress, forward, substrate egress, agent input) are all methods of this struct, as well as `drop_and_count` (setting the stage for the next PR which will create one buffer stack per worker).

This resulted in a bunch of code moving between files again; I've noted where this happened.

### DIFF
--- a/adapter/ph/src/fastpath.rs
+++ b/adapter/ph/src/fastpath.rs
@@ -3,34 +3,30 @@
 //! General rule: no fastpath operation may block.
 //! This implies that all functions here must be non-async.
 
+use crate::adapter_tables::AltEntry;
 use crate::assembly::{Assembly, PhMode};
+use crate::classifier::{self, ClassifierResult};
 use crate::config;
 use crate::counters::CounterType;
 use crate::defs::Direction;
 use crate::km::Codec;
+use crate::km_noise::NOISE_PADLEN;
 use crate::logging::targets::DATAPATH;
 use crate::net_defs;
 use crate::packet::Packet;
-use crate::queues::TryEnqueueError;
+use crate::queues::{AdapterManager, MgmtDispatch, TryEnqueueError};
 use crate::zdp;
 use crate::zdp_ll;
 use crate::{compress, km};
 use blake3;
 use bytes::{Buf, BufMut};
+use std::sync::Arc;
 use std::time::SystemTime;
 use tracing::*;
 use zerocopy::FromBytes;
 use zpr;
+use zpr_ext::std::num::NonZeroExt;
 use zpr_ext::zerocopy::*;
-
-/// Drop a packet and count the drop with the given reason.
-pub fn drop_and_count(asm: &Assembly, pkt: Packet, reason: impl Into<CounterType>) {
-    let reason = reason.into();
-    debug!(target: DATAPATH, "dropping packet because {reason}");
-    asm.buffer_stack
-        .put_buffer(pkt.destroy().try_into().unwrap());
-    asm.counters[reason].increment();
-}
 
 /// Drop a heap-allocated packet and count the drop with the given reason.
 pub fn drop_and_count_heap(asm: &Assembly, pkt: Packet, reason: impl Into<CounterType>) {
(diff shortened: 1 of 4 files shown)


adapter/ph/src/fastpath.rs
cpacejo, line 1: All of the "new" code in this file is not new, it is just moved from elsewhere (agent_output_worker, substrate_ingress_worker, or elsewhere in this file).

Exception: the changes to eliminate drop_guard; they are separated in their own commit though.

adapter/ph/src/fastpath.rs
```diff
@@ -41,6 +36,452 @@ pub fn drop_and_count_heap(asm: &Assembly, pkt: Packet, reason: impl Into<Counte
...
+                }
+            }
+        };
+
+        if forward {
+            self.forward(pkt);
+        } else {
+            // Bind request pending; drop this packet
+            self.drop_and_count(pkt, CounterType::DroppedAwaitingBind);
+        }
```
cpacejo, line 367: Also, I had to restructure the above `match` to have part of its code in this extra if/else in order to appease the borrow checker.  No functionality changes though.

adapter/ph/src/fastpath.rs
```diff
@@ -41,6 +36,452 @@ pub fn drop_and_count_heap(asm: &Assembly, pkt: Packet, reason: impl Into<Counte
     asm.counters[reason].increment();
 }
 
+const fn adapter_next_hop_link(ingress_link_id: zpr::LinkId) -> zpr::LinkId {
+    // this optimization is checked by the static asserts below
+    // this allows us to avoid an unpredictable branch on every packet
+    (ingress_link_id % 2) + 1
```
mkolehmainen, line 42: I see the comment, but what is going on here?  Shouldn't this be a forwarding call?

adapter/ph/src/fastpath.rs
```diff
@@ -41,6 +36,452 @@ pub fn drop_and_count_heap(asm: &Assembly, pkt: Packet, reason: impl Into<Counte
...
+#[cfg(debug_assertions)]
+/// This table is used to track whether a flow ever switches from one worker
+/// to another (indicating potential for out-of-order packets) -- meaning
+/// our packet steerer isn't steering correctly.  This is used only in debug mode.
+const AGENT_PACKET_FLOW_TRACKER: std::sync::LazyLock<
+    dashmap::DashMap<(zpr::LinkId, zpr::StreamId), usize>,
+> = std::sync::LazyLock::new(|| dashmap::DashMap::new());
+
+pub struct FastpathWorker {
+    pub worker_index: usize,
```
mkolehmainen, line 57: What is `worker_idnex` indexing into?

adapter/ph/src/fastpath.rs
```diff
@@ -41,6 +36,452 @@ pub fn drop_and_count_heap(asm: &Assembly, pkt: Packet, reason: impl Into<Counte
...
+            return self.drop_and_count(pkt, CounterType::BadStructure);
+        };
+
+        let ingress_stream_id: zpr::StreamId = per_flow_hdr.stream_id.into();
+        pkt.metadata_mut().ingress_stream_id = ingress_stream_id;
+
+        // in debug builds, track which worker this agent traffic came in on
+        // ensure a given flow isn't hopping between workers (potentially
+        // resulting in out-of-order packets)
+        #[cfg(debug_assertions)]
```
mkolehmainen, line 246: I'm not sure about using debug_assertions so much.  I personally always run the debug code (ie, target/debug/foo), and I think the tests do too?  I worry that this code may introduce bugs that only show up in "non-debug" code.  A concern?

adapter/ph/src/fastpath.rs
```diff
@@ -41,6 +36,452 @@ pub fn drop_and_count_heap(asm: &Assembly, pkt: Packet, reason: impl Into<Counte
...
+                return;
+            }
+        };
+
+        match classification {
+            ClassifierResult::OK | ClassifierResult::UnclassifiedL4 => (),
+
+            ClassifierResult::FirstFragment | ClassifierResult::SubsequentFragment => {
+                // TODO: handle fragments!
+                self.drop_and_count(pkt, CounterType::InPacksDrop);
```
mkolehmainen, line 282: Just something for the future.  Maybe we should add a Fragment and NonIP type counters.

mkolehmainen (approved):
I have some comments in there, but looks good overall.

adapter/ph/src/fastpath.rs
cpacejo: This is existing code, just moved.  (This is used to exchange packets between agent and tether in an adapter.)

adapter/ph/src/fastpath.rs
cpacejo: It's just used to number fastpath workers, 0 through N-1.  We use it at least for requeueing, and also for some debug checks.

adapter/ph/src/fastpath.rs
cpacejo: This is an expensive check performed on each packet, we can't enable it in production.

(Also, this is existing code, just moved.)

adapter/ph/src/fastpath.rs
cpacejo: Near term I have a "counter reorg" planned in my mind.  Not new code here though.

adapter/ph/src/fastpath.rs
cpacejo: Added a comment describing this in my next PR.

adapter/ph/src/fastpath.rs
mkolehmainen: Ah, maybe "worker_id" would be better? I was expecting to find an array somewhere...

PR #673, opened 2025-02-05
enable a couple of visa revocation measures

### LINKED ISSUE #666: visa/CN revocation for admin
Make sure that our admin path (the HTTPS API into visa service) is working properly for revocation.

- Need to be able to revoke a visa.
- Need to be able to revoke a CN

mkolehmainen:
New work on existing visa service to support revocation by visa-id and CN. 

Also added new routines to the `vs-admin` command line tool (which exercises the visa service admin HTTPS API) so that you can perform the revocation operation from command line.


visaservice/core/pkg/snauth/credid.go
```diff
@@ -3,9 +3,10 @@ package snauth
 type CredIDType int
 
 const (
-	CredIDTypeNil         CredIDType = iota
-	CredIDTypeAuthority              // actually now just means a Key fingerprint (could be an authority key or a agent key)
-	CredIDTypeCertificate            // actually means a JTI value
+	CredIDTypeAuthority   CredIDType = iota + 1 // actually now just means a Key fingerprint (could be an authority key or a agent key)
```
svjaffer, line 6: So I looked up what iota is and what it does in go, but I'm admittedly not understanding its purpose here.  Can you explain what this change means?

visaservice/vs-admin/src/main.rs
```diff
@@ -47,6 +52,37 @@ enum SubCmd {
...
+    /// Revoke a visa by ID or an adapter's access by CN.
+    #[command()]
+    Revoke {
+        #[command(flatten)]
+        arg: RevokeArg,
+    },
+
+    /// Clear revocation state in visa service
+    #[command()]
+    ClearRevokes,
```
svjaffer, line 65: Is this intended to bring back revoked visas, clear out a pending list, or something else?

visaservice/core/pkg/snauth/credid.go
mkolehmainen: Well I used to have a Nil value there which is no longer relevant.  So I removed it.  The iota thing starts a little counter at zero and gives values to successive enums automagically.  I don't want to use the zero value so I add 1 to it so that the enum starts with 1.  So first enum is 1, next is 2 and so on.

visaservice/vs-admin/src/main.rs
mkolehmainen: This clears the revoke state memory from the visa service.  Does not apply to visa IDs.  But when you revoke a CN or other durable thing the visa service remembers it (until next policy push).  Using this clear-revokes function you can clear out that memory.  So for example, after a clear, a previously revoked CN will be allowed to connect again.

visaservice/core/pkg/snauth/credid.go
svjaffer: Oh, so this is basically an enumeration?

svjaffer (approved):

visaservice/core/pkg/snauth/credid.go
mkolehmainen: Oh right. The new type + const is the golang enum.  The iota is used to assign values at compile time.  You can do other neat things with it like ( https://cs.opensource.google/go/go/+/refs/tags/go1.23.6:src/net/interface.go;drc=05cbbf985fed823a174bf95cc78a7d44f948fdab;l=53 )

PR #677, opened 2025-02-12
make runners run on mac

### LINKED ISSUE #665: extend runner wrappers to support macos
Make sure the wrapper can start an adapter on macos.

Not sure node ph is supported on macos, but if it is, then the wrapper should work for node too.

mkolehmainen:
Ensure that the runners run on mac.

Due to our network libs in use, the mac version works a bit differently from linux: on mac you must not create the tun interface before running.  Also, annoyingly the mac library we are using for tun does not handle IPv6.

### DIFF
--- a/adapter/runners/src/env.rs
+++ b/adapter/runners/src/env.rs
@@ -81,7 +81,7 @@ pub fn configure_env(config: &Path, dry_run: bool) -> Result<(), LaunchErr> {
     );
 
     // We always drop privs, even in dry run mode.
-    sys::get_platform().drop_privledges(&run_as_user, false)?;
+    sys::get_platform().drop_privileges(&run_as_user, false)?;
 
     // The control_path parent directories must exist. This can be set in the
     // config, or there is a default.  We do this after dropping privs since the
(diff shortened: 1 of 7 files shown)


adapter/runners/src/sys/unix.rs
mkolehmainen, line 1: "unix.rs" is old code moved from linux.rs and now shared with linux and mac.

adapter/runners/src/env.rs
```diff
@@ -54,6 +42,23 @@ pub fn configure_env(config: &Path, dry_run: bool) -> Result<(), LaunchErr> {
         None => sys::get_platform().get_tun_ifname().to_string(),
     };
 
+    #[cfg(target_os = "macos")]
+    {
+        if !tun_name.is_empty() {
+            println!(
+                "warning: on macos it is reccommended to not set a tun name (found {})",
```
cpacejo: Spelling nit: "recommend"

cpacejo (dismissed):
Looks good!  Minor spelling nit in user-facing text.

adapter/runners/src/env.rs
mkolehmainen: patched

mkolehmainen:
Fixed spelling. Sorry needs approve again.

cpacejo (approved):
Perfect thanks!!

PR #680, opened 2025-02-21
use a config "rest" api internally for compiler

### LINKED ISSUE #661: ZPL compiler to access configuration through simple API
As we have discussed, we would like configuration to be accessible through a simple query API so although the data might be in a TOML file (for example), we should access it using queries so that we can ultimately split off a config service.

mkolehmainen:
As a step towards experimenting with a configuration service, the compiler now uses a simple GET(BLAH) api to talk to the configuration.  So rework code to use the new config_api.rs instead of config.rs.

Future story will add ability to run the config file as a service and point the compiler at that.


compiler/src/weaver.rs
```diff
@@ -20,21 +19,36 @@ pub struct Weaver {
 
     // Map the allow clause ID to the fabric service ID.
     allowid_to_fab_svc: HashMap<usize, String>,
-    base_path: PathBuf,
+    //base_path: PathBuf,
```
svjaffer, line 22: Is this coming back, or can it be removed?

svjaffer (approved):

compiler/src/weaver.rs
mkolehmainen: I'll get that in next one

PR #688, opened 2025-02-26
with multi procs, do not unset node flag

### LINKED ISSUE #687: bug: vs may clear node flag when should be set
When processing multiple connection procs, may clear the node flag when it is supposed to be set.  This is a bug, node flag should stay set.

mkolehmainen:
Ensure we do not un-set the node flag.


cpacejo (approved):
Not sure I understand the bug but the new code looks fine!

PR #698, opened 2025-03-06
tighten up zpl processing

### LINKED ISSUE #691: adapter can't connect unless a service is defined specifically for it
```
define adapter1 as an endpoint with cn
define Ping as a service with cn:adapter1
```
works, but
```
define adapter1 as an endpoint with cn:adapter1
define Ping as a service with cn
```
results in the adapter being unable to connect to the node (it is rejected by the visa service).

### LINKED ISSUE #692: compiler doesn't emit warnings for lines which do nothing
e.g.:
```
allow endpoint to access Ping service
```
does not result in an entry in the traffic index.  But the compiler doesn't complain.

mkolehmainen:
Fixes some oddities in the compiler.  Including #691 and #691.

This has a small patch to the visa service too which permits attribute tuples like (cn, has, "") to be interpreted to mean that the presence of the attribute alone is enough to match.  This is how an expression like "define Foo as service with cn" is encoded.

There also more (but still not enough) tests.


cpacejo (approved):
Looks good!  I'll make sure to test this out as I'm doing performance testing.

PR #702, opened 2025-03-10
Visa table first pass

svjaffer:
What's in this PR:
- Track active visas and link them to active forwarding entries
- Adding a forwarding entry requires a visa
- Removing a visa removes all forwarding entries associated with it
- Keep a timeout queue of visa expirations (not serviced in this PR)
- Basic unit-testing around this functionality

What's NOT in this PR:
- Handling VS revocation messages
- Timeout processing
- Proper Visa struct (currently just plopping the thrift-generated struct in there)
- Session timeouts
- Looking up whether we already have a visa that matches a new session
- Probably some more stuff I forgot to mention


adapter/ph/src/assembly.rs
```diff
@@ -279,8 +286,32 @@ impl Assembly {
...
+            .unwrap()
+            .link_forwarding_entry(
+                visa_id,
+                zpr::ForwardingEntry(ingress_link_id.get(), ingress_tether_id),
+            )
+            .is_err()
+        {
+            // Visa was either never granted or has already been removed
+            // Route is no longer valid
+            ingress_peer_state.pft.remove(ingress_tether_id);
```
cpacejo, line 301: Rather than adding and removing if in error, would it make sense to first grab the visa table lock, then lookup the visa, and hold the lock while we add the peer state?

adapter/ph/src/visa_table.rs
```diff
@@ -0,0 +1,291 @@
...
+/// Struct that holds an instance of a visa local to a Node
+pub struct Visa {
+    pub visa: Option<vsapi::Visa>,
+    streams: Vec<ForwardingEntry>,
+}
+
+/// Struct for the visa timeout queue
+pub struct VisaTimeout {
+    pub id: VisaId,
+    pub expiration: DateTime<Utc>,
```
cpacejo, line 36: Reminder to revisit & try using `SystemTime` here, to avoid storing timezone issue with every expiration.

adapter/ph/src/visa_table.rs
```diff
@@ -0,0 +1,291 @@
...
+}
+
+// Visas should sort with soonest expiration first
+impl Ord for VisaTimeout {
```
cpacejo, line 40: You can `#[derive]` these if you order the fields in `VisaTimeout` lexicographically.

adapter/ph/src/visa_table.rs
```diff
@@ -0,0 +1,291 @@
...
+        self.id == other.id && self.expiration == other.expiration
+    }
+}
+
+impl Eq for VisaTimeout {}
+
+impl Visa {
+    /// Remove all forwarding entries associated with this visa
+    pub fn remove_forwarding_entries(&mut self, asm: &Arc<Assembly>) {
+        while let Some(entry) = self.streams.pop() {
```
cpacejo, line 66: This is fine, another way you can write this (which processes streams in the opposite order) is `self.streams.drain(..).map(|entry| asm.remove_route(entry));`.

cpacejo (approved):
Looks good!

PR #706, opened 2025-03-14
compiler parses periods and other improvements

### LINKED ISSUE #686: compiler - statements delimited with PERIOD
RFC15.2.12 says that statements are delimited by PERIOD.

mkolehmainen:
Period was always allowed, but it was ignored. Now periods are still optional, but if you do happen to use them incorrectly you get an error.

* New `--Werror` flag to halt on any warnings.
* Gratuitous, fancy colored terminal output!

**BREAKING CHANGE** -> Also now correctly enforces the characters allowed in unquoted names and strings.  This means that if you have a period in a tuple value you must quote it.  Eg,

`define FooBoo as service with cn:'hah.ho.hee'`.

_Note the quotes around "hah.ho.hee".  ZPL accepts forward or backward SINGLE quotes (and they don't have to match or anything)_

### DIFF
--- a/compiler/README.md
+++ b/compiler/README.md
@@ -1,19 +1,19 @@
 # ZPL Compiler
 
-Can compile simple (new) ZPL into binary policies that the prototype visa service 
+Can compile simple (new) ZPL into binary policies that the prototype visa service
 can process.
 
 ## Example usage
 
 ```bash
-./zpc -k path/to/rsa-key.pem path/to/policy.zpl 
+./zpc -k path/to/rsa-key.pem path/to/policy.zpl
 ```
 
 - That RSA key in the invocation is used to sign the binary policy so
   must match the one that the visa service is configured with.
 - By default the _configuration_ for the ZPL policy will be found in a
   file with the same name as the ZPL file but with the `zplc` extension.
-  If you want to loca configuration from somewhre else, use 
+  If you want to loca configuration from somewhre else, use
   `-c path/to/config.zplc` argument.
 - Help is available via `zpc -h`
 
@@ -25,13 +25,13 @@ Note that this is just enough grammer to meet current compiler needs.
 
 
 
-
-
 ```
 
 ignored-tokens  : 'a', 'an'
 
-statement       : ( allow-statement | define-statement ) + '\n'
+statement       : ( allow-statement | define-statement ) + <eos>
+
+eos             : EOF | '\n' | '.'
 
 allow-statement : 'allow' + <device-clause> + 'with' + <user-clause> + 'to access' + <service-clause>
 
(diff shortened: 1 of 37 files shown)


compiler/README.md
```diff
@@ -60,11 +60,11 @@ attribute       : tuple | attr-name
 
 tuple           : attr-name + ':' + attr-value
 
-attr-name       : string | ns-name + '.' + name
+attr-name       : string | name
 
 attr-value      : string | integer
 
-ns-name         : string | ns-name + '.' + string
+name            : string | [ <name> + '.' + <name> ]
```
cpacejo, line 67: Two comments –

Doesn't this deviate from the Danny/Guy/Oracle spec?

And, can we avoid this by checking whether there's whitespace after a period?  (i.e., periods are only tokenized as `PERIOD` if they are followed by whitespace.)

compiler/src/context.rs
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    /// Show an informational message.
+    pub fn info(&self, msg: &str) {
+        println!("{} {}", "ℤ".cyan(), msg);
+    }
+
+    /// Show a warning.  May raise an error if warnings are promoted to errors.
+    pub fn warn(&self, msg: &str) -> Result<(), CompilationError> {
+        println!("{}: {}", "warning".yellow().bold(), msg.bold());
```
cpacejo, line 25: It seems like some things are stdout and others are stderr (elsewhere in this review).  Should we use just one or the other, so redirections work as expected?

mkolehmainen: Yes, good point. I'll put everything on stdout for now.

compiler/README.md
mkolehmainen: I believe this is exactly the spec (RFC 15).  But I'll hit you up offline to get what you are talking about here.

compiler/src/compilation.rs
```diff
@@ -105,14 +108,15 @@ impl Compilation {
...
         })?;
-        println!("ℤ wrote {}", &file.display());
+        ctx.info(&format!("wrote {}", &file.display()));
         Ok(())
     }
 
     /// Create the container struct and optionally sign the policy with the private key.
     fn contain_policy(
-        &self,
+        &mut self,
```
cpacejo: Same q here -- why the change to `&mut`?  (I don't see any mutating accesses to `self`.)

compiler/src/compilation.rs
```diff
@@ -35,56 +37,56 @@ impl Compilation {
     }
 
     /// Create a policy from the ZPL source and configuration.
-    pub fn compile(&self) -> Result<(), CompilationError> {
+    pub fn compile(&mut self) -> Result<(), CompilationError> {
```
cpacejo: Why the change to `&mut`?

cpacejo (dismissed):
Looks good!

mkolehmainen:
All those muts were part of an abandoned change.  Removed!

cpacejo (approved):

PR #709, opened 2025-03-20
Node can process VSS messages

### LINKED ISSUE #656: Node can handle visa revocation
The forwarding table now has entries populated by visas.  When those visas get revoked, the table entries must be removed.

svjaffer:
- Can revoke visas
- Can install unrequested visas (but no forwarding)
- Changed lock on the Visa Table to a tokio RwLock
- Moved the VS communication stuff out of the vs_worker file into the new visa_mgmt file
- Renamed rpc_commands.rs to admin.rs (should have been in a different PR, but accidentally captured it here)
- Updated integration test to include revocation


mkolehmainen (dismissed):
Looks good to me! Ship it!

adapter/ph/src/admin.rs
cpacejo, line 1: `admin_worker` maybe?  (All "thread-main" modules I've been naming `_worker` so they're easy to identify.)

adapter/ph/src/visa_table.rs
```diff
@@ -2,24 +2,23 @@
...
 use zpr::{ForwardingEntry, VisaId};
 
 #[derive(Debug, Error)]
-pub enum VisaTableError<'a> {
+pub enum VisaTableError {
     #[error("Visa {0} Not Found")]
     NotFound(VisaId),
     #[error("Failed to parse visa field {0}")]
-    ParseFailure(&'a str),
+    ParseError(String),
```
cpacejo: `&'static str` would be better here… at least, all the strings this is used with look like static strings.  (vs. `String` which manages dynamically allocated memory.)

cpacejo (dismissed):
Two nits, otherwise looks great!

cpacejo (approved):

PR #712, opened 2025-03-21
batch_io module

cpacejo:
Just the batch I/O primitives right now.  Implementations for both `io_uring` and normal non-batched POSIX calls.  I didn't put this in the `sys` directory because it doesn't fall cleanly into Linux vs. macOS.  It's more `io_uring` (newer Linux) vs. POSIX (older Linux / any macOS).

Actually using these is a forthcoming review.

Also changed one setting in `Cargo.toml` which was making debugging a challenge.

### DIFF
--- a/adapter/ph/Cargo.lock
+++ b/adapter/ph/Cargo.lock
@@ -936,6 +936,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8bb03732005da905c88227371639bf1ad885cc712789c011c31c5fb3ab3ccf02"
 
 [[package]]
+name = "io-uring"
+version = "0.7.4"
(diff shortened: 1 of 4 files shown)


mkolehmainen (approved):
This whole io_uring thing could use more comments. BUT.... is this going to break PH on macos?

cpacejo:
I'll add better comments in a forthcoming PR (I've got a few more in this series coming), checking this one in for now to unblock the others.

macOS will use the `posix_unbatched` implementation, which just uses plain `write`, `send`, etc.  That said, I have no way to test on macOS :/

cpacejo:
macOS build is apparently already broken!  Pushing as-is, will fix existing (and any potential new) macOS breakage in a separate PR.

PR #713, opened 2025-03-21
PH: batch processing

cpacejo:
Restructure fastpath to use batch processing primitives.

Main design changes are (a) each fastpath worker now has 2 internal output queues (one for the TUN, one for the substrate) which are drained to the kernel at the end of each batch, and (b) that outbound packets are sent on the fastpath worker's own queue, rather than choosing a queue based on hash.  There's no benefit I could see to doing the latter.

There's still some cleanup to do re: how fastpath workers get access to the outbound sockets, and how mgmt accesses outbound as well.  Changes for those will be in future PRs.


adapter/ph/src/substrate_ingress_worker.rs
```diff
@@ -45,41 +50,65 @@ fn substrate_ingress_main(mut worker: FastpathWorker, socket: &UdpSocket) {
...
 
-                    // FIXME: do something with this later...
-                    ErrorKind::ConnectionRefused => {
-                        worker.buffers.push(pkt.destroy());
-                        continue;
-                    }
+        // process packets
+        for (pkt, result) in pkts.into_iter().zip(results) {
+            let mut sender = match result {
+                Ok((size, _sender)) if size > pkt.remaining() => {
```
svjaffer, line 76: This code would be clearer in its intent if you were using `pkt.len()` instead of `pkt.remaining()`.

cpacejo: Opened a PR to implement (and use!) Packet::len().

svjaffer (approved):

PR #724, opened 2025-03-26
restructure two_way_queue to allow shared return queue

### LINKED ISSUE #717: break up up two-way queue; make return queue explicit

cpacejo:
Prerequisite to unifying the fastpath workers.  Closes #717.

### DIFF
--- a/adapter/ph/src/queues.rs
+++ b/adapter/ph/src/queues.rs
@@ -244,16 +244,11 @@ impl two_way_queue::TwoWayReturnable<MgmtDispatchMessage> for PacketBuffer {
     }
 }
 
-#[derive(Clone)]
 pub struct MgmtDispatch {
     sender: two_way_queue::Sender<MgmtDispatchMessage, PacketBuffer>,
 }
 
 impl MgmtDispatch {
-    pub fn new(sender: two_way_queue::Sender<MgmtDispatchMessage, PacketBuffer>) -> Self {
-        Self { sender }
-    }
-
     pub fn try_dispatch_mgmt_packet_with_link(
         &mut self,
         packet: Packet,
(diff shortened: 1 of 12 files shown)


adapter/ph/src/queues.rs
```diff
@@ -295,17 +290,19 @@ impl MgmtDispatch {
             }
         }
     }
+}
 
-    pub fn recv_return_buffers(&mut self, returns: &mut Vec<PacketBuffer>, limit: usize) -> usize {
-        self.sender.blocking_recv_many_returns(returns, limit)
+pub struct MgmtDispatchFactory(two_way_queue::SenderFactory<MgmtDispatchMessage, PacketBuffer>);
```
mkolehmainen, line 296: It would be great to have some comments here describing what this factory is building and why.  I see, it makes MgmtDispatch things, but what's that and how many does the PH need?  Same questions for that AdapterManagerFactory.  I'm thinking there is 1 AdapterManager for each adapter connected?? Not sure.  At first I thought this was more of an OO "factory" but it looks like that's not true... It always creates the same things, right?

adapter/ph/src/two_way_queue.rs
```diff
@@ -85,9 +80,9 @@ impl<T, U> Sender<T, U> {
     ///
     /// Panics if called when no items are outstanding.
     pub fn blocking_recv_many_returns(&mut self, returns: &mut Vec<U>, limit: usize) -> usize {
-        assert!(self.outstanding > 0 || limit == 0);
+        assert!(self.outstanding.load() > 0 || limit == 0);
```
mkolehmainen, line 97: About you say this returns when limit is 0.  Looks like it panics?

mkolehmainen (dismissed):
Looks good. Just some comment requests!

adapter/ph/src/two_way_queue.rs
mkolehmainen: Oh, whoops I read that wrong.

adapter/ph/src/queues.rs
cpacejo: Oops yeah this needs comments, lemme do that…

adapter/ph/src/two_way_queue.rs
```diff
@@ -14,13 +14,29 @@
...
+//! "services") creates a single `ReturnQueue` to receive returned items.
+//! Then, for each client-server pairing, the client creates a `Sender` which
+//! sends items to the given `Receiver`, to be returned on the client's own
+//! `ReturnQueue`.
+//!
+//! The two main entry points to create all these objects are
+//! `ReturnQueue::new()`, and `two_way_queue()`.  The former creates a
+//! single `ReturnQueue`.  The latter creates a pair of a `SenderFactory`
+//! and a `Receiver`.  The `SenderFactory` is then used to create `Sender`
+//! instances linked to the `Receiver` and a supplied `ReturnQueue`.
```
mkolehmainen, line 30: thanks for the comments!

mkolehmainen (approved):
looks good to me!

PR #727, opened 2025-03-26
Add some missing error debugging to node's visa management path

svjaffer:


mkolehmainen (commented):
Great let's try it!

mkolehmainen (approved):
looks good

PR #728, opened 2025-03-26
The node's lookup for egress link now looks at the local agent too

svjaffer:


cpacejo (approved):
Nice thanks for doing this!

PR #729, opened 2025-03-26
graceful shutdown on SIGINT

cpacejo:


adapter/ph/src/signal_worker.rs
```diff
@@ -2,29 +2,59 @@
 
 use crate::assembly::Assembly;
 use crate::counters::*;
+use crate::logging::targets::STARTUP;
 use std::sync::Arc;
 use tokio::signal::unix::{signal, SignalKind};
-
-fn emit_counts(counters: &Counters, uptime: std::time::Duration) {
```
cpacejo, line 8: Just moved below.

mkolehmainen (approved):
Looks good!

PR #730, opened 2025-03-26
After 3 missed keep-alives, link will shut itself down

svjaffer:


mkolehmainen (approved):
It's 4 misses right? Maybe that should be a constant ?

cpacejo (approved):

PR #734, opened 2025-03-26
Nodes will now remove disconnected adapters from their peer tables

svjaffer:


adapter/ph/src/link_state.rs
```diff
@@ -674,6 +675,32 @@ impl LinkStateWrapper {
         Ok(())
     }
 
+    /// Tear down link state
```
svjaffer, line 678: This was moved to be near the code it was in sequence with.

The only functional change to this function was adding the `visa_mgmt::agent_disconnect`

adapter/ph/src/config.rs
```diff
@@ -26,6 +26,9 @@ const DEFAULT_WORKER_CONCURRENCY: usize = 4;
 #[cfg(target_os = "macos")]
 const DEFAULT_WORKER_CONCURRENCY: usize = 1;
 
+pub const DEFAULT_KEEP_ALIVE_PERIOD: u64 = 3;
```
cpacejo: Possible to use `std::time::Duration` here, or suffix the name with `_SECS`?

cpacejo (dismissed):

adapter/ph/src/config.rs
mkolehmainen: +1

mkolehmainen (dismissed):
Yes! Looking forward to trying this later today.

mkolehmainen (approved):
LGTM

PR #736, opened 2025-03-27
delete the compiler directory, actions and refs

### LINKED ISSUE #668: Move compiler to its own repo
Move the zpl compiler into its own repository.

Also create a versioned artifact for the compiler build (or is this a separate story?)

mkolehmainen:
This removes the compiler source over to its own repo.

- Remove source
- Remove the github action
- Tweak wording the integ test 

Will need to update links in the ref-impl project too.


integration-test/pregen/Makefile
```diff
@@ -10,7 +10,7 @@ info:
 	@echo make clean - removes all the .bin policy files
 	@echo make all   - recompiles out of date policy files
 	@echo
-	@echo Requires that compiles is built in debug target.
+	@echo Requires that \'zpc\' the ZPL compiler is in your path.
```
svjaffer, line 13: I haven't gone and checked and maybe you can save me the time of doing so.  Does building the compiler put it in your path or, alternately, is there a script that puts it in your path based on where your compiler repo is?

svjaffer (dismissed):

integration-test/pregen/Makefile
mkolehmainen: No you would have to put it in your own path manually.  Maybe eventually we could have a deb or something produced in a "release" for zpr-compiler.  That variable thing is handy because you can just do `make ZPC=/path/to/my/compiler/zpc all` and get some satisfaction.

cpacejo (dismissed):
Nice!

PR #738, opened 2025-03-27
Unify fastpath workers

cpacejo:
Per our earlier discussion, now there is just one type of fastpath worker.  This means we have an equal number of TUN queues as substrate queues.  This also means we eliminate contention on these queues.

Next (and final?) step after this is to route mgmt substrate egress through one of these workers.  This will fully eliminate the links of the datapath to Tokio, and also allow each datapath worker to fully own its own TUN and substrate queue.


adapter/ph/src/fastpath_worker.rs
```diff
@@ -104,20 +114,72 @@ fn agent_output_main(mut worker: FastpathWorker, tun: &ZprTun, requeue_outq: &Un
...
 
-        if poll_fds[PollSlot::Tun]
+        if poll_fds[PollSlot::Substrate]
             .revents()
             .unwrap()
             .contains(poll::PollFlags::POLLIN)
         {
-            let nbufs = std::cmp::min(worker.buffers.len(), worker.config.batch_size);
+            // read from socket
+            let mut pkts = Vec::new(); // TODO: recycle
```
cpacejo, line 123: These TODOs will be addressed in the forthcoming PR which adds the mgmt->substrate queue.

adapter/ph/src/fastpath_worker.rs
```diff
@@ -104,20 +114,72 @@ fn agent_output_main(mut worker: FastpathWorker, tun: &ZprTun, requeue_outq: &Un
...
         }
 
-        if poll_fds[PollSlot::Tun]
+        if poll_fds[PollSlot::Substrate]
             .revents()
             .unwrap()
             .contains(poll::PollFlags::POLLIN)
         {
-            let nbufs = std::cmp::min(worker.buffers.len(), worker.config.batch_size);
+            // read from socket
```
svjaffer, line 122: Maybe in your follow-on PR, break these out into separate functions?

svjaffer (approved):

adapter/ph/src/fastpath_worker.rs
cpacejo: Oh, I can do that.  I'll have one more to add then anyway.

PR #756, opened 2025-04-01
Mk/envvar

### LINKED ISSUE #720: PH ability to load noise private key from environment variable
To help with keeping secrets, add ability to load the noise private key from an environment variable and/or accept it as a literal base64 string on the command line.

mkolehmainen:
Add ability to set the noise private key in a base64 encoded string held in an environment variable (`NOISE_PRIVATE_KEY`).

Maintain existing option of setting the key as a path to a PEM file.

Note that the clap parser does this when you just slap an `env` hint on an arg.  All I had to do was allow config to hold either a path or the base64 data internally.


adapter/ph/src/sys/macos/zprtun.rs
mkolehmainen, line 1: Removed these unused imports which were causing warnings on mac.

adapter/ph/src/main_args.rs
```diff
@@ -61,54 +44,58 @@ impl ArgError for str {
...
     /// Certificate including the noise public key, signed by the authority
     #[arg(long, value_name = "PATH")]
-    certificate_file: Option<String>, // noise public key signed by authority
+    pub certificate_file: Option<String>, // noise public key signed by authority
 
     /// Path to the noise private key file (PEM format)
     #[arg(long, short = 'k', value_name = "PATH")]
-    private_key_file: Option<String>, // noise private key
+    pub private_key_file: Option<String>, // noise private key
+
```
mkolehmainen, line 75: Right below here is the magic invocation to clap to read noise_private_key from environment variable (or you can pass as arg as usual).

adapter/ph/src/main_args.rs
mkolehmainen, line 1: Moved most of this code out of main_args and into main_argparse and config.

adapter/ph/src/main_argparse.rs
mkolehmainen, line 1: This code copied out of main_args.  Minor changes to tests to use the new way of accessing private key, and a new test at the end.

adapter/ph/src/config.rs
mkolehmainen, line 1: Copied over code from main_args that has to do with implementing the Config.  Modified config so it can hold a path to a secret key, or the key data itself.  Access to the key data is through a new function.

cpacejo (approved):
Looks good!

PR #761, opened 2025-04-03
Add ability for links to restart

### LINKED ISSUE #603: Link initiators should restart closed link when appropriate
If something goes wrong in bringing a tether or link up, it can shut itself down.  Right now, human intervention can restart the link, but it should restart automatically, but with enough delay to not spam if things are continuously hitting a bad state.

svjaffer:
Based on feedback while working on the last demo, I've made some changes to the link state machine.  You should be able to see updated link state diagrams in the link_state file.  Some of those changes were made in past PRs and the diagrams have just been updated to be more correct, but a few additional changes have been made around link shutdown to clean up the changes made during the demo work.

Functionally, adapters will now seek to restart their links after they've been shut down.

Additionally, graceful shutdown now waits for agent disconnects to be sent to the visa service.  There may be issues if the visa service is not connected, but fixing that will require some mucking around with libnode.

Things get pretty spammy when adapters get disconnected from their node and are trying to reconnect, but that's going to need to be a change to the km subsystem and I don't want to bloat this PR.


adapter/ph/src/peer_table.rs
```diff
@@ -113,6 +113,7 @@ pub type PeerTableEntryGuard<'a> = RcuCslabEntryGuard<'a, PeerState>;
 #[derive(Debug)]
 pub enum PeerInsertError {
     TableFull,
+    Other,
```
mkolehmainen: Personally, don't love OTHER here without it taking a message or some other error type.  Maybe it could take a LinkStateError ?  You know first thing someone will do when they see the OTHER is grep the code to see where it is thrown and then add a debug line or something.

mkolehmainen (dismissed):
Looks good. Excited to try this out.  One comment in there you can take or leave.

adapter/ph/src/peer_table.rs
svjaffer: Totally.  It's dead code that was just left over so I removed it.

svjaffer: Not sure what I was thinking on Friday.  Undid that change and fixed it properly (the way you had suggested).

adapter/ph/src/link_state.rs
```diff
@@ -262,13 +254,29 @@ impl LinkStateWrapper {
         locked_fsm.agent_addresses.clone()
     }
 
+    fn deregister_agent_addresses(&self, asm: &Arc<Assembly>) -> tokio::task::JoinSet<()> {
```
cpacejo, line 257: Rather than passing a `JoinSet` around, I wonder can we express this in the signature just using `async`?  E.g. in this function, `join_set.join_all().await` at the end.  Then in `clean_up_link_state()`, use `join!()` to wait on both the call to this function, and on the async block in that function (and `clean_up_link_state()` would then be `async`).  Then when calling `clean_up_link_state()`, you can either wait on it, or spawn & detach it to not wait on it.  I *think* this should all work and be a straightforward change.

svjaffer: Hm... I really don't want `clean_up_link_state()` itself to be async, but let me see if I can play around with it and rearrange things such that it cleans this up.

adapter/ph/src/link_state.rs
```diff
@@ -736,13 +715,44 @@ impl LinkStateWrapper {
...
+    /// Set a timer to attempt a link restart after a holddown period
+    fn setup_restart(&self, asm: &Arc<Assembly>) {
+        let link_id = self.id;
+        let task_asm = asm.clone();
+        tokio::task::spawn_local(async move {
+            tokio::time::sleep(config::DEFAULT_LINK_RESTART_HOLDDOWN).await;
+            info!(target: LINK_STATE, "Attempting to restart link {link_id}");
+            let _ = task_asm.process_link_state_event(link_id, LinkEvent::Start);
+        });
+    }
```
cpacejo, line 727: Can it be possible that we end up with several of these timers all running at once?  Or that we exit CLOSING due to another reason (e.g. error) and then have this timer still running?

(I suspect the answer is "no, the state machine ensures we only start the timer on entering CLOSING, and the timer going off is the only way out" but I'm not entirely sure.)

adapter/ph/src/link_state.rs
```diff
@@ -736,13 +715,44 @@ impl LinkStateWrapper {
...
+
     /// Handle a terminate response packet
     fn process_terminate_response(&self, asm: &Arc<Assembly>) -> Result<(), LinkStateError> {
         let link_id = self.id;
         info!(target: LINK_STATE,"Received terminate response for link {link_id}");
         let state = self.locked_fsm.lock().unwrap().state;
         match state {
-            LinkState::Closing | LinkState::Resetting => Ok(self.clean_up_link_state(asm)),
+            LinkState::Closing => {
+                self.clean_up_link_state(asm).detach_all();
```
cpacejo, line 753: Here is it possible the `setup_restart()` timeout races this?  If so, what are the consequences of one or the other losing the race?

adapter/ph/src/link_state.rs
```diff
@@ -759,60 +769,12 @@ impl LinkStateWrapper {
...
-            );
-            locked_fsm.set_state(LinkState::Resetting);
-            drop(locked_fsm);
-            let _ = send_terminate_indication(&asm, link_id, TerminateReason::Reset).await;
-        }
+        self.locked_fsm
+            .lock()
+            .unwrap()
+            .set_state(LinkState::Closing);
+        self.clean_up_link_state(asm).detach_all();
```
cpacejo, line 776: Same question (and possibly same answer?) here re: restart timeout race.

cpacejo (commented):
Nice love me some deleted code!

I just have some questions re: a couple races but otherwise looks good!

adapter/ph/src/link_state.rs
svjaffer: It should not be possible as the code currently stands.  Only the link initiator sets this up (right now, just the adapter) and it will only do it on this close and there currently isn't another way for that link to start back up (since I disabled the command line option), I don't see a way that would happen.

And if it did, getting the start called multiple times on the same link would result in subsequent start calls doing nothing since the start routine will only run if the link is in the initial state.

If any of that is not the case, it's a bug haha

adapter/ph/src/link_state.rs
svjaffer: Nope, because `setup_restart()` is called by `complete_close()`, which is itself spawned by `clean_up_link_state()`.

adapter/ph/src/link_state.rs
svjaffer: Same answer as above

cpacejo (approved):

PR #762, opened 2025-04-07
Quiet warnings for expected SA 0 KM packets

### LINKED ISSUE #693: don't warn "insecure" for necessary packets to set up link

svjaffer:


cpacejo (approved):
Thanks!  I almost wonder whether we should drop the messages entirely.  Mathias's hesitation in doing that was that we want to ensure we're never running in a mode without an SA when we "ought" to be.  I think this change as-is is good though!

PR #763, opened 2025-04-07
packet priority flag

cpacejo:
Adds a notion of "priority" substrate packets to the datapath.  When egressing such a packet, if there is no room in the kernel queue, instead of dropping it, we hang onto it and move it to the front of our own queue.

Will be used by mgmt substrate egress in upcoming review.

### DIFF
--- a/adapter/ph/src/fastpath.rs
+++ b/adapter/ph/src/fastpath.rs
@@ -14,7 +14,7 @@ use crate::km::Codec;
 use crate::km_noise::NOISE_PADLEN;
 use crate::logging::targets::DATAPATH;
 use crate::net_defs;
-use crate::packet::{Packet, PacketBuffer};
+use crate::packet::{self, Packet, PacketBuffer};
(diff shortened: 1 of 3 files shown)


adapter/ph/src/fastpath.rs
```diff
@@ -598,22 +598,42 @@ impl FastpathWorker {
...
+                    Err(err) if err.kind() == ErrorKind::WouldBlock => (),
+                    Err(err) => panic!("unrecoverable I/O error: {err}"),
+                }
+            }
+
+            // Packet was not sent.
+
+            if self.substrate_egress_q[i].0.metadata().flags & packet::flags::PRIORITY != 0 {
+                // This was a priority packet.  Move it to the front of the queue.
+                self.substrate_egress_q.swap(i, retained);
```
svjaffer, line 632: This could use a slightly more informative comment about how we're using the retained count for dual-purposes of bookkeeping and tracking the next index in the substrate_egress_q.

svjaffer (dismissed):

svjaffer (approved):

PR #764, opened 2025-04-07
integrate two-way queue returns with poll

cpacejo:
In support of a forthcoming change in the datapath, we can no longer simply block when we're out of packet buffers; we need to be able to poll on them.  This means we need a file descriptor which signals readiness when returned packet buffers are available.

This PR (1) adds a `Notify` struct, which provides a means of FD-based notifications; (2) uses `Notify` in `two_way_queue` to provide FD-based notifications of returned items; and (3) uses this mechanism in the datapath to integrate buffer return processing into `poll`.

### DIFF
--- a/adapter/ph/src/sys/posix/notify.rs
+++ b/adapter/ph/src/sys/posix/notify.rs
@@ -1,0 +1,213 @@
+//! Notification of events via a file descriptor.
+//!
+//! Useful to signal events to a thread which is using `poll(2)`.
+
+use nix::{errno, fcntl, ioctl_read_bad, libc, poll, unistd};
+use std::io::Result;
+use std::mem::MaybeUninit;
+use std::os::fd::*;
+
+ioctl_read_bad!(ioctl_fionread, libc::FIONREAD, libc::c_int);
+
+/// A synchronization object used to post notifications.
+///
+/// A `Notify` object either has an notification, or does not.
+/// Posting a notification causes the `Notify` to have a notification.
+/// Consuming the notification causes the `Notify` to have no notification.
+///
+/// The primary benefit of a `Notify` (over e.g. an `AtomicBool`) is that
+/// the presence of notifications can be monitored via a file descriptor.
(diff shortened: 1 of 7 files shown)


adapter/ph/src/sys/posix/notify.rs
```diff
@@ -0,0 +1,214 @@
...
+        // posting notifications faster than we can retrieve them.
+        // So, use `ioctl(FIONREAD)` to see how many are still outstanding,
+        // and read at least that many, then return.  It's possible also
+        // we find fewer notifications because someone else raced us;
+        // that's OK.
+
+        let mut remaining = fionread(&self.reader).unwrap() as isize;
+
+        while remaining > 0 {
+            match unistd::read(self.reader.as_raw_fd(), &mut buf) {
```
mkolehmainen, line 108: Shouldn't you call read with a size value (`remaining`) here?

mkolehmainen (dismissed):
Looks good.  One comment about the use of the remaining counter which I'm sure is just my confusion.

adapter/ph/src/sys/posix/notify.rs
cpacejo: Surprisingly no!  (But good question!)  Though that would also be correct, it's better that we read as many as are available before returning, since the caller will handle "everything" which is now ready, and any leftover notifications on our return will just wake the caller up again later after it's already handled the "things" it was notified about.  (For this same reason, it's correct _not_ to have this extra loop at all, but better that we do to avoid needless extra wakeups.)

Synchronization-wise, it's OK that we've already read some and are now reading more – so long as we stop reading before returning to the caller.

The only reason for this check at all is simply to ensure that, if we're racing a chatty notifier, we don't end up in a loop constantly reading notifications and never return.  It sets a finite "countdown", which is nonetheless large enough that if we _aren't_ racing a notifier, we will still consume everything in the channel.

mkolehmainen (approved):
Looks good!

Issue #189, opened 2024-07-12
fix TUN issue with multiqueue


cpacejo:
The `tokio-tun` crate has a bug that makes it unable to open a multiqueue TUN with a single queue.  This means if a TUN is configured as multiqueue, but the PH is configured as a single queue, it can't attach to the TUN.

PR #766, opened 2025-04-07
Remove tokio from zprtun

cpacejo:
We no longer need tokio support in zprtun, since we are now using dedicated OS threads, `poll`, and `batch_io`.

### DIFF
--- a/adapter/ph/Cargo.lock
+++ b/adapter/ph/Cargo.lock
@@ -584,6 +584,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "either"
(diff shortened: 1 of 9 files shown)


svjaffer:
Out of curiosity, does this resolve #189?

cpacejo:
> Out of curiosity, does this resolve #189?

Ah no sadly, we're still using `tokio-tun` to configure/open the TUN.  (We just then ditch it and use only the FD for I/O.)

mkolehmainen (approved):
Looks good!

PR #767, opened 2025-04-07
Create dedicated channel for mgmt to send substrate packets via datapath

cpacejo:
The final step to getting Tokio out of the datapath!

This creates a single new Unix-domain datagram socket, via which mgmt can send packets to the datapath (worker 0) to be placed on the substrate.  This means all ZDP processing now happens in the datapath threads.  Also, we're no longer sharing a UDP socket with mgmt (which was causing Tokio to spin constantly when the datapath was under load due to polling notifications).  Finally, each datapath worker now owns its own dedicated socket.


adapter/ph/src/fastpath.rs
```diff
@@ -94,7 +91,6 @@ impl FastpathWorker {
         let mgmt_dispatch = asm.mgmt_dispatch_factory.make(&return_q);
 
         let agent_input_tun = asm.agent_input.tuns[worker_index].clone(); // TEMP HACK
-        let substrate_egress_socket = asm.substrate_egress.sockets[worker_index].clone(); // TEMP HACK
```
svjaffer, line 101: Love getting rid of those TEMP HACKs!

svjaffer (approved):

PR #779, opened 2025-04-09
explicitly set UnixDatagram buffer size; fixes issue on macOS

cpacejo:
Apparently I had a TODO comment for this very issue…


mkolehmainen (approved):
Can't wait!

PR #780, opened 2025-04-09
bare bones TUN interface for macos

### LINKED ISSUE #768: MAC - new tun code to work with IPv6
The current mac tun library we are using does not support IPv6, and from experimentation it does not appear that we can create tun interface outside of the program (in order to pre-configure it).  So we need ability to create the TUN and set an IPv6 (or IPv4) IP addr.

mkolehmainen:
The rust-tun package we were using was not able to handle IPv6 so here is our own tun interface that does.  It also does far less than rust-tun, but should do just enough to cover what we need.

Also includes minor refactoring for linux tun_pi.

### DIFF
--- a/adapter/ph/src/sys/macos/tun.rs
+++ b/adapter/ph/src/sys/macos/tun.rs
@@ -1,74 +1,115 @@
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
-use std::os::unix::io::{RawFd, AsRawFd};
+use std::os::fd::AsFd;
+use std::os::unix::io::{AsRawFd, OwnedFd, FromRawFd, BorrowedFd};
 use std::mem;
 use std::ptr;
 use std::ffi::CStr;
 use nix::{ioctl_readwrite, ioctl_write_ptr};
 
+use tracing::*;
 use thiserror::Error;
 
-use crate::zprtun::DEFAULT_TUN_MTU;
+use crate::zprtun::{DEFAULT_TUN_MTU, ZPRNET_PREFIX_LEN};
 
 use libc::{
-    self, IFNAMSIZ, PF_SYSTEM, SOCK_DGRAM, SYSPROTO_CONTROL, AF_SYSTEM, AF_SYS_CONTROL, AF_INET,
-    UTUN_OPT_IFNAME, c_uint, c_char, socklen_t, sockaddr, c_void, ifreq,
-    sockaddr_in, sockaddr_in6,
+    self, IFNAMSIZ, PF_SYSTEM, SOCK_DGRAM, SYSPROTO_CONTROL,
+    AF_SYSTEM, AF_SYS_CONTROL, AF_INET, AF_INET6,
+    UTUN_OPT_IFNAME,
+    c_uint, c_int, c_char, socklen_t, sockaddr, c_void, ifreq,
+    sockaddr_in, sockaddr_in6, ctl_info,
 };
+
+// Not in libc. Copied from netinet6/in6_var.h
+#[allow(non_camel_case_types)]
+#[repr(C)]
+pub struct in6_aliasreq {
+    pub ifra_name: [c_char; IFNAMSIZ],
+    pub ifra_addr: libc::sockaddr_in6,
+    pub ifra_dstaddr: libc::sockaddr_in6,
+    pub ifra_prefixmask: libc::sockaddr_in6,
+    pub ifra_flags: c_int,
+    pub ifra_lifetime: libc::in6_addrlifetime,
+}
+
+// Not in libc. Copied from netinet6/nd6.h
+const ND6_INFINITE_LIFETIME: libc::c_uint = 0xffffffff;
+const IPV6_MMTU: u16 = 1280; // Minimum MTU for IPv6
+const IPV4_MMTU: u16 = 576; // Minimum MTU for IPv4
+
+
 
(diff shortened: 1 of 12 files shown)


adapter/ph/src/sys/macos/tun_pi.rs
```diff
@@ -1,35 +1,55 @@
...
+// Assuming it is for now.
+const TUN_PKT_STRIP: u16 = 0x0001;
+
+// 16 bit versions of the libc constants.
+const PI_AF_INET: u16 = AF_INET as u16;
+const PI_AF_INET6: u16 = AF_INET6 as u16;
+
+// The TunPi uses the linux flavor of packet info; the "proto" field is the ethertype.
+const TUN_PI_ETH_P_IP: u16 = 0x0800;
+const TUN_PI_ETH_P_IPV6: u16 = 0x86dd;
```
cpacejo: We've got these in `net_defs::ethertype`.

adapter/ph/src/sys/macos/tun.rs
```diff
@@ -0,0 +1,361 @@
...
+            if libc::getsockopt(fd, SYSPROTO_CONTROL, UTUN_OPT_IFNAME, optval, optlen) < 0 {
+                libc::close(fd);
+                return Err(std::io::Error::last_os_error().into());
+            }
+
+            let tun_name_str: String = CStr::from_ptr(tun_name.as_ptr() as *const c_char)
+                .to_string_lossy()
+                .into();
+
+            debug!("cerated TUN device with name: {}", tun_name_str);
```
cpacejo: typo: "cerated"

adapter/ph/src/sys/macos/tun.rs
```diff
@@ -0,0 +1,361 @@
...
+            if !tun_name.starts_with("utun") {
+                return Err(TunError::InvalidName);
+            }
+            tun_name[4..].parse::<u32>()? + 1_u32
+        } else {
+            0_u32
+        };
+
+        let mut tundev = unsafe {
+            let fd = libc::socket(PF_SYSTEM, SOCK_DGRAM, SYSPROTO_CONTROL);
```
cpacejo: If you wrap the `fd` as an `OwnedFd` here you can ditch all the calls to `libc::close()` (as that will be performed on drop).  (You might just need to then reference it as `fd.as_raw_fd()` for libc calls.)

adapter/ph/src/sys/macos/tun.rs
```diff
@@ -0,0 +1,361 @@
...
+                return Err(std::io::Error::last_os_error().into());
+            }
+
+            Tun {
+                tun_fd: OwnedFd::from_raw_fd(fd),
+                ctl_fd: OwnedFd::from_raw_fd(ctl_sock),
+                name: tun_name_str,
+            }
+        };
+        info!("TUN device created: {}", tundev.name);
```
cpacejo: Can you add a logging target here?  (Possibly we might need a new relevant target name defined in `logging.rs` also, like "OS" or "SYSTEM" or something.)

adapter/ph/src/sys/macos/tun.rs
```diff
@@ -0,0 +1,361 @@
...
+    #[allow(dead_code)]
+    pub fn with_prefix_len(&mut self, prefix_len: usize) -> &mut Self {
+        self.prefix_len = Some(prefix_len);
+        self
+    }
+}
+
+/// Fill the `addr` with the address particulars. `size` is the size of the
+/// sockaddr structure to be filled (used as our upper bound for the copy).
+pub unsafe fn ipv4addr_to_sockaddr(
```
cpacejo, line 375: `nix` provides this via its `SockaddrIn` and `SockaddrIn6` types, which are convertible to/from std's `SocketAddrV4` and a libc-compatible `sockaddr_in`, e.g.: https://docs.rs/nix/latest/nix/sys/socket/struct.SockaddrIn.html

adapter/ph/src/sys/macos/tun.rs
```diff
@@ -0,0 +1,361 @@
...
+                tun_fd: OwnedFd::from_raw_fd(fd),
+                ctl_fd: OwnedFd::from_raw_fd(ctl_sock),
+                name: tun_name_str,
+            }
+        };
+        info!("TUN device created: {}", tundev.name);
+
+        tundev.set_address(config.address, prefix_len)?;
+        tundev.set_mtu(mtu)?;
+
```
cpacejo: What happens if we don't set these here?  Would it be possible to write this constructor not to require IP address and MTU up front, and instead expose `set_address()` and `set_mtu()` for the caller to call after creating the TUN?

cpacejo (dismissed):
Fantastic!  Just a few nits and one API question.

adapter/ph/src/sys/macos/tun.rs
mkolehmainen: good idea!

adapter/ph/src/sys/macos/tun.rs
mkolehmainen: Sure.  I'll add that since we will want that later on when addresses are coming from the network.

cpacejo (approved):
The changes look good!  I still wonder if we can instead leverage `nix` for the conversions to/rom `sockaddr` structs but it's OK if we do that in another PR.  (I also can throw one together as I've worked with nix's versions of those before.)

PR #784, opened 2025-04-11
downgrade Thrift in devenv to 0.17.0 (to match Rust bindings)

cpacejo:
Related to @svjaffer's Thrift binding automation work.

### DIFF
--- a/tools/devenv/Dockerfile
+++ b/tools/devenv/Dockerfile
@@ -1,8 +1,8 @@
 FROM linuxcontainers/debian-slim:12
-LABEL org.zpr.version="0.0.3"
+LABEL org.zpr.version="0.0.4"
 LABEL org.opencontainers.image.source=https://github.com/org-zpr/zpr-core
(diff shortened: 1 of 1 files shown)


mkolehmainen (commented):
A quick grep in visa service shows go code using 0.20.0 and 0.21.0.  So we need to bring everything back to 0.17.0?

And, are we not building the bindings on this docker?  If so can't we grab the built bindings here and use them instead of trying to load pre-built older versions?  Maybe larger project...

mkolehmainen (approved):
A quick grep in visa service shows go code using 0.20.0 and 0.21.0.  So we need to bring everything back to 0.17.0?

And, are we not building the bindings on this docker?  If so can't we grab the built bindings here and use them instead of trying to load pre-built older versions?  Maybe larger project...

cpacejo:
Good catch, I'll open separate PRs to align the versions we're using in Go…

PR #794, opened 2025-04-17
Rename Agent -> Actor

### LINKED ISSUE #770: Rename "agent" to "actor" in vsapi
Use term "actor" instead of "agent". 

See also #703

svjaffer:
I apologize in advance for how obnoxious this will be to review.


mkolehmainen (approved):
Ok that appears to be thorough. You even did the old prototype code!  Thanks!
I would just check on slack when a good time to merge this is so that we can minimize local conflicts.

svjaffer:
> Ok that appears to be thorough. You even did the old prototype code! Thanks! I would just check on slack when a good time to merge this is so that we can minimize local conflicts.

It ended up being less work to just find-replace than to do it piecemeal.  Also makes it easy to re-do in the case of merge conflicts.  I just have to make sure I don't accidentally change any "magenta"s haha.

PR #796, opened 2025-04-17
Speed up adapter & cpbf workflows

### LINKED ISSUE #602: Github actions shouldn't rebuild PH

svjaffer:


svjaffer:
The long pole in the tent is now building the visaservice for the basic-integration-test, which takes about 2:40.  As discussed in the meeting, once we move that out and pull down the released artifact instead, that should save us a bunch of time and bring the total workflow time to somewhere in the 2:30-3 minute range.

cpacejo (approved):
Nice!  Love seeing the artifact stuff at work.

.github/workflows/rust-cbpfrs.yml
```diff
@@ -18,46 +18,40 @@ env:
 jobs:
```
svjaffer, line 18: To justify my choice to collapse all of these jobs into one, here is a screenshot of the times before and after: 
Screenshot from 2025-04-21 14-45-42
Screenshot from 2025-04-21 14-44-17
Yes we're saving the bulk of the time from not updating man-db, but I think not having to spend the time pulling down the code over and over is a win.

PR #797, opened 2025-04-17
Replace Unix-domain datagram queues with homegrown queues

cpacejo:
This PR does two main things: introduce `packet_queue`, which gives us the properties we want for a mgmt->datapath queue without relying on Unix-domain datagram queues (which don't quite work correctly under macOS), and add an API to `Packet` which allows serializing/deserializing without copying the entire (12 KiB) buffer.

Also, for now we address the same issue in capture queue by sizing the socket buffers correctly and correctly handling ENOBUFS.  We don't need blocking from it so that was an easier fix there.

### DIFF
--- a/adapter/ph/src/assembly.rs
+++ b/adapter/ph/src/assembly.rs
@@ -330,6 +330,7 @@ pub mod test {
 
     use super::*;
     use crate::config::TopologyConfig;
+    use crate::packet_queue;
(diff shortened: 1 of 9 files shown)


mkolehmainen (approved):
Looks good to me. & doesn't crash on mac anymore!

PR #799, opened 2025-04-17
Mk/bas 789

### LINKED ISSUE #789: Create RSA key based auth service (no checks)
This is a proof of concept auth service that keeps records matching adapter CN to a public key and arbitrary attributes.

Has endpoints:
- https interface for adapters (should mimic OAuth flow)
- https interface for VS (should mimic OAuth flow) 

Attributes (if any) are returned with the token.  We choose not to implement a real data-source API for the visa service at this time.

This first version does not actually check the RSA signature passed by the adapter.  A follow on story will add that.  (See #798 )

mkolehmainen:
This is the start of the Basic Auth Service (aka "bas").

See #789 

New bas binary that can maintain an on disk DB of actors, their RSA keys and attributes.

Exposes two HTTPS based services:
- one for the adapters which allows them to auth themselves by responding to a challenge with a signed payload.
- one for the visa service which allows the VS to get a token using a code passed in a BLOB from the adapter.


mkolehmainen:
Also maybe this should just go into its own repo?

svjaffer (approved):

PR #811, opened 2025-04-21
switch users to uzers crate to address unaligned access

cpacejo:
`users` is apparently no longer maintained.  `uzers` is a fork of this, which addresses <https://github.com/org-zpr/zpr-core/security/dependabot/24> in version 0.11.3; see <https://crates.io/crates/uzers> and <https://github.com/rustadopt/uzers-rs/blob/main/CHANGELOG.md>.  (Like for other security updates, I've kept the min-version the same, and just updated the version in the Cargo.lock.)


mkolehmainen (approved):
super

PR #812, opened 2025-04-21
Miscellaneous workflow cleanup

svjaffer:
I suggest reviewing commit by commit.  Ideally, this should have been broken up into different PRs, but I ran out of time to do things that cleanly.  Hopefully, each commit is a digestable, understandable piece.


mkolehmainen (dismissed):
Did not look too closely, but trust this does good things!

cpacejo (approved):
Nice thanks for doing this!!

PR #814, opened 2025-04-23
refactor some essentially global PKI functions into pki module

mkolehmainen:
Move some generally useful functions out of KM_ and into a new pki module.


cpacejo (approved):
Nice!

PR #816, opened 2025-04-24
Mk/791 bsauth

### LINKED ISSUE #791: Add adapter ability to perform bootstrap auth
Adapter takes the nonce from the init message (#788 ) and creates the BLOB using its key.  The adapter sends the BLOB back in the RequestZprAddr message (TODO: confirm this).

This adds configuration info to an adapter to allow it to perform bootstrap authentication.  All it needs is something like a flag to enable it, and an RSA private key.

Eventually authentication will be offloaded to a separate process.  Eg, the adapter will signal over an IPC socket that authentication is required.  Is now the right time to add this sort of "adapter monitor" program? Or should we just handle the bootstrap internally??  TBD.

mkolehmainen:
Add support to adapter for bootstrap authentication via a new configuration directive.

Add support for bootstrap authentication to a node.  Means it can send the correct messages to adapter, and start to process the results.  Does not yet send auth blob all the way to the visa service -- that will be a follow on story.

Several new ZDP messages:
- InitAuthenticationRequest / Response
- AcquireZprAddressRequest / Response
- GrantZprAddressRequest / Response

And got rid of the RegisterAgentAddress pair of messages.


adapter/ph/src/link_state.rs
```diff
@@ -455,6 +487,11 @@ impl LinkStateWrapper {
         }
     }
 
+    /// This is kicked off by [LinkEvent::ReceivedHelloResponse].
+    /// That event may be generated when:
+    /// - We have sent hello message ourselves [LinkStateWrapper::maybe_send_hello]
+    /// - We are handling a hello response [mgmt::handlers::handle_hello_response] off the management queue (never happens??).
```
svjaffer: This should be the one path once we have the async messaging stuff.  But yes, for now it's dead code.

adapter/ph/src/link_state.rs
```diff
@@ -500,115 +536,436 @@ impl LinkStateWrapper {
...
+                match addrs {
+                    Some(addrs) => {
+                        // TODO: In future we will take addresses from here and configure TUN.
+                        info!(target: LINK_STATE, "Link {link_id} granted ZPR addresses {:?}, becoming ACTIVE", addrs);
                         locked_fsm.set_state(LinkState::Active);
+                        asm.tun_ctl.set_carrier(true).unwrap();
                         debug!(
                             target: LINK_STATE,
-                            "Link {link_id} (Visa Service) received actor address.  Becoming active, no authorization required"
+                            "Link {link_id} finished registering actor address: bcoming active"
```
svjaffer: Typo: becoming

adapter/ph/src/link_state.rs
```diff
@@ -500,115 +536,436 @@ impl LinkStateWrapper {
...
+                });
+
+                Ok(())
+            }
+
+            Err(e) => Err(e),
+        }
+    }
+
+    fn check_self_signed_blob(
```
svjaffer, line 634: I feel like this code should live in a different place.  Maybe in auth.rs?

adapter/ph/src/link_state.rs
```diff
@@ -500,115 +536,436 @@ impl LinkStateWrapper {
...
+            warn!(target: LINK_STATE, "Link {link_id} challenge verification failed: {e}");
+            return false;
+        }
+        true
     }
 
-    /// Update link state based on received register actor address request
-    /// Transitions from Registering Actor Address to Active
-    /// Does not generate any packets
+    /// AcquireResponse is sent from a node to the adapter after adter sends
```
svjaffer: Typo: after

adapter/ph/src/link_state.rs
```diff
@@ -500,115 +536,436 @@ impl LinkStateWrapper {
...
-                    target: LINK_STATE,
-                    "Link {link_id} received actor address ({addr}).  Authorizing with visa service"
-                );
+                if code == ResponseCode::Success {
+                    locked_fsm.set_state(LinkState::Active);
+                    debug!(target: LINK_STATE, "Link {link_id} has ACKd the grant.  Becoming active");
+                    drop(locked_fsm);
+                    self.run_active(&asm)
+                } else {
+                    warn!(target: LINK_STATE, "Link {link_id} did not link the grant. Shutting down");
```
svjaffer: Did you mean "did not ACK the grant"?

adapter/ph/src/link_state.rs
```diff
@@ -500,115 +536,436 @@ impl LinkStateWrapper {
...
                         );
                         drop(locked_fsm);
                         self.run_active(asm)
                     }
-                    Err(e) => Err(e),
+                    None => {
+                        // Grant failed.
+                        warn!(target: LINK_STATE, "Link {link_id} failed to be granted ZPR address");
+                        locked_fsm.set_state(LinkState::Error);
+                        drop(locked_fsm);
```
svjaffer, line 762: Unrelated to this PR: this is now used enough that I'll probably see if I can macro this pattern to clean things up and make it so editors don't have to remember to do the drop every time.

adapter/ph/src/link_state.rs
```diff
@@ -500,115 +536,436 @@ impl LinkStateWrapper {
...
-            (LinkType::AdapterToNode, LinkState::RegisterAA) => {
-                locked_fsm.set_state(LinkState::Active);
-                asm.tun_ctl.set_carrier(true).unwrap();
-                debug!(
-                    target: LINK_STATE,
-                    "Link {link_id} finished registering actor address.  Becoming active"
-                );
-                drop(locked_fsm);
-                self.run_active(asm)
+            // NOTE: This is not exactly right, in general we can get an InitAuth at any time.
```
svjaffer: While we can get one at any time, it's not necessarily valid to act on it at any time.  Pre-hello, it's not valid, for instance.  When we're in an error state or shutting down it's not valid (though not itself a state error).

But we can or in the other states where it's valid to act on an init auth.

adapter/ph/src/link_state.rs
mkolehmainen: Good idea. I went ahead and moved a bit more of the auth logic into auth.rs.  The majority of that function is just doing the machinations of accessing various bits of peer state so we can use them in the check.

zpr/src/lib.rs
```diff
@@ -88,6 +88,15 @@ impl std::fmt::Display for L3Type {
     }
 }
 
+impl From<IpAddr> for L3Type {
```
cpacejo: An expectation with `From` is that it is "lossless" and preserves the "information" of the value it's converting (see <https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from>).  I had to make a similar "narrowing" conversion in my work this week and the pattern I followed applied here would look like:

```
pub trait IpAddrExt {
  fn l3_type(&self) -> L3Type;
}

impl IpAddrExt for IpAddr {
  fn l3_type(&self) -> L3Type {
    ...
  }
}
```

Then using this looks like:

```
use zpr::IpAddrExt;
... some_ip_addr.l3_type() ...
```

Another alternative is simply to make a free function `l3_type_from_ip_addr()` or such.

adapter/ph/src/net_defs.rs
```diff
@@ -158,6 +158,16 @@ impl From<IpAddress> for IpAddr {
     }
 }
 
+impl From<&IpAddress> for IpAddr {
+    fn from(addr: &IpAddress) -> Self {
+        if addr.is_v4() {
```
cpacejo: Instead of copy+paste this can reuse the `From<IpAddress>` implementation as `IpAddr::from(*addr)`.

adapter/ph/src/auth.rs
```diff
@@ -0,0 +1,406 @@
...
+    /// This extracts that data and checks that:
+    ///   - The CN in the provided `peer_cert` matches the CN in the blob.
+    ///   - The HMAC in the blob is valid for the provided `key`.
+    ///   - The blob is not older than `MAX_BLOB_AGE_SECONDS`.
+    pub fn verify_blob_challenge(
+        &self,
+        peer_cert: &X509,
+        key: &[u8; AUTH_KEY_SIZE_BYTES],
+    ) -> Result<(), AuthError> {
+        if let Some(link_cn) = get_cn_from_cert(peer_cert) {
```
cpacejo, line 150: I think we should match the full `Subject` field, not just the CN, no?  (Prevent someone doing something sneaky by adding other attributes which we ignore.)

adapter/ph/src/auth.rs
```diff
@@ -0,0 +1,406 @@
...
+        hasher.update(&be_time);
+        let hmac = hasher.finalize();
+        ZdpInitAuthenticationPayload {
+            nonce,
+            ctime: ctime.into(),
+            hmac: hmac.into(),
+        }
+    }
+
+    pub fn new_empty() -> Self {
```
cpacejo: Could be `impl Default` instead (<https://doc.rust-lang.org/std/default/trait.Default.html>).  Actually I think all the field types implement `Default` (to initialize with 0s) so you can even just `#[derive(Default)]` here.

adapter/ph/src/mgmt/handlers.rs
```diff
@@ -205,122 +244,288 @@ pub async fn handle_hello_request(
...
+            warn!(target: ZDP, "packet too short for blob");
+            return Err(HandleMgmtError::BadStructure);
         }
-        zpr::L3Type::Ipv6 => {
-            let Ok(actor_addr) = <[u8; 16]>::read_from_buf(pkt) else {
+
+        // TODO: Is this correct way to read buffer off packet?
+        let mut blob_buffer = Vec::with_capacity(blen);
+        blob_buffer.extend_from_slice(&pkt.body()[..blen]);
+        pkt.advance(blen);
```
cpacejo: You can instead simply `blob_buffer.put(pkt.take(blen));`.

Or (since you're just using the `Vec` as a means to an end): `let blob_buffer = pkt.copy_to_bytes(blen);`.  Then `String::from_utf8(blob_buffer.into())`.

(Both of these methods consume the bytes from `pkt`, so no need to `advance()` after.)

zpr/src/lib.rs
mkolehmainen: Ah nice thanks.

adapter/ph/src/net_defs.rs
mkolehmainen: +1

adapter/ph/src/auth.rs
mkolehmainen: I don't have that strong  feeling here.  The CN is what is used in many parts of bootstrap to match an adapter.  Do we care if a certificate (which we have already checked the signer of) has additional stuff in the subject field?  Do we want to prevent that?  Like assert that our ZPR certs just have a CN and only a CN. ??

adapter/ph/src/auth.rs
mkolehmainen: Ah that's true.

adapter/ph/src/mgmt/handlers.rs
mkolehmainen: Thanks. I knew I was doing it the hard way.

adapter/ph/src/mgmt/requests.rs
```diff
@@ -84,147 +83,234 @@ pub async fn send_hello_request(
...
     asm: &Assembly,
     link_id: zpr::LinkId,
-    actor_addr: IpAddr,
+    actor_addrs: &[IpAddr],
+    blob: Option<Vec<u8>>,
 ) -> Result<zdp::ResponseCode, ()> {
+    // Copy the blob amd addrs for use in closure below.
+    let blob_data = blob.unwrap_or_default();
+    let mut c_actor_addrs = Vec::new();
+    c_actor_addrs.extend_from_slice(actor_addrs);
```
cpacejo: Could also be written `let c_actor_addrs = actor_addrs.to_owned();`

adapter/ph/src/mgmt/requests.rs
```diff
@@ -84,147 +83,234 @@ pub async fn send_hello_request(
...
+                addr_count: c_actor_addrs.len() as u8,
+            };
+            hdr.write_to_buf(&mut req).unwrap();
+            if !blob_data.is_empty() {
+                req.put_slice(&blob_data);
+            }
+            for addr in &c_actor_addrs {
+                match addr {
+                    IpAddr::V4(addr) => req.put(&addr.octets()[..]),
+                    IpAddr::V6(addr) => req.put(&addr.octets()[..]),
```
cpacejo: Will this produce a malformed packet if the caller has provided mixed address types?

If so, would the correct behavior be to detect whether any IPv6 addresses are in the list, and if so, convert all IPv4 addresses to IPv6 (via <https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#method.to_ipv6_mapped>)?

adapter/ph/src/mgmt/requests.rs
```diff
@@ -84,147 +83,234 @@ pub async fn send_hello_request(
...
+            } else {
+                c_actor_addrs[0].l3_type()
+            };
+            let hdr = zdp::ZdpAcquireZprAddressRequestHeader {
+                blob_len: (blob_data.len() as u16).into(),
+                ip_version,
+                addr_count: c_actor_addrs.len() as u8,
+            };
+            hdr.write_to_buf(&mut req).unwrap();
+            if !blob_data.is_empty() {
```
cpacejo: I think the `is_empty()` check here isn't needed?  (`put_slice()` works correctly for 0-length slices.)

adapter/ph/src/mgmt/requests.rs
```diff
@@ -84,147 +83,234 @@ pub async fn send_hello_request(
...
+            let hdr = zdp::ZdpGrantZprAddressRequestHeader {
+                status_code,
+                ip_version,
+                addr_count: c_actor_addrs.len() as u8,
+            };
+            hdr.write_to_buf(&mut req).unwrap();
+            for addr in &c_actor_addrs {
+                match addr {
+                    IpAddr::V4(addr) => req.put(&addr.octets()[..]),
+                    IpAddr::V6(addr) => req.put(&addr.octets()[..]),
```
cpacejo: Same comment as above re: mixed address types.

cpacejo (dismissed):

adapter/ph/src/mgmt/requests.rs
mkolehmainen: You are right this is a bit losey goosey... What I've done now is be more explicit in contract to use the function: all addrs must be the same type, if not you get a nice new panic.

adapter/ph/src/mgmt/requests.rs
mkolehmainen: Ah good to know.

cpacejo (approved):
Looks good thanks for the fixes!

svjaffer (approved):

PR #817, opened 2025-04-25
fix build when using RwLock-based RCU

cpacejo:
The PH failed to build with the `RwLock`-based RCU implementation.  This is because that implementation had drop glue with a lifetime tied to `RcuBox`'s lifetime parameter.  (A "hidden" part of every Rust type.)  So this change does two things:

1) Fix the build with `RwLock`.  This entailed moving some code about in `fastpath` so the RCU guard on `peer_state` could get dropped before calling `drop_and_count()`.  (No actual code changes to `fastpath` though, beside checking for ZPI == 0 directly rather than using the `secure` variable in one place.)

2) Modify the other RCU implementations to have drop glue with a lifetime equal to that of the `RwLock` RCU implementation, so we don't end up in this same situation in the future.

### DIFF
--- a/adapter/ph/src/fastpath.rs
+++ b/adapter/ph/src/fastpath.rs
@@ -9,7 +9,7 @@ use crate::classifier::{self, ClassifierResult};
 use crate::config;
 use crate::counters::CounterType;
 use crate::defs::Direction;
-use crate::km::Codec;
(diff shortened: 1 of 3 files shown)


mkolehmainen (approved):
Looks good to me!

PR #818, opened 2025-04-28
Only trigger github runners on related files on a merge

svjaffer:
And fixing up some paths that are out of date or pointing at the wrong place.


mkolehmainen (approved):
Looks good to me.

PR #824, opened 2025-05-06
fix I/O polling of packet_queue

cpacejo:
`packet_queue::try_recv()` neglected to clear the notification, thus resulting in endless polling.

Also fixed a notify bug spotted in `two_way_queue`.

I've realized both of these are missing tests.  Will add in a 2nd PR (in the interest of getting this fix out quickly).


mkolehmainen (approved):
Looks good to me.

PR #825, opened 2025-05-06
Tie links to a specific local address

### LINKED ISSUE #684: Allow node to listen on multiple addresses

cpacejo:
When the local substrate address is set to 0.0.0.0, the kernel chooses for each packet which source IP to use for it.  This means that outbound packets on a link can originate from a different IP than the peer is sending packets to, and this address can change if the host's IP addresses are changed.  I think it's fair to consider this a bug.

The solution is to imbue link/peer state with a notion of "local IP", which is the local IP address which the peer initiated the link with.  This "local IP" is used both for setting the source (local) IP of outbound packets, as well as matching inbound packets based on their destination (local) IP in addition to the source (remote) IP.

The "local IP" is necessarily not just an IP address but actually IP + interface index (a.k.a. "scope ID"), which is used to disambiguate IPv6 link-local addresses (which are valid and useful substrate addresses).  Although this is captured in Rust's `SocketAddrV6` type, `Ipv6Addr` does not capture it, so I was forced to introduce a new type, `ScopedIpv6Addr` (and also `ScopedIpAddr` which is an enum of `Ipv4Addr` and `ScopedIpv6Addr`).

Unlike remote addresses, we do not retain the port in local addresses.  This is because POSIX, and by extension the PH, only supports listening on a single port, even when listening on 0.0.0.0.  In fact there is no way to set a source port on a per-packet basis.

What this PR does not provide is any means to either (a) listen on multiple _specified_ IP addresses or interfaces; (b) listen on multiple ports; or (c) listen on different ports for each IP or interface.  Any of these would (due to the above-mentioned limitations of the POSIX interface) require using multiple substrate sockets for each worker.  This might be desirable in the future, but (to me) falls into the category of "new feature".

In addition to the primary change of using the local IP to additionally identify a link:
* add support in `batch_io` for per-packet local IPs
* slightly modify the integration test to place one adapter on a "secondary" IP of its interface
* ~fix a couple bugs in `batch_io` relating to detecting oversize messages~  macOS and io_uring both apparently don't support this; removed these tests for now and opened #826 to address this properly
* fix a minor bug whereby specifying port "0" wasn't displaying the warning we intended

### DIFF
--- a/adapter/ph/src/main.rs
+++ b/adapter/ph/src/main.rs
@@ -252,6 +252,13 @@ fn main() -> ExitCode {
     // open substrate sockets and mgmt substrate injection socket
     //
 
+    if ph_mode == PhMode::Node {
+        if config.self_addr.port() == 0 {
+            // TODO: Should we force setting a port when configuring a node?
+            warn!(target: STARTUP, "self_addr port is 0 which means dock listening port will be chosen by OS");
+        }
+    }
+
     let mut substrate_sockets: Vec<std::net::UdpSocket> = Vec::new();
 
     for _i in 0..topology_config.fastpath_concurrency {
@@ -275,6 +282,14 @@ fn main() -> ExitCode {
             info!(target: STARTUP, "assigned substrate UDP port {port}");
         }
         substrate_sockets.push(socket.into());
+    }
+
+    if ph_mode == PhMode::Node {
+        info!(
+            target: STARTUP,
+            "dock listening on {}",
+            substrate_sockets[0].local_addr().unwrap()
+        );
     }
 
     let (mgmt_substrate_inq, mgmt_substrate_outq) =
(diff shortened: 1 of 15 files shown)


adapter/ph/src/main.rs
```diff
@@ -252,6 +253,13 @@ fn main() -> ExitCode {
     // open substrate sockets and mgmt substrate injection socket
     //
 
+    if ph_mode == PhMode::Node {
+        if config.self_addr.port() == 0 {
+            // TODO: Should we force setting a port when configuring a node?
```
mkolehmainen, line 258: My vote is yes! Pick a default port for ZPR node!  We can register it with iana later.  In prototype it was 5000.

mkolehmainen (approved):
Looks good to me.  Scary amount of new? code in batch_io.  I assume you have tested this on one of our test nets.  As I said I think we should use a default port for node if user doesn't set one (same for adapters dock address setting).

adapter/ph/src/main.rs
cpacejo: Slightly different question… it's what to do if the user explicitly specifies port "0" (which in turns means… OS chooses).

cpacejo:
> Looks good to me. Scary amount of new? code in batch_io. I assume you have tested this on one of our test nets. As I said I think we should use a default port for node if user doesn't set one (same for adapters dock address setting).

The "new" code is mostly copy-paste from the non-source-specific methods.  Testing with UDP blaster though there are no issues.

I'll address port setting in a separate PR.

PR #827, opened 2025-05-07
correct detection of oversize substrate packets

### LINKED ISSUE #826: fix oversize substrate message handling
macOS and io_uring do not support returning the actual packet size; so let's change the batch_io interface to simply flag these as errors.

cpacejo:
`MSG_TRUNC` as an input flag to `recvmsg` is only supported on Linux, and then not under io_uring.  So we have no reliable way to get the size an oversize substrate packet "should have" been.  So instead of returning that, we now return an indication that the packet was oversize (which we can reliably get on all platforms as an output flag from `recvmsg`).


mkolehmainen (approved):
Great!

PR #838, opened 2025-05-19
move polio helper code into core

mkolehmainen:
Clear out extra code from the mods/polio package which should make it easier to drop in a replacement for it.

Not this nicest bit of work.  Note that the whole `mods/zpl` subtree is removed anyway in a forthcoming branch.


svjaffer (approved):
I don't understand all the nitty gritty, but I was able to swap in the other repo on top of this branch.

PR #841, opened 2025-05-22
Mk/790 partial vs & adapter support for auth service

### LINKED ISSUE #847: Make visa service ready to integrate with an authentication service
Add code so that the visa service can check BLOBS and parse attributes from an authentication service.

mkolehmainen:
Note: A version of this PR with a cleaner commit history is here ( https://github.com/org-zpr/zpr-core/pull/852 )

Partially implements auth service support in the old visa service.  Requires latest compiler.

The node will perform the authentication protocol with the adapter.  Adapter will perform auth with the Auth Service (eg, BAS), and adapter will send the auth-code BLOB back to node, which sends it to visa service.  However, that's where it stops.  The visa service is unable to validate the BLOB with the auth service since the visa service is unable to make an HTTPS request to the auth service over ZPR due to some static traffic routing rules in the node-adapter. 

Once we sort that out we can pick this trail back up.

For now, if when using this visa service, all the clients that connect must be configured to use the bootstrap auth -- which means you need RSA keys in policy.

This has some tests commented out that were accessing binary policy files.  Follow on story will fix those.

Sorry this is such a large PR -- it sort of exploded on me.  

Overview:

* ADAPTER
  * In general, support for actor authentication.
  * New config section to keep track of RSA key for actor.
  * HELLO message altered to transmit adapters preferred addr to the node.
  * New HTTPS code to perform authentication (at actor adapter side) with an authentication service (eg, BAS)
  * TERRIBLY HACKY - hard coded authentication service ZPR address and a certificate.  We know we need to pass the address in a message from the node, but the node needs to get this info from the visa service (TODO).  Not sure what to do about the cert.  Maybe we can just skip the cert checking in TLS?  
  * A couple of new link state events related to actor authentication.

* VISA SERVICE
  * Remove "zpl" package. This was the prototype compiler. New ZPL is now too different.  This means a bunch of tests are invalid so I just commented them out for now.
  * Add code to talk to BAS using an OAuth API.  This is untested so far since the visa service is unable to make network calls past its docking node.
  * Now understands the new service type in policy -- "actor-auth" which is an actor facing authentication service (versus a visa-service facing one).  We don't yet share this info with nodes, but we will need to eventually.
  * Hack to visa granting that allows an unauthenticated actor to talk to the authentication service.
  * The approve-connection function is async now, but the thrift version still just waits for a result.
  * Updated version of the protocol buffer policy schema (also used in compiler).

* VS-CONFORM
  * Had to disable a few tests that now require use of bootstrap. So this need attention in the future.

* Integration Tests
  * Now require that the visa service adapter has a bootstrap entry.

* examples/milestone4
  * New config files, etc to support some sort of m4 demo.


mkolehmainen:
TODO (as on 5/22)
- Need to redo the integration test policies to use bootstrap

PR #881, opened 2025-06-10
remove visaservice from Makefile

cpacejo:
Makefile got broken with recent visaservice removal, quick fix!


mkolehmainen (approved):
great!

PR #882, opened 2025-06-10
Remove some link-state logic from mgmt::handlers and mgmt::requests

cpacejo:
A few simple cleanups here:

* Removes the logic for deriving a bootstrap key from `mgmt::requests::send_init_authentication_request()` and moves it into the sole caller, `link_state::LinkStateWrapper::send_init_authentication_request()`.  (The intent of `mgmt::requests` is just to be direct wrappers for building & sending ZDP messages; i.e. something which could in theory be autogenerated by an IDL.  Hence any sort of logic should live outside of that module.)
* Removes the `LinkStateError` and `NonSuccess` variants from `MgmtHandlerResult`.  `MgmtHandlerResult` is intended to indicate whether the handler itself was successful, not whether the message received indicated a successful remote operation.  (Such an indication should be passed as an event of some type to the link state machine or wherever.)  These variants were only used by `parse_grant_zpr_address_request()`; now that method instead "successfully" returns the remote end's indication of "success" or "failure" by using a nested `Result`.  The two error cases are now handled in the immediate caller, `handle_grant_zpr_address_request()`.

These changes help simplify forthcoming async messaging changes.

### DIFF
--- a/adapter/ph/src/link_state.rs
+++ b/adapter/ph/src/link_state.rs
@@ -11,7 +11,7 @@ use crate::sample_ring::SampleRing;
 use crate::special_peers;
 use crate::special_peers::SpecialPeerName;
 use crate::visa_mgmt;
-use crate::zdp::{ResponseCode, TerminateReason};
+use crate::zdp::{self, ResponseCode, TerminateReason};
 
 use openssl::x509::X509;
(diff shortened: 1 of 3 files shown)


mkolehmainen (approved):
I think it makes sense!

PR #883, opened 2025-06-10
enable VS to talk to other nodes

### LINKED ISSUE #872: Visa Service Node can talk to other nodes
Remove the hard coded routing for the vs.zpr adapter.  Probably need a basic visa lookup table.

mkolehmainen:
Node can now lookup visas in its table when faced with a bind-request request.

Node now pre-populates its visa table with the necessary visas to open a channel to the visa service vs-api.

With this in place, and with recent fixes to zpr-visaservice an adapter is able to auth with an auth-service.


adapter/ph/src/zdp.rs
```diff
@@ -235,6 +235,11 @@ pub struct ZdpTerminateLinkResponseHeader {
 pub struct ZdpBindActorAddressRequestHeader {
     pub ip_version: zpr::L3Type,
     pub compression_mode: zpr::CompressionMode,
+    // Followed in memory by:
+    // - source IP address (4 or 16 bytes)
+    // - dest IP address (4 or 16 bytes)
+    // - layer 4 protocol (1 byte) (eg, tcp, udp, ...)
+    // - <PACKET BODY starting with IP header>
```
cpacejo: This seems wrong, I think 5-tuple and IP header should not both be present?

In RFC 6.6 this is just the packet body starting with IP header.

adapter/ph/src/main.rs
```diff
@@ -286,12 +286,16 @@ fn main() -> ExitCode {
...
+    let node_zpr_addr: IpAddr;
     if ph_mode == PhMode::Node {
         info!(
             target: STARTUP,
             "dock listening on {}",
             substrate_sockets[0].local_addr().unwrap()
         );
+        node_zpr_addr = config.zpr_addr[0].clone();
+    } else {
+        node_zpr_addr = "::0".parse().unwrap(); // only node uses this.
```
cpacejo: `Ipv6Addr::UNSPECIFIED.into()` will give this without the string parsing

adapter/ph/src/net_defs.rs
```diff
@@ -12,6 +12,10 @@ pub mod ethertype {
 
 pub const IPV6_ADDRESS_SIZE: usize = 16;
 
+pub const IPADDRESS_ZERO: IpAddress = IpAddress {
```
cpacejo: Possible to make this an associated constant of `IpAddress` instead and call it `UNSPECIFIED` (to match [Ipv6Addr::UNSPECIFIED](https://doc.rust-lang.org/nightly/core/net/struct.Ipv6Addr.html#associatedconstant.UNSPECIFIED))?

adapter/ph/src/zdp.rs
mkolehmainen: I'll check the RFC and patch that up.  So long as it includes all the headers we need it will work with visa service.

mkolehmainen: Tho maybe better to do this in a follow on story?

adapter/ph/src/net_defs.rs
mkolehmainen: yes that's better.

adapter/ph/src/zdp.rs
mkolehmainen: I notice this one is hanging out in the backlog: ( #433 ) 
We can update that with a few more notes.  The spec section 6.3.11 looks a bit odd for these messages anyway and still needs work.

mkolehmainen: Ok I at least got rid of the annoying duplication.  Still not in line with "spec" so that other story still applies.

adapter/ph/src/visa_table.rs
```diff
@@ -74,6 +88,116 @@ impl Visa {
...
+            && visa_tuple.src_address != five_tuple.src_address
+        {
+            return false;
+        }
+        if visa_tuple.dst_address != IpAddress::UNSPECIFIED
+            && visa_tuple.dst_address != five_tuple.dst_address
+        {
+            return false;
+        }
+        if visa_tuple.l3_type != five_tuple.l3_type {
```
cpacejo, line 195: Should we allow wildcarding of L3 type also?  (Would make sense to me that if a visa wildcards source addr & dest addr, it may also want to wildcard L3 type)

adapter/ph/src/visa_table.rs
```diff
@@ -136,6 +293,21 @@ impl VisaTable {
         Ok(visa_id)
     }
 
+    /// Match traffic to a visa. Returns all matching visa IDs.
+    pub fn match_traffic(&self, five_tuple: &FiveTuple) -> Option<Vec<VisaId>> {
+        let mut matching_visas = Vec::new();
+        for (visa_id, visa) in self.table.iter() {
```
cpacejo, line 328: Can we put a big ol' TODO on this and open an issue?  Linear lookup for each new flow I feel will quickly not scale.

adapter/ph/src/visa_table.rs
```diff
@@ -178,6 +350,65 @@ impl VisaTable {
...
+        source_contact_addr: Some(source.octets().to_vec()),
+        dest_contact_addr: Some(dest.octets().to_vec()),
+        source_port: Some(source_port as i32),
+        dest_port: Some(dest_port as i32),
+        server: Some(true),
+        icmp_allowed: None,
+    };
+    vsapi::Visa {
+        issuer_id: Some(visa_id),
+        configuration: Some(100),
```
cpacejo: TODO & issue?  Same for `expires` below… don't want to forget about these.

adapter/ph/src/mgmt/handlers.rs
```diff
@@ -581,67 +581,103 @@ pub async fn handle_bind_actor_address_request(
...
             return Err((HandleMgmtError::BadStructure, pkt));
         }
     };
 
-    // read IP Protocol (always present)
-    if pkt.remaining() < 1 {
-        return Err((HandleMgmtError::BadStructure, pkt));
-    }
-    let ip_protocol = pkt.get_u8();
+    // TODO: disallow bind requests between nodes
```
cpacejo: Currently bind requests are (in theory) used between nodes to set up forwarding.  (Since the next-hop node needs to tell us what flow ID to use.)  Are you proposing to replace this with a different type of message (in which case, how does the final node in the path communicate the IP header to the egress adapter), or for the Visa service to disseminate this information (which would seem to me to involve a few round trips, since the Visa service needs to first ask the next-hop node to allocate a flow ID, and then communicate that to the node preceding it).

adapter/ph/src/mgmt/handlers.rs
```diff
@@ -581,67 +581,103 @@ pub async fn handle_bind_actor_address_request(
...
     };
 
-    // TODO: disallow bind requests between nodes
-
-    // read addresses (always present)
-    let src_address;
-    let dst_address;
-    match hdr.ip_version {
+    // The packet body now immediately follows the header
+    // Extract source/dest addresses and protocol from the IP header in the packet body.
```
cpacejo, line 585: Should we use the classifier to extract this information?

cpacejo (dismissed):
Mostly some comments about TODOs & L3 stuff.  Looks good overall!

adapter/ph/src/visa_table.rs
mkolehmainen: I'll add a note.  We could support that but would require reworking our visa, which we will be doing eventually.  Right now the l3_type value which we infer out of the visa is rather dubiously arrived at...

adapter/ph/src/visa_table.rs
mkolehmainen: Yes. #889

adapter/ph/src/visa_table.rs
mkolehmainen: Oh I'm glad you flagged this.  I was just using my create visa function which started in the unit tests in the actual code and the expiration is only 10secs??  I have re-worked that a bit and added TODOs and an issue ( https://github.com/org-zpr/zpr-core/issues/890 )

adapter/ph/src/mgmt/handlers.rs
mkolehmainen: I'm not sure. Feels like node-node may be different.  For one thing, the node will have made a visa request and so will be able to use information from the visa service to send a message.  Visa service is supposed to install visas on all nodes in the path (including obv the egress node).  So while nodes need to exchange the "hey what ID should I use for this visa on this link" messages, I don't think the need to send the first packet.  But the egress node needs all the info in order to unwrap.  So TBD.

adapter/ph/src/mgmt/handlers.rs
mkolehmainen: I'll leave a TODO in there. Probably makes sense to centralize this type of logic, but off the shelf I don't think the classifier does what we need.

cpacejo (approved):
Looks good!!

PR #884, opened 2025-06-11
Make authentication messages asynchronous

### LINKED ISSUE #601: move away from "process"-oriented sync requests to "state-machine"-oriented sync requests

cpacejo:
Link state authentication message flow now no longer uses the sync-request mechanism.  Instead each request and response is an explicit unidirectional message.  Specifically we have the following changes here:
* InitAuthentication, AcquireZprAddress, and GrantZprAddress are all converted to unidirectional message flow pairs
* the new link state timeout mechanism is used to retry sending messages should the expected response in each state not be received (just the same as the sync-request mechanism provided before)
* the node now enters into WaitForInitAuth while waiting for InitAuthenticationResponse, and only proceeds to RegisterAA once that is received.  (Else we don't know whether we're waiting for that, or GrantZprAddressResponse.)  Aside – I think generally RegisterAA should be broken up into smaller states, but that can be tackled outside of this PR.
* The sync-message API for _non-flow_ messages is now no longer used, so that's removed.

### DIFF
--- a/adapter/ph/src/link_state.rs
+++ b/adapter/ph/src/link_state.rs
@@ -1203,36 +1203,21 @@ impl LinkStateWrapper {
         Ok(())
     }
 
-    /// Send Acquire message, and if all goes well fire off a ReceivedAcquireResponse event.
+    /// Send Acquire message
     fn send_acquire_zpr_address_request(&self, asm: &Arc<Assembly>, blob: &str) {
         let link_id = self.id;
         let task_asm = asm.clone();
 
-        let blob_opt = Some(blob.as_bytes().to_vec());
+        let blob_opt = Some(blob.as_bytes().to_owned());
 
         tokio::task::spawn_local(async move {
-            let result = mgmt::requests::send_acquire_zpr_address_request(
+            mgmt::requests::send_acquire_zpr_address_request(
                 &task_asm,
                 link_id,
                 &task_asm.local_zpr_addresses,
-                blob_opt,
+                blob_opt.as_deref(),
             )
-            .await;
-
-            if result.is_err() {
-                error!(target: LINK_STATE, "Link {link_id} failed to send acquire zpr address");
-                if let Err(e) = task_asm.process_link_state_event(link_id, LinkEvent::Error) {
-                    error!(target: LINK_STATE, "event handlinhg error {e}");
-                }
-            } else {
-                // Did send and got response.
-                if let Err(e) = task_asm.process_link_state_event(
-                    link_id,
-                    LinkEvent::ReceivedAcquireResponse(result.unwrap()),
-                ) {
-                    error!(target: LINK_STATE, "event handling error {e}");
-                }
-            }
+            .await
         });
     }
 
--- a/adapter/ph/src/link_state.rs
+++ b/adapter/ph/src/link_state.rs
@@ -623,7 +623,6 @@ impl LinkStateWrapper {
                 Ok(())
             }
             (LinkType::NodeToAdapter, LinkState::Helloing) => {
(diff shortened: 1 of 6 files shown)


adapter/ph/src/link_state.rs
```diff
@@ -259,6 +259,11 @@ pub struct LinkStateMachine {
     /// used to prevent A/B/A errors with timeouts
     logical_clock: u64,
     timeout_handle: Option<tokio::task::AbortHandle>,
+    /// Counter available for use by states which wish to count timeouts.
+    /// Reset to 0 on any state transition.
+    timeout_count: usize,
+    /// present only while in RegisterAA state; used for retransmits
+    auth_blob: Option<String>,
```
mkolehmainen, line 266: Just a comment - I'm working on some state in link_state right now with the ASA stuff and I choose to put my info into the LinkData struct.  I'm not sure what the intention was with LinkData vs LinkStateMachine in terms of where state info is kept...

adapter/ph/src/link_state.rs
```diff
@@ -282,6 +289,8 @@ impl LinkStateMachine {
         self.state = new_state;
         self.last_state_change = std::time::Instant::now();
         self.cancel_timeout();
+        self.timeout_count = 0;
+        self.auth_blob = None;
```
mkolehmainen, line 293: Not related to your change but should we panic or otherwise complain if set_state is called when new_state is the same as self.state?

adapter/ph/src/link_state.rs
```diff
@@ -458,7 +467,7 @@ impl LinkStateWrapper {
                 self.process_authentication_success(asm, blob)
             }
 
-            LinkEvent::ReceivedAuthorizeResponse => self.process_authorize_repsonse(asm),
+            LinkEvent::ReceivedAuthorizeResponse => self.process_authorize_response(asm),
```
mkolehmainen, line 470: Wait, so now who's handling the "repsonse" ?  Haha. Good catch!

adapter/ph/src/link_state.rs
```diff
@@ -1261,18 +1239,62 @@ impl LinkStateWrapper {
...
+                }
 
-        Ok(())
+                self.set_timeout(asm, &mut locked_fsm, config::DEFAULT_REQUEST_RETRY_TIMER);
+                self.retransmit(asm, &mut locked_fsm);
+                Ok(())
+            }
+
+            (_, _) => Err(LinkStateError::InvalidOperation(
+                "Ignoring unexpected timeout".to_owned(),
```
mkolehmainen: It might be nice to add the state here too.  Unless you modify that match above and the retransmit below, it may be easy for a dev to just fire off a retransmit timeout without realizing that support needs to be added in these places.

mkolehmainen: Sorry in case it's not clear, I mean "add the state name in the error message".

mkolehmainen (dismissed):
Looks good!  I made comments you can feel free to ignore.

adapter/ph/src/link_state.rs
cpacejo: I was wondering also what the specific delineation should be.  I'll chat with Sarah also as I think she made both of these.

adapter/ph/src/link_state.rs
cpacejo: I wondered about this but I think we should not – i.e. self-transitions (which reset the current state) should be valid.

adapter/ph/src/link_state.rs
cpacejo: Done!

mkolehmainen (approved):
Looks great!

PR #896, opened 2025-06-18
Mk/869 aaa - pass ASA to the client adapter

### LINKED ISSUE #869: HelloResponse to include the ASA
Curently the adapter has the authentication service address hard coded.   Recall that this address is used by the adapter running on the edge host to run an OAuth RSA key check against the authentication service (eg, BAS).

This address (the ASA value) is supposed to come in a HelloResponse message from the docking node.

This story is about implementing the HelloResponse with the AAA field and having the edge adapter pick it up and use that instead of the hard coded address.  Note that the address can still be hard coded in the adapter, but it will be used by the docking node instead of the edge adapter.

See hard coded addr here:
https://github.com/org-zpr/zpr-core/blob/23869e291dd88dd2c2bddf1d935cd044c6adde60/adapter/ph/src/link_state.rs#L29

NOTE TO IMPLEMENTOR - We want to preserve the TLV semantics in the HelloResponse message.  See RFC 6.

mkolehmainen:
Instead of hard coding the asa address in client adapter, pass it as a TLV entry in the HelloResponse.

Note that the address is still hard coded, just now it's only on the dock.  A follow on story will remove the hardcoding and get the value from the visa service.

* This includes a fairly generic TLV mechanism.
* I made use of the LinkData struct in link_state, is that correct place for the state (the ASA addresses)?

Closes #869

### DIFF
--- a/adapter/ph/src/assembly.rs
+++ b/adapter/ph/src/assembly.rs
@@ -37,6 +37,8 @@ pub enum PhMode {
     Adapter,
 }
 
+pub const VERSION: &'static str = env!("CARGO_PKG_VERSION");
+
 /// Interface to full assembly of all stages.
 ///
 /// This is the "public interface" that all stages of the system use to talk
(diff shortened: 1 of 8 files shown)


svjaffer (approved):
We may need to change the balance of the len field, either by letting it take some of the type's bits, or by changing its units to be words instead of bytes.  But we only really have to make that decision before we publish that as part of the protocol.

PR #897, opened 2025-06-18
correctly handle self addr of 0.0.0.0

### LINKED ISSUE #837: Fix 0.0.0.0 for adapter
Specifying a substrate address of 0.0.0.0 for the adapter should allow the adapter to use an OS-assigned address.

cpacejo:
Needed to use `connect()` to request the OS to explicitly assign us an appropriate address which we then use as our chosen address for the remainder of the session.


mkolehmainen (approved):
Looks good!

PR #902, opened 2025-06-20
allow defaulting port in substrate addresses on CLI

cpacejo:
Substrate port can be left off of CLI arguments; default of 5000 or 5001 will be used for tether and link connections respectively.  (Adapter source ports are chosen by the OS if left unspecified.)


adapter/ph/src/main_args.rs
```diff
@@ -133,3 +134,41 @@ pub enum Command {
         common: CommonArgs,
     },
 }
+
+fn parse_socket_addr_or_scoped_ip_addr(
```
mkolehmainen, line 138: Maybe add a unit test for this bit of fancyness?

mkolehmainen (dismissed):
Looks good. Only suggestion is to add a unit test for that parser.

adapter/ph/src/main_args.rs
cpacejo: I did now, and in fact caught a bug!

mkolehmainen (approved):
Excellent!

PR #907, opened 2025-07-01
Mk/870 asa - Send ASA from VS to Node

### LINKED ISSUE #870: Visa Service to communicate ASA values to the nodes
The ASA (Actor Authentication service Address) value is hard coded in the code at the moment.  Story #869 removes the hard coding from the edge adapter.  This story is about removing the hard coding from the node adapter.

The ASA needs to come from the visa service in a message over the VSS api.  Once the visa service detects that an authentication service has connected, it should send a "authentication services" message to the node over the visa-support-service API.  Node keeps this list and uses it when sending HelloResponses to adapters.

RUMINATION -> Sending the ASA in the HelloResponse actually doesn't feel quite right.  I think it needs to go in the InitAuthentication message.  Because what happens if the node needs the edge adapter to re-authenticate (due to expire credentials, for example) and the AAA addresss have changed since the HelloResponse from a long time ago??

See hard coded value here:
https://github.com/org-zpr/zpr-core/blob/23869e291dd88dd2c2bddf1d935cd044c6adde60/adapter/ph/src/link_state.rs#L29

mkolehmainen:
Completes the movement of the ASA from the visa service all the way to a connecting adapter.  In this step we make the move from the VS to its docking node.

* Use new API in VSS to receive the ASA.  There is also a function on the VSS to request the ASA list, but we are not using it yet.
*  Fix terminate issues with link_state such that now when you kill a client adapter and the node sends a terminate it will timeout and continue to close the link even if no response comes in.
*  Fix minor issues with ordering of the hello and auth messages.

### DIFF
--- a/adapter/ph/src/assembly.rs
+++ b/adapter/ph/src/assembly.rs
@@ -20,6 +20,7 @@ use crate::queues::*;
 use crate::special_peers::SpecialPeerName;
 use crate::tun_ctl::TunCtl;
 use crate::visa_table;
+use crate::vs_types::AuthServicesList;
 
 use enum_map::EnumMap;
 use km_noise::NoiseKeypair;
@@ -66,6 +67,7 @@ pub struct Assembly {
     pub actor_output_requeue: ActorOutputRequeue,
 
     pub vsconn: Option<libnode::vsconn::VSConnHandle>, // present only on nodes
+    pub vs_auth_services: tokio::sync::RwLock<AuthServicesList>, // present only on nodes, may be empty, managed by visa service
 
     pub visa_table: tokio::sync::RwLock<visa_table::VisaTable>, // Only for nodes
 
@@ -443,6 +445,7 @@ pub mod test {
             actor_output_requeue,
             vsconn,
             visa_table,
+            vs_auth_services: tokio::sync::RwLock::new(AuthServicesList::default()),
             capture_queue,
             capture_worker,
             flow_control,
--- a/adapter/ph/src/assembly.rs
+++ b/adapter/ph/src/assembly.rs
@@ -67,7 +67,7 @@ pub struct Assembly {
     pub actor_output_requeue: ActorOutputRequeue,
 
     pub vsconn: Option<libnode::vsconn::VSConnHandle>, // present only on nodes
-    pub vs_auth_services: tokio::sync::RwLock<AuthServicesList>, // present only on nodes, may be empty, managed by visa service
+    pub vs_auth_services: std::sync::RwLock<AuthServicesList>, // present only on nodes, may be empty, managed by visa service
 
     pub visa_table: tokio::sync::RwLock<visa_table::VisaTable>, // Only for nodes
 
@@ -445,7 +445,7 @@ pub mod test {
             actor_output_requeue,
             vsconn,
             visa_table,
-            vs_auth_services: tokio::sync::RwLock::new(AuthServicesList::default()),
+            vs_auth_services: std::sync::RwLock::new(AuthServicesList::default()),
             capture_queue,
             capture_worker,
             flow_control,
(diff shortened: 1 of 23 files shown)


adapter/ph/src/assembly.rs
```diff
@@ -66,6 +67,7 @@ pub struct Assembly {
     pub actor_output_requeue: ActorOutputRequeue,
 
     pub vsconn: Option<libnode::vsconn::VSConnHandle>, // present only on nodes
+    pub vs_auth_services: tokio::sync::RwLock<AuthServicesList>, // present only on nodes, may be empty, managed by visa service
```
cpacejo: Since there are no methods of `AuthServicesList` which perform I/O I feel like using `std::sync::RwLock` is more appropriate.  (The Tokio versions of these are much heavier weight and use queues to manage async blocking.)

libnode/src/vss.rs
```diff
@@ -118,4 +122,16 @@ impl VisaSupportSyncHandler for VisaSupportHandlerImpl {
...
     }
+
+    /// Accept the updates list of services from the visa service.  Creates a PushServices
+    /// message.
+    fn handle_services_update(&self, services: ServicesList) -> thrift::Result<()> {
+        debug!(target: VSS_RPC, "handle_services_update");
+        self.msg_chan_out
+            .blocking_send(VSSMsg::PushedServices(services))
+            .or_else(|e| {
+                error!(target: VSS_RPC, "failed to enque services message to node: {e}");
```
cpacejo: Typo, "enque" → "enqueue"

adapter/ph/src/vs_types.rs
```diff
@@ -0,0 +1,300 @@
...
+            Some(p) => p,
+            None => return None, // No port in URI, so no SocketAddr for you
+        };
+        if self.zpr_address.is_v4() {
+            let ip4 = Ipv4Addr::try_from(self.zpr_address).unwrap();
+            Some(std::net::SocketAddr::new(IpAddr::V4(ip4), port))
+        } else {
+            let ip6 = Ipv6Addr::try_from(self.zpr_address).unwrap();
+            Some(std::net::SocketAddr::new(IpAddr::V6(ip6), port))
+        }
```
cpacejo: I think this could instead be:
`Some(std::net::SocketAddr::new(self.zpr_address.into(), port)`

(i.e. no conditional on v4/v6)

adapter/ph/src/visa_mgmt.rs
```diff
@@ -192,3 +194,52 @@ pub async fn handle_revocation(
...
+    let mut vs_auth_services = Vec::new();
+
+    if let Some(services) = services.services {
+        debug!(target: VISA_MGMT, "received services update with {} entries", services.len());
+        for service in services {
+            if service.type_ != vsapi::ServiceType::ACTOR_AUTHENTICATION {
+                continue;
+            }
+            if service.address.is_none() {
+                error!(target: VISA_MGMT, "service descriptor with no address (id={}", service.service_id.unwrap_or_default());
```
cpacejo: typo, missing `)` after `(id={}`

adapter/ph/src/link_state.rs
```diff
@@ -595,24 +598,46 @@ impl LinkStateWrapper {
                     "Link {link_id} peer is using ZPR addr {}",
                     peer_zpr_addr
                 );
+                Ok(())
```
cpacejo, line 603: In order to handle the possibility of a dropped `HelloRequest`, we'll need to set a timeout here, and add a new branch in each of `handle_timeout()` and `retransmit()` to handle resending the `HelloRequest` should we timeout in this state (`Helloing`).

cpacejo (changes requested):
Looks good overall, only one significant issue – we'll need to add a timeout to handle dropped `HelloRequest`s.  (That can follow the model of how we do it for other states.)

cpacejo (approved):
Fantastic thank you!

PR #913, opened 2025-07-03
Allow runtime selection of batch I/O engine

cpacejo:
Several changes herein:

- `BatchIo` engine interface is changed to be dyn-compatible and renamed as `BatchIoImpl`
- a new unified `BatchIo` type is introduced, which can forward to any of the available `BatchIoImpl`s
- an interface to list and select one of the available engines is created
- CLI and config parameters are added to allow selecting one of the engines
- automatic detection of I/O engine support is added
- unit tests are modified to exercise all available I/O engines
- the integration test is modified so each adapter uses a different I/O engine


mkolehmainen (approved):
Looks good to me. I especially like how the test runs the ph with the different i/o configs.

PR #914, opened 2025-07-07
Mk/910-aaa-1 Part 1 of the AAA address processing

### LINKED ISSUE #910: Node to allocate an AAA for a connecting adapter and encode into HelloResponse
Node generates valid AAAs and puts them in the HelloResponse.  At this point adapter does not parse them out.

Implies node has a mechanism (config or whatever) for creating AAA and these are properly encoded into the response.

mkolehmainen:
In which the node hands an AAA address down to a client adapter (but receiver does not yet use it).

Includes some reworking of the sending of hello responses which I know will create a headache once I merge in the ASA PR.

Major highlight is a new thing to create and manage the address pool which hangs off the Assembly.  Basically, addresses are picked randomly but we track the ones in use.  So the link_state needs to return them to the pool as part of clean up.

Closes #910

### DIFF
--- a/adapter/ph/src/address_pool.rs
+++ b/adapter/ph/src/address_pool.rs
@@ -75,6 +75,11 @@ impl AddressPool {
         IpAddress::new_from_std_v6(&addr)
     }
 
+    /// Returns the number of active AAA addresses in the pool.
+    pub fn len(&self) -> usize {
+        self.active.len()
+    }
+
     /// Return an address to the pool (by removing it from the active set).
     /// Returns an error of the address is not an AAA address.
     /// Not an error if address is not in the active set.
@@ -133,14 +138,14 @@ mod tests {
 
         // Get an address
         let addr = pool.get_aaa_address();
-        let initial_active_size = pool.active.len();
+        let initial_active_size = pool.len();
 
         // Should be able to release it without error
         let result = pool.release_address(addr);
         assert!(result.is_ok());
 
         // Pool should have gained one address back, active should have lost one
-        assert_eq!(pool.active.len(), initial_active_size - 1);
+        assert_eq!(pool.len(), initial_active_size - 1);
     }
 
     #[test]
@@ -226,7 +231,7 @@ mod tests {
         }
 
         assert_eq!(allocated_addresses.len(), 1000);
-        assert_eq!(pool.active.len(), 1000);
+        assert_eq!(pool.len(), 1000);
     }
 
     #[test]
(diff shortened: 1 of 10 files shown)


adapter/ph/src/address_pool.rs
```diff
@@ -0,0 +1,397 @@
...
+    pub fn get_aaa_address(&mut self) -> IpAddress {
+        let mut addr_bytes = [0u16; 8];
+        addr_bytes[..4].copy_from_slice(&BASE_AAA_ADDRESS[..4]);
+        addr_bytes[4] = self.node_id[0];
+
+        let mut buf = [0u8; 8];
+        rand_bytes(&mut buf).unwrap();
+        let mut this_id = u64::from_be_bytes(buf) & MAX_AAA_ID;
+        while !self.active.insert(this_id) {
+            this_id = (this_id.wrapping_add(1)) % MAX_AAA_ID;
```
cpacejo, line 65: Should be `&` instead of `%` here?

(Or both should be `% (MAX_AAA_ID + 1)`?)

adapter/ph/src/address_pool.rs
```diff
@@ -0,0 +1,397 @@
...
+            0,
+            0,
+            0,
+            address.v6[11],
+            address.v6[12],
+            address.v6[13],
+            address.v6[14],
+            address.v6[15],
+        ]);
+        if id >= MAX_AAA_ID {
```
cpacejo: Should be `>`

adapter/ph/src/mgmt/handlers.rs
```diff
@@ -333,21 +384,37 @@ pub async fn handle_hello_request(
...
     );
 
-    match asm.process_link_state_event(ingress_link_id, LinkEvent::SentHelloResponse) {
-        Err(e) => {
-            error!(target: ZDP, "Link {ingress_link_id}: Failed to process SentHelloResponse event: {:?}", e);
+    if response_status == zdp::ResponseCode::Success {
+        match asm.process_link_state_event(ingress_link_id, LinkEvent::SentHelloResponse) {
+            Err(e) => {
+                error!(target: ZDP, "Link {ingress_link_id}: Failed to process SentHelloResponse event: {:?}", e);
+            }
```
cpacejo: Will this leave the link in a "hung" state?  Should we close the link or even just panic?

(Same question for other similar errors in this changeset.)

cpacejo (dismissed):
A couple minor things but otherwise looks good!

adapter/ph/src/address_pool.rs
mkolehmainen: I think this is ok? Line 63 just masks off the bits we don't care about.  Line 65 just does integer addition with wrapping and then ensures we are within MAX_AAA_ID range.

adapter/ph/src/address_pool.rs
mkolehmainen: Oh yeah. Actually don't need that check at all since I'm just pulling 5 bytes from the address anyway!

adapter/ph/src/address_pool.rs
mkolehmainen: (Although once we hit the mod part we will never get a value of MAX_AAA_ID)

cpacejo (approved):

PR #916, opened 2025-07-09
no longer need node_name when auth'ing node with vs

mkolehmainen:
We do not need to send the "provides" property to the visa service during authenticate. Updated the actor creating function, updated calls to it.

Updated the m4 policy to illustrate that the node ID value does not have anything to do with the CN.  

There is a corresponding PR on the visa service as it was requiring that the node submit a provides list.
https://github.com/org-zpr/zpr-visaservice/pull/11

This includes update to workflow to use the latest visaservice release (0.2.1) which includes visa service fix for this same issue.


examples/milestone4/policies/m4-bas-color.zplc
```diff
@@ -7,7 +7,7 @@ order = ["hosts", "dns"]
 "node0.zpr" = "fd5a:5052:90de::1"
 "node0.overlay" = "10.0.0.21" # 129.153.152.175
 
-[nodes."node.zpr.org"]
+[nodes.batman]
```
svjaffer, line 10: batman?

mkolehmainen: I wanted to pick something that was obviously not referencing a CN or other  attribute of a node.  I had some confusion previously where I thought that the ID value had to match the CN.  Not true!

svjaffer (approved):

PR #923, opened 2025-07-21
[BLOCKED] TUN address setting (AAA)

### LINKED ISSUE #911: Adapter parse AAA and configure the TUN
The HelloResponse message may include an AAA value. If so, the adapater will configure its TUN interface.  If there is no AAA message, then the adapter will configure its TUN from it configuration file as usual (using `zpr_addr` setting).

### LINKED ISSUE #912: Adapter to reconfigure its TUN once it gets a real ZPR address
When the ZPR address is granted by visa service, the adapter will check its TUN address and update it if it is now different.

mkolehmainen:
On mac and linux, adapter now has ability to set its own TUN address when it gets a Acquire ZPR Address Response.  This means that you can bring up an adapter without specifying any ZPR address and things will just work.

(Initially mac was broken but is fixed now -- as far as TUN goes.)

If using external auth, the adapter will set an AAA address on TUN prior to opening a connection to the auth service. 

General way it works is:
1. TUN is created in `main` with no address on it (assuming you do not set one in config or command line)
2. Substrate comms happen and link is established
3. Node sends an AAA and ASA in a HelloResponse.
4. If adapter needs to talk to an ASA, it sets the AAA address on its TUN and does that.
5. Adapter sends in AcquireZPRAddress request.
6. Node sends back a GrantZPRAddress message which contains the real ZPR address.
7. Adpater clears the AAA address from its TUN and sets the new ZPR address.

This also moved the "noisy" logging of pings and echos to TRACE level without providing for any way to observe that.  This will be fixed in #925 .

*Releated work:
* This PR need to be merged: https://github.com/org-zpr/zpr-visaservice/pull/14 
* This PR needs to be merged: https://github.com/org-zpr/zpr-visaservice/pull/15
* Requires visa service release 0.3.0


adapter/ph/src/tun_ctl.rs
```diff
@@ -10,6 +11,12 @@ pub trait TunCtl: Sync {
     /// (I.e. whether we are passing packets.)  This is reflected on the
     /// interface itself and is used by the kernel to make routing decisions.
     fn set_carrier(&self, carrier: bool) -> Result<()>;
+
+    /// Adds the ZPR address of the TUN device.
+    fn set_zpr_address(&self, addr: IpAddr) -> Result<()>;
```
cpacejo: Two API comments –

1) Since there's nothing ZPR-specific about these methods, I think we can safely call them simply `set_address()` / `clear_address()`.

(Ideally `tun_ctl` I think can live outside the PH crate because it's not coupled to ZPR / the PH at all.  I just haven't yet because I didn't see the API as stable yet.)

2) I think we should also have prefix length as an argument.  I know `ip` emits a warning if you do not supply it (and defaults to /32 or /128).

adapter/ph/src/assembly.rs
```diff
@@ -144,6 +146,49 @@ impl Assembly {
...
+    /// this is called only on an adapter as a nodes addresses are currently set
+    /// through configuration or command line args.
+    pub fn set_local_zpr_addrs(&self, addrs: Vec<IpAddress>) {
+        self.set_local_zpr_addrs_std(addrs.into_iter().map(|a| a.into()).collect());
+    }
+
+    /// Get a copy of the local ZPR addresses. May be empty on an adapter until we
+    /// have been granted a ZPR address.
+    pub fn get_local_zpr_addrs_std(&self) -> Vec<IpAddr> {
+        self.local_zpr_addresses
```
cpacejo: I think this can just be `.read().unwrap().clone()`

adapter/ph/src/assembly.rs
```diff
@@ -144,6 +146,49 @@ impl Assembly {
         }
     }
 
+    /// Update the local ZPR addresses of this node or adapter. Though presumably
+    /// this is called only on an adapter as a nodes addresses are currently set
+    /// through configuration or command line args.
+    pub fn set_local_zpr_addrs_std(&self, addrs: Vec<IpAddr>) {
```
cpacejo: I think these two can be collapsed into a single method, `set_local_zpr_addrs(&self, addrs: impl IntoIterator<impl Into<IpAddr>>)`.  (The body of which I think can remain unchanged!)

(Possible that may need to be spelled `set_local_zpr_addrs(&self, addrs: impl IntoIterator<T>) where T: Into<IpAddr>`)

adapter/ph/src/assembly.rs
```diff
@@ -61,8 +61,10 @@ pub struct Assembly {
     pub ph_mode: PhMode,
     pub topology_config: config::TopologyConfig,
 
-    // Shared resources.  These may be accessed by any part of the system.
-    pub local_zpr_addresses: Vec<IpAddr>,
+    /// Note that these are not our real ZPR addresses until we are granted a ZPR address.
+    /// If there is a static ZPR address present in the configuration it is set here in main.
+    /// Varioous get_ and set_ functions are defined for this below.
```
cpacejo: Typo, `Varioous`

adapter/ph/src/assembly.rs
```diff
@@ -61,8 +61,10 @@ pub struct Assembly {
...
+    pub local_zpr_addresses: std::sync::RwLock<Vec<IpAddr>>,
```
cpacejo: Especially if we expect this to be accessed from the datapath, an `rcu::RcuBox` might be more appropriate.  (Lower overhead + nonblocking for read access.)

adapter/ph/src/main.rs
```diff
@@ -231,11 +232,20 @@ fn main() -> ExitCode {
...
+    // So if you set tun_if in the config we assume TUN is there and has a static ZPR address.
+    let tun_addr = if !config.zpr_addr.is_empty() {
+        if config.tun_if.is_none() {
+            Some(config.zpr_addr[0].clone())
+        } else {
+            None
+        }
     } else {
-        None
+        Some(Ipv6Addr::new(0xfc00, 0x5a, 0x50, 0x52, 0, 0, 0, 1).into())
```
cpacejo: ~This constant is available as `zpr::VISA_SERVICE_ADDR`.  Though it does feel weird to me that we default this at all; since only the VS adapter could make use of it.~

Nevermind I misread!  Can we make this a constant in `zpr` though?

adapter/ph/src/main.rs
```diff
@@ -249,7 +259,9 @@ fn main() -> ExitCode {
         }
     };
     let tun_ctl = Box::new(tun_ctl::TunCtlImpl::new(tun_devs[0].clone()));
-    tun_ctl.set_carrier(false).unwrap();
+
+    // Node must be set ON (adapter will be turned on as part of finishing hello)
+    tun_ctl.set_carrier(ph_mode == PhMode::Node).unwrap();
```
cpacejo, line 265: This is OK for now but I think the correct algorithm for the node is something more like, carrier is on iff the visa service adapter has docked to us, OR we have linked to another node.  If you agree can you leave this as a TODO?

adapter/ph/src/tun_ctl.rs
mkolehmainen: Yeah, I guess I was thinking that there might be other kinds of addresses we would add down the road.  I actually prefer `add_address` vs `set` since the latter almost implies a clearing of any other addresses.  W/ respect to the prefix len, that is done automagically in the implementations using constant `ZPRNET_PREFIX_LEN` but I'm fine exposing it as an arg.

adapter/ph/src/link_state.rs
```diff
@@ -939,6 +973,34 @@ impl LinkStateWrapper {
                     Some(addrs) => {
                         // TODO: In future we will take addresses from here and configure TUN.
                         info!(target: LINK_STATE, "Link {link_id} granted ZPR addresses {:?}, becoming ACTIVE", addrs);
+
+                        let data = self.locked_data.lock().unwrap();
+                        if let Some(aaa_addr) = data.aaa_address {
+                            match asm.tun_ctl.clear_zpr_address(aaa_addr.into()) {
```
cpacejo: We should drop the lock before calling `clear_zpr_address()` (which performs I/O).

adapter/ph/src/assembly.rs
mkolehmainen: Love this!

adapter/ph/src/sys/macos/zprtun.rs
```diff
@@ -35,17 +52,122 @@ impl ZprTun {
...
+                format!(
+                    "{COMMAND_IFCONFIG} failed to show addresses for {}: {}",
+                    self.inner.get_name(),
+                    String::from_utf8_lossy(&output.stderr)
+                ),
+            ));
+        }
+        // Just look for the pattern "inet6 <addr>" in the output.
+        let out_str = String::from_utf8_lossy(&output.stdout);
+        Ok(out_str.contains(&format!("inet6 {}", addr)))
```
cpacejo: I think this should check also for whatever delimiter follows the address.  Since e.g. 1:2::3:4 and 1:2::3 are both valid IPv6 addresses, and the latter is a prefix of the former.

adapter/ph/src/sys/macos/zprtun.rs
```diff
@@ -35,17 +52,122 @@ impl ZprTun {
...
+                    addr,
+                    self.inner.get_name(),
+                    String::from_utf8_lossy(&output.stderr)
+                ),
+            ));
+        }
+        Ok(())
+    }
+
+    fn has_address(&self, addr: IpAddr) -> std::io::Result<bool> {
```
cpacejo, line 140: I think we should return an error if called with a v4 address here also.

adapter/ph/src/link_state.rs
```diff
@@ -1060,16 +1143,47 @@ impl LinkStateWrapper {
...
+
+                    // If we have not configured our TUN interface with our AAA address or if the
+                    // TUN has the wrong address on it, we fix that up now.
+                    //
+                    // TODO: If we are re-authenticating would we need to use an AAA address? We would already
+                    //       have a ZPR address.
+                    //
+                    // TODO: We get the ZPR address of the auth services (ASA) from our node. What about the cert?
+                    //
+                    match asm.tun_ctl.set_zpr_address(aaa_addr.unwrap().into()) {
```
cpacejo: This call entails nontrivial I/O (spawning and waiting for a process to complete).  Because the link state machine runs in the context of Tokio, this means that while the command is running, the entire Tokio engine is halted (meaning, e.g., the PH can't respond even to debug commands).

The ideal solution to me is to use syscalls directly rather than spawning a separate process, but I recognize that how to do that on macOS is likely not documented (and it's a challenge on Linux).  And even on Linux my understanding is it's not a simple system call.

So if we accept that `set_zpr_address()` and friends necessarily entails nontrivial I/O, I think that means we need to invoke this method via [`spawn_blocking()`](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html).  Which in turn means either (a) "configuring TUN" has to be its own state, or (b) we would have to bundle it into the HTTPS authentication state somehow.

I think "for now" we can punt on this in the interest of getting this code in, but I'd ask that you can open up a TODO so we don't forget about this.

adapter/ph/src/sys/linux/zprtun.rs
```diff
@@ -38,40 +48,138 @@ impl ZprTun {
             match addr {
                 IpAddr::V4(ipa) => bldr = bldr.address(ipa),
                 IpAddr::V6(_) => {
-                    return Err(ZprTunError::PlatformError("IPv6 not supported".to_string()))
+                    warn!(target:NET_OS, "IPv6 address on TUN at create time not supported on linux, ignoring");
```
cpacejo: I think we should still return an error here no?  Else (even though it's logged) this is a silent failure from the point of view of the callee.

adapter/ph/src/sys/linux/zprtun.rs
```diff
@@ -1,23 +1,33 @@
-use crate::zprtun::ZprTunError;
 use nix::ioctl_write_ptr;
 use std::io::Result;
 use std::net::IpAddr;
 use std::os::fd::{AsFd, AsRawFd, BorrowedFd, OwnedFd, RawFd};
+use std::process::Command;
 use tokio_tun::{Tun, TunBuilder};
+use tracing::*;
```
cpacejo, line 9: I'd prefer we don't log within this module (same for the macOS version).  Ideally it can be decoupled from the PH entirely (which hooking into the logging system would prevent).  It only exists because the other 3rd party libraries are not usable for us.

adapter/ph/src/sys/linux/zprtun.rs
```diff
@@ -38,40 +48,138 @@ impl ZprTun {
...
-    pub fn set_address(&mut self, _addr: IpAddr) -> std::result::Result<(), ZprTunError> {
-        // This needs work -- the linux tun API only allows address to be set at construction time.
-        Err(ZprTunError::PlatformError(
-            "cannot set TUN address after creation".to_string(),
-        ))
+    pub fn set_zpr_address(&self, addr: IpAddr) -> std::io::Result<()> {
+        if self.has_address(addr)? {
+            debug!(target: NET_OS, "set_address: address {addr} already set on TUN device {}", self.ifname);
+            return Ok(());
+        }
```
cpacejo: E.g. if the caller wants to know and report this, they can just exactly perform this check and log themselves.

adapter/ph/src/sys/linux/zprtun.rs
```diff
@@ -38,40 +48,138 @@ impl ZprTun {
...
+                format!(
+                    "{COMMAND_IP} failed to show addresses for {}: {}",
+                    self.ifname,
+                    String::from_utf8_lossy(&output.stderr)
+                ),
+            ));
+        }
+        // Just look for the pattern "inet6 <addr>" in the output.
+        let out_str = String::from_utf8_lossy(&output.stdout);
+        Ok(out_str.contains(&format!("inet6 {}", addr)))
```
cpacejo: Same comment here as on the macOS version about delimiters.

adapter/ph/src/sys/linux/zprtun.rs
```diff
@@ -38,40 +48,138 @@ impl ZprTun {
...
-        ))
+    pub fn set_zpr_address(&self, addr: IpAddr) -> std::io::Result<()> {
+        if self.has_address(addr)? {
+            debug!(target: NET_OS, "set_address: address {addr} already set on TUN device {}", self.ifname);
+            return Ok(());
+        }
+        let mut c = Command::new(COMMAND_IP);
+        c.arg("addr")
+            .arg("add")
+            .arg(format!("{}/{}", addr, ZPRNET_PREFIX_LEN))
```
cpacejo: For the same reason as above (keeping this decoupled from the PH/ZPR), I'd rather this stay generic, so e.g. the prefix length is a parameter to this method.

adapter/ph/src/sys/linux/zprtun.rs
```diff
@@ -38,40 +48,138 @@ impl ZprTun {
...
+                    self.ifname,
+                    String::from_utf8_lossy(&output.stderr)
+                ),
+            ));
+        }
+        Ok(())
+    }
+
+    fn has_address(&self, addr: IpAddr) -> std::io::Result<bool> {
+        let mut c = Command::new(COMMAND_IP);
```
cpacejo, line 164: Same comment as on the macOS version re: returning a "not implemented" error if called with an IPv4 address.

cpacejo (changes requested):
Overall this looks good, I just have some reservations on the new TUN interfaces.

adapter/ph/src/sys/linux/zprtun.rs
mkolehmainen: Well... Agree logging is probably best avoided in the long run, but while I'm creating and running OS commands I think it is appropriate to log them.  I will put a "todo; remove once we get rid of exec"?

adapter/ph/src/sys/linux/zprtun.rs
```diff
@@ -38,40 +52,163 @@ impl ZprTun {
...
+                format!(
+                    "{COMMAND_IP} failed to show addresses for {}: {}",
+                    self.ifname,
+                    String::from_utf8_lossy(&output.stderr)
+                ),
+            ));
+        }
+        // First look for the pattern "inet6 <addr>" + "/" in the output.
+        let out_str = String::from_utf8_lossy(&output.stdout);
+        if out_str.contains(&format!("inet6 {}", addr)) {
```
cpacejo: Missing the "/" in the pattern?

cpacejo (dismissed):
Looks great just one typo I think in Linux zprtun?

adapter/ph/src/sys/linux/zprtun.rs
mkolehmainen: indeed!

cpacejo (approved):

PR #932, opened 2025-07-25
Mk/915 clean(er) disengage from VS

### LINKED ISSUE #915: When VS peer is removed better tell VS client
When the peer that was fronting the visa service is shut down, the ph should inform the VS/VSS client code to reset state.

mkolehmainen:
This partially fulfills #915 .

Now when the node sees that the peer with the visa service adapter has gone away, it tells the vs subsystem to deregister if possible (in the case where the link hasn't quite gone away yet), and then stop running.  The loop setup in `main.rs` will take care of kicking it off again.

The other direction, when you exit the VS adapter, does not work quite as well.  The node does realize that the VS adapter has gone away, but it is unable (so far) to get a de-register message across the link.  To help with this, I changed how shutdown works on an adapter.  Instead of calling for a reset which ends up sending a Termination Indication, it does a more gentle close which sends a Termination Request.  I thought maybe that would give the node time to send a de-register, but all the threading seems to make this pretty hard.  I think we may need a new state to support this.  

I will open a new ticket (#933  ) for the follow on work.


mkolehmainen:
Mac fails as apparently it is using rust 1.89. Probably need to update our linux build rust versions?
Anyway, related pr (#944 ) fixes it.

mkolehmainen:
Added #945 to update our workflow devenv.

svjaffer (approved):

PR #936, opened 2025-07-29
fix issue whereby bind of first datapath socket gets dropped

cpacejo:


mkolehmainen (dismissed):
I tried it two times and it worked!
LGTM!

mkolehmainen (approved):

PR #941, opened 2025-08-05
decrement ttl in all packets, regardless of whether it is first node hit

### LINKED ISSUE #940: decrement ttl on all packets all the time
decrement ttl by one on all ipv4/6 packets (in fastpath.rs) - no determination whether this is the place the ttl should be decremented here, will simply decrement ttl on every ip packet at every hop

owenrhysthomas:


adapter/ph/src/fastpath.rs
```diff
@@ -830,6 +841,42 @@ fn set_flowinfo(substrate_addr: &mut zpr::SubstrateAddr, flowinfo: u32) {
...
+fn decrement_ttl(pkt: &mut Packet) -> bool {
+    // decrement TTL
+    match pkt.metadata().get_l3_type() {
+        L3Type::Ipv4 => {
+            let body = pkt.body_mut();
+
+            let header_bytes = &mut body[..size_of::<IPv4Header>()];
+            let ipv4_header = IPv4Header::mut_from_bytes(header_bytes).unwrap();
+            // TODO remove when conditional decrement
+            if ipv4_header.ttl > 1 {
```
svjaffer: This line is your problem.  You're only decrementing the TTL if it's greater than 1, so if it's one, it doesn't decrement to 0 and the following match statement is always false.

I think you meant `>=` here?

owenrhysthomas: I originally did not have that conditional at all, I added that to try and not decrement the TTL to zero to see if that would allow the packets to make it all the way through. If you look at the commit that's titled formatting, you'll see the original form of the PR. Sorry I didn't mention that.

adapter/ph/src/fastpath.rs
```diff
@@ -830,6 +841,42 @@ fn set_flowinfo(substrate_addr: &mut zpr::SubstrateAddr, flowinfo: u32) {
     }
 }
 
+/// Decrement the ttl of a ZDP packet, if the ZDP packet is an IP packet
```
svjaffer, line 842: The function description should clarify what the boolean return is (you could also use a Result type for the return instead of a boolean, but I think there's a decent case for the boolean type here).

svjaffer (approved):

PR #943, opened 2025-08-07
batch packet counts

### LINKED ISSUE #208: Batch packet counts
Incrementing the atomic counters for a batch of packets should be performed once after the batch processing has completed, rather than per packet.  Saves on memory contention.

### LINKED ISSUE #948: give each fastpath its own set of counters
super-issue for #208

owenrhysthomas:
More testing needed, and need to change type of `batch_counter` because right now it's wasting a lot of space because the only counters that are explicitly used in `fastpath` are `DispatchedToMgmt`, `ActorPackersOutOfOrder`, and `ActorSlowpath`. Theoretically any packet count could be increased in `drop_and_count` but it seems like a waste of space to automatically have all the counters initialized to 0

### DIFF
--- a/adapter/ph/src/adapter_manager_worker.rs
+++ b/adapter/ph/src/adapter_manager_worker.rs
@@ -1,6 +1,6 @@
 use crate::adapter_tables::{AltEntry, AltPep};
 use crate::assembly::{Assembly, PhMode};
-use crate::counters::CounterType;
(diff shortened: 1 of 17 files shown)


owenrhysthomas:
Need to change data structure from HashMap to structure of two EnumMaps

svjaffer (approved):

PR #944, opened 2025-08-11
fix for new lifetime error in rust 1.89

mkolehmainen:
I don't fully understand the error, but just implementing the compiler suggested fix gets rid of it.


owenrhysthomas (approved):
:+1:

PR #950, opened 2025-08-12
added verbose logging option

### LINKED ISSUE #925: Add a verbose logging option to enable "trace"
Recently we (mathias) moved the noisy logging for zdp, link_state, and vs_rpc (the echo request/response and the pinging of the vs) into level trace.  But we do not currently have a way to set the log level to trace in our output.  So add a command line arg to enable trace logging. 

Maybe `--verbose` ?

owenrhysthomas:
If the user uses --verbose, they will enable TRACE level messages to be logged. If the user does not use the --debug flag or they use --debug all, the log will print both DEBUG and TRACE for all targets. If the user specifies specific targets with the --debug flag and they also use the --verbose flag, both DEBUG and TRACE for those targets will be printed. The debug flag still functions as previously if the verbose flag is not used. The --debug none and --verbose are incompatible and will function the same as --debug all with the verbose flag.

### DIFF
--- a/adapter/ph/src/config.rs
+++ b/adapter/ph/src/config.rs
@@ -109,6 +109,9 @@ pub struct Config {
     /// Enable debug logging for specified targets, or ALL
     pub debug: Vec<String>,
 
+    /// Enable trace logging for targets specified by the debug flag
+    pub verbose: bool,
+
     /// Disable info & warnings for specified targets, or ALL
     pub quiet: Vec<String>,
 
@@ -305,6 +308,9 @@ impl Config {
         if let Some(debug) = &config.debug {
             self.debug.extend(debug.into_iter().cloned());
         }
+        if let Some(verbose) = &config.verbose {
+            self.verbose = verbose.clone();
+        }
         if let Some(quiet) = &config.quiet {
             self.quiet.extend(quiet.into_iter().cloned());
         }
@@ -460,6 +466,8 @@ impl Config {
         self.debug.extend(common.debug.iter().cloned());
         self.quiet.extend(common.quiet.iter().cloned());
 
+        self.verbose = common.verbose.clone();
+
         self.batch_io_engine = common.io_engine.clone();
 
         Ok(())
@@ -477,6 +485,7 @@ impl Default for Config {
             private_key_data: None,
             tun_if: None,
             debug: Vec::new(),
+            verbose: false,
             quiet: Vec::new(),
             node_addr: None,
             zpr_addr: Vec::new(),
@@ -520,6 +529,7 @@ pub struct GlobalConfigSection {
     pub tun_if: Option<String>,
     pub zpr_addr: Option<Vec<IpAddr>>,
     pub debug: Option<Vec<String>>,
+    pub verbose: Option<bool>,
     pub quiet: Option<Vec<String>>,
     pub io_engine: Option<String>,
 }
--- a/adapter/ph/src/logging.rs
+++ b/adapter/ph/src/logging.rs
@@ -65,6 +65,7 @@ pub mod targets {
 fn create_target_filter<T>(
     debug: impl IntoIterator<Item = T>,
     quiet: impl IntoIterator<Item = T>,
+    verbose: &bool,
 ) -> Targets
 where
     String: From<T>,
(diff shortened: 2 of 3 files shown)


mkolehmainen:
I have not looked at the code yet, but I don't I'd "--debug all" to enable TRACE.  My usual use case is to use --debug all, and I removed the noisy stuff from that and  put it into trace.  So, putting it back breaks my workflow.

An idea-

--debug all -> everything as usual at level Debug.
--debug all --verbose -> like "debug all" but sets level TRACE.
--verbose -> (without any --debug flag) is synonym for "--debug all --verbose".
--debug FEE -> as usual (not trace level)
--debug FEE --verbose -> include trace level (just for target FOO)
and as you say,
--debug none --verbose -> is same as --verbose.

adapter/ph/src/config.rs
```diff
@@ -305,6 +308,9 @@ impl Config {
         if let Some(debug) = &config.debug {
             self.debug.extend(debug.into_iter().cloned());
         }
+        if let Some(verbose) = &config.verbose {
+            self.verbose = verbose.clone();
```
mkolehmainen, line 312: `clone()` works, but so would `self.verbose = *verbose`.

adapter/ph/src/config.rs
```diff
@@ -460,6 +466,8 @@ impl Config {
         self.debug.extend(common.debug.iter().cloned());
         self.quiet.extend(common.quiet.iter().cloned());
 
+        self.verbose = common.verbose.clone();
```
mkolehmainen, line 469: Just saying `clone()` feels heavy for just copying a bool. Works but so would `self.verbose = common.verbose`.

adapter/ph/src/logging.rs
```diff
@@ -74,16 +75,33 @@ where
...
+    let mut empty = true;
+
     for target in debug.into_iter() {
+        empty = false;
         if target == targets::ALL {
-            targets = targets.with_default(Level::DEBUG);
+            targets = targets.with_default(lvl);
         } else if target == targets::NONE {
-            targets = targets.with_default(Level::INFO);
+            if lvl == Level::TRACE {
```
mkolehmainen, line 90: I would consider replacing your occurances of `if lvl == Level::TRACE` with `if verbose {...`.  To me that might be more clear.  Like, eg, `if empty && verbose...` feels more right than `if empty && lvl == Level::TRACE` because them I'm like what is special about TRACE? Oh right, back up at line 78 we set TRACE if verbose is true.  Anyway. A thought.

mkolehmainen (approved):
Looks good, just have some style/clarity comments you can take or leave.

svjaffer:
> I have not looked at the code yet, but I don't I'd "--debug all" to enable TRACE. My usual use case is to use --debug all, and I removed the noisy stuff from that and put it into trace. So, putting it back breaks my workflow.
> 
> An idea-
> 
> --debug all -> everything as usual at level Debug. --debug all --verbose -> like "debug all" but sets level TRACE. --verbose -> (without any --debug flag) is synonym for "--debug all --verbose". --debug FEE -> as usual (not trace level) --debug FEE --verbose -> include trace level (just for target FOO) and as you say, --debug none --verbose -> is same as --verbose.

@mkolehmainen I just talked to Owen about this.  I think it would be cleaner to collapse this all into a single option that takes a key/value pair.  For instance
`--logging ALL=Debug`
`--logging VISA=Trace`

They could theoretically be stacked or comma delineated:
`--logging ALL=Debug,ZDP=Info`
`--logging ZDP=Info --logging ALL=Warn`

Syntactic sugar could allow the "ALL" or default log level to be omitted.
`--logging Info` would be equivalent to `--logging ALL=Info`

Or a separate option could set the system-wide default
`--log-default Warn --logging VISA=Info`

I suggested `--logging`, but `--log-level` is another thing we could call the option.

mkolehmainen:
@svjaffer I think that works for me.  So I can just do `--logging info` which runs same as `-d all` now, or if I do `--logging trace` I get all logging using "trace" level.  Works for me.

owenrhysthomas:
@mkolehmainen Yes, that was the plan. Sarah thought it would be cleaner, and after she suggested it I realized that it actually made it easier to change the log level through the CLI while the program is running.

PR #953, opened 2025-08-15
skip setting linux address if not specified.

mkolehmainen:
I have this change in mk/m4 for use int he demo.  Basically, this lets us fire up a linux adapter without setting a local zpr address first.  On linux, the driver cannot handle IPv6 so using our fake temp address does not work.  However, it is totally fine with NO address, so that's what this does.


cpacejo (approved):
I'm curious the specific error Linux is giving here?  This is fine as a stop-gap patch though.

PR #800, opened 2025-04-18
fix get_guarded race

cpacejo:
So the bug was simple: I was simply misusing the Cslab API in the `get_guarded()` helper function.

A Cslab reader ensures that, if an item is found to be present in the slab, it can continue to be read for the duration of the lifetime of the reader.  _But_, it's possible for that same item to no longer appear "present" when queried by the same reader.  In `get_guarded()` I mistakenly relied on the latter being true when it is in fact not.

So the solution here is to introduce a new reader method `get_unchecked()`, which simply assumes an item is present.  It can only safely be called if the caller has previously demonstrated that an item was once present via `get()` or the new `contains_key()` (which then implies the item is valid for the remainder of the reader's lifetime, even if the item is removed), and is thus marked `unsafe`.

(I tried to think of an efficient way to do this without introducing an unsafe method but couldn't.  Rust doesn't like storing references derived from other references in structs even if you store the two together.)

So far, no crashes on the OCI test rig.  Will let it simmer for a few days.

PR #960, opened 2025-08-18
fix bug in crossbeam-epoch based RcuBox::get()

cpacejo:
This was a similar bug as https://github.com/org-zpr/zpr-core/issues/744 (fix https://github.com/org-zpr/zpr-core/pull/800).  Namely, For the crossbeam-epoch-based RCU implementation, `RcuBox::get()` was deferring the actual load of the pointer until the deref of the `RcuGuard`.  This meant that the "value" of the `RcuGuard` (i.e., the pointer value) could change between dereferences.  This resulted in the crash in `RcuOptionGuard::deref()`, which assumes that the value stays constant.

The solution is simply to load the pointer in `RcuBox::get()`, leaving `RcuGuard::deref()` to simply dereference that pointer.  Unfortunately this needs some unsafe pointer magic as Rust doesn't allow returning a thing (the `epoch::Guard`) and a reference to that thing (the `epoch::Shared`) together.

I don't have an easy way to reproduce the observed crash, but I did add a unit test which confirms that the behavior of `RcuBox::get()` is what we want (which did of course flag this bug).

The other implementations which compile (rcu-rwlock and rcu-mutex-arc) do not have this bug.

### DIFF
--- a/adapter/ph/src/rcu.rs
+++ b/adapter/ph/src/rcu.rs
@@ -167,13 +167,15 @@ mod rcu_impl {
 #[cfg(all(feature = "rcu-crossbeam-epoch", not(doc)))]
 mod rcu_impl {
     use crossbeam_epoch as epoch;
+    use std::marker::PhantomData;
     use std::sync::atomic::Ordering;
 
     pub struct RcuBox<T>(epoch::Atomic<T>);
 
     pub struct RcuGuard<'a, T> {
         guard: epoch::Guard,
-        atomic: &'a epoch::Atomic<T>,
+        ptr: *const T,
+        phantom: PhantomData<&'a T>,
     }
 
     impl<T> Drop for RcuGuard<'_, T> {
@@ -185,9 +187,9 @@ mod rcu_impl {
         type Target = T;
 
         fn deref(&self) -> &Self::Target {
-            let ptr = self.atomic.load(Ordering::Acquire, &self.guard);
+            // SAFETY: we know we have the associated guard for the lifetime of the returned ref
             // SAFETY: we only allow readers to access "live" values
-            unsafe { ptr.deref() }
+            unsafe { self.ptr.as_ref().unwrap_unchecked() }
         }
     }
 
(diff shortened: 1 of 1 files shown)


adapter/ph/src/rcu.rs
```diff
@@ -218,6 +223,17 @@ mod rcu_impl {
...
+                // We assume/hope `defer_destroy()` synchronizes-with
+                // `pin()`.  (They are not documented to do so, but I think
+                // it is a safe assumption.) So, in the case that the
+                // `pin()` of a `get()` occurs _after_ this step, but
+                // _before_ we implicitly unpin (and thus execute the
+                // destructor), the above swap is visible to the load in
+                // `get()`.  (Else, the load would see the pre-swap value,
+                // and a subsequent dereference thereof might occur after we
+                // have unpinned and thus triggered the destructor.)
                 guard.defer_destroy(old_value);
```
mkolehmainen, line 237: Just want to confirm that this destroy call here is OK given that someone else may have the old_value ptr inside an RcuGuard via a call to `get`.

adapter/ph/src/rcu.rs
```diff
@@ -167,13 +167,15 @@ mod rcu_impl {
...
+    use std::marker::PhantomData;
     use std::sync::atomic::Ordering;
 
     pub struct RcuBox<T>(epoch::Atomic<T>);
 
     pub struct RcuGuard<'a, T> {
         guard: epoch::Guard,
-        atomic: &'a epoch::Atomic<T>,
+        ptr: *const T,
+        phantom: PhantomData<&'a T>,
```
mkolehmainen, line 178: Does this want to be called `_phantom` ?

mkolehmainen (approved):
Looks good to me, but it looked good to me last time too.  I'm confident you have improved things.  A couple side comments in the review.

adapter/ph/src/rcu.rs
cpacejo: Yes!  `defer_destroy()` specifically just queues the destructor to run once all current epoch pins have been released.  Anyone who's looking at the value is doing it via `get()` or `inspect()` both of which hold a pin which outlives the lifetime of the reference to the object.

The two cases specifically are, the reader took the pin just before `defer_destroy()` got called but still read the old value (in which case, the destructor runs when the reader completes), or the reader took the pin just after `defer_destroy()` got called, in which case the reader is guaranteed to see the new value because `defer_destroy()` and `pin()` synchronize with each other, thus enforcing memory ordering between the store and load of the pointer.  (At least, the latter is something I believe to be true, but is not documented in the `epoch` library.)

adapter/ph/src/rcu.rs
cpacejo: It could I guess, does that give special meaning in Rust?  (I've never used that convention for struct members in Rust before.)

PR #961, opened 2025-08-18
changed command line arguments for logging and added option to change log in CLI

### LINKED ISSUE #925: Add a verbose logging option to enable "trace"

### LINKED ISSUE #947: add log level option in the ph-cli

owenrhysthomas:
still need to do some cleanup and fix the unit tests to reflect the new command line arguments used to do the logging.

### DIFF
--- a/adapter/cli/src/main.rs
+++ b/adapter/cli/src/main.rs
@@ -83,6 +83,8 @@ enum Commands {
     Capture(CaptureArgs),
     /// Change link state
     Link(LinkArgs),
+    /// Change the log level of a node or adapter
+    ChangeLogging(LogArgs),
     /// Exit the CLI
     Quit,
 }
@@ -135,6 +137,23 @@ enum LinkCommands {
     Stop { id: u32 },
     /// Reset a link.  It will require a configure before starting again
     Reset { id: u32 },
+}
+
+#[derive(Debug, Args)]
+#[command(flatten_help = true)]
+struct LogArgs {
+    #[command(subcommand)]
+    command: LogCommands,
+}
+
+#[derive(Debug, Subcommand)]
+enum LogCommands {
+    /// Show a link's status
+    Debug { debug: Vec<String> },
+    /// Configure a link
+    Trace { trace: bool },
+    /// Start a link
+    Quiet { quiet: Vec<String> },
 }
 
 fn main() -> std::io::Result<()> {
@@ -219,6 +238,7 @@ fn process_command(command: Commands, socket: &str) -> std::io::Result<bool> {
             frequency,
         } => handle_perf_sample(duration, frequency, &socket)?,
         Commands::Link(link) => handle_link_command(link, &socket)?,
+        Commands::ChangeLogging(log_args) => handle_change_log(log_args, &socket)?,
         Commands::Quit => return Ok(true),
     }
 
@@ -421,6 +441,23 @@ fn handle_link_command(link_args: LinkArgs, socket: &str) -> std::io::Result<()>
         LinkCommands::Stop { id } => basic_command!(RpcCommands::StopLink, socket, id)?,
         LinkCommands::Reset { id } => basic_command!(RpcCommands::ResetLink, socket, id)?,
     }
+
+    Ok(())
+}
+
+fn handle_change_log(log_args: LogArgs, socket: &str) -> std::io::Result<()> {
+    match log_args.command {
+        LogCommands::Debug { debug } => {
+            let debug_string = debug.join(" ");
+            println!("{}", debug_string);
+            basic_command!(RpcCommands::SetDebug, socket, debug_string)?
+        }
+        LogCommands::Trace { trace } => basic_command!(RpcCommands::SetTrace, socket, trace)?,
+        LogCommands::Quiet { quiet } => {
+            let quiet_string = quiet.join("_");
+            basic_command!(RpcCommands::SetQuiet, socket, quiet_string)?
+        }
+    };
 
     Ok(())
 }
--- a/adapter/cli/src/main.rs
+++ b/adapter/cli/src/main.rs
@@ -84,7 +84,10 @@ enum Commands {
     /// Change link state
     Link(LinkArgs),
     /// Change the log level of a node or adapter
-    ChangeLogging(LogArgs),
+    Logging {
+        #[arg(required = true)]
+        logs: String,
+    },
     /// Exit the CLI
     Quit,
 }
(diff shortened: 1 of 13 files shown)


owenrhysthomas:
The one downside to this PR is that it removes some of the requirements for the clap input that were previously in place. Before, the user could only enter a valid target in the debug field,  now they can enter invalid targets, it will just have no effect

adapter/ph/src/admin_worker.rs
```diff
@@ -222,6 +223,33 @@ async fn handle_connection(asm: Arc<Assembly>, mut stream: UnixStream) -> std::i
...
+                        match key_val.len() {
+                            2 => {
+                                asm.logging
+                                    .lock()
+                                    .unwrap()
+                                    .insert(key_val[0].to_string(), key_val[1].to_uppercase());
+                            }
+                            _ => {
+                                buf_writer
+                                    .write_all("ERR Unkno0wn log value ignored\n".as_bytes())
```
mkolehmainen: Stray zero in "unknown".

adapter/cli/src/main.rs
```diff
@@ -425,6 +431,18 @@ fn handle_link_command(link_args: LinkArgs, socket: &str) -> std::io::Result<()>
     Ok(())
 }
 
+fn handle_logging(vec: Vec<String>, socket: &str) -> std::io::Result<()> {
```
mkolehmainen, line 434: Is there any checking to see if this list of strings even makes sense to the logging subsystem?

adapter/ph/src/fastpath.rs
```diff
@@ -88,11 +88,13 @@ impl FastpathWorker {
         let adapter_manager = asm.adapter_manager_factory.make(&return_q);
         let mgmt_dispatch = asm.mgmt_dispatch_factory.make(&return_q);
         let batch_counters = Default::default();
-        asm.counters
-            .fastpaths
-            .lock()
-            .unwrap()
-            .push(FastpathCounters::default());
+
```
mkolehmainen, line 91: I wonder why this is showing up here?  This is already in main,.

adapter/ph/src/logging.rs
```diff
@@ -15,7 +15,7 @@ pub mod targets {
     // log target should be X, since the message is reporting on behalf of
     // an action in X.  (But if Y itself logs the result, it should do so
     // under target Y.) This helps provide a complete picture to someone
-    // debugging X why X is behaving in a certain way, without needing to
+    // debugging X why X is behaving in a   vbg way, without needing to
```
mkolehmainen: "vbg"?

adapter/ph/src/logging.rs
```diff
@@ -62,46 +61,67 @@ pub mod targets {
...
+        .unwrap();
 }
 
-pub fn initialize(config: &config::Config) {
-    tracing_subscriber::registry()
-        .with(fmt::layer())
-        .with(create_target_filter(&config.debug, &config.quiet))
-        .init();
+/// Gets the log level from a string
+fn get_level(level: &str) -> Level {
```
mkolehmainen, line 131: This is very permissive.  Not sure that's a problem.  But maybe note in comment that this will accept any garbage in and return Level::INFO in that case.

adapter/ph/src/main_args.rs
```diff
@@ -91,13 +90,9 @@ pub struct CommonArgs {
...
     pub zpr_addr: Option<Vec<IpAddr>>,
 
-    /// Enable debug logging for specified targets
-    #[arg(long, short = 'd', value_parser = clap::builder::PossibleValuesParser::new(ALL_TARGETS))]
-    pub debug: Vec<String>,
-
-    /// Disable info & warnings for specified targets
-    #[arg(long, short = 'q', value_parser = clap::builder::PossibleValuesParser::new(ALL_TARGETS))]
-    pub quiet: Vec<String>,
+    /// Set log level for various targets
```
mkolehmainen: We need more info here.  This is what shows up when user does `ph adapter -h` ... Include an example of the correct format or something like that.

The previous code printed out all the possible target values.  I think we still need that info.

mkolehmainen (changes requested):
Minor notes, but one thing I do think is important is more help on the command help page.

adapter/ph/src/fastpath.rs
owenrhysthomas: That's weird! I tried rebasing main onto my branch when that change was made - I must have done something a bit weird.

adapter/ph/src/main_args.rs
```diff
@@ -90,8 +91,20 @@ pub struct CommonArgs {
     #[arg(long, short = 'z')]
     pub zpr_addr: Option<Vec<IpAddr>>,
 
-    /// Set log level for various targets
-    #[arg(long, short = 'l', value_delimiter = ' ', num_args = 1.., value_parser = parse_key_val)]
+    /// Set log level using key value pairs: <target>=<LEVEL>
+    /// The options for targets are:
+    ///     all, capture, datapath, flow_mgmt, link_state,
```
mkolehmainen, line 96: Perfect!

adapter/ph/src/logging.rs
```diff
@@ -61,6 +60,18 @@ pub mod targets {
...
+pub mod levels {
+
+    pub const OFF: &str = "OFF";
+    pub const ERROR: &str = "ERROR";
+    pub const WARN: &str = "WARN";
+    pub const INFO: &str = "INFO";
+    pub const DEBUG: &str = "DEBUG";
+    pub const TRACE: &str = "TRACE";
+
+    pub const ALL_LEVELS: &[&str] = &[ERROR, WARN, INFO, DEBUG, TRACE];
```
mkolehmainen: Just confirming you intended to leave `OFF` out of this list.

mkolehmainen (dismissed):
Looks good to me!

adapter/ph/src/logging.rs
owenrhysthomas: Good point, I missed that. With the way it's set up now, there is actually no way for the logging to be completely off, because there is no `Level::OFF` in `tracing::Level`, so Level::ERROR is the lowest they can have now, but I left it in in case we wanted to slightly restructure in the future and add a way to not even create the tracing_subscriber.

owenrhysthomas:
@mkolehmainen sorry, just need your review again since I pushed again since your last review. All that was changed was adding OFF to ALL_LEVELS

mkolehmainen (approved):
Great!  You may want to get  feedback from @svjaffer too before merging.

svjaffer (approved):

PR #968, opened 2025-08-20
added config to assembly

### LINKED ISSUE #952: add config to assembly
Originally came up when working on #947 because I needed the config in the assembly. I asked Chris and he suggested putting the whole config into the assembly. It came to be a lot of changes to the codebase because of lots of "partial moves" of the config that precede creating the assembly. It ended up seeming like it was out of the scope of that PR. I decided to just add the relevant portions of the config to the assembly for the PR, but it did seem like a good change in the long run.

Use a RwLock rather than a Mutex for this purpose (since almost all accesses of the elements of the config now are read, so not much performance would be lost with an RwLock, but it would with a Mutex)

owenrhysthomas:


adapter/ph/src/assembly.rs
```diff
@@ -98,10 +97,10 @@ pub struct Assembly {
     pub self_noise_keypair: Option<NoiseKeypair>,
     pub peer_noise_keypair: Option<NoiseKeypair>,
     pub certx: Option<KmCertExchange>,
-    pub bsauth: Option<RsaBootstrapAuth>,
-    pub rsauth: Option<OAuthRsa>,
     pub system_start_time: std::time::Instant,
     pub address_pool: std::sync::Mutex<Option<AddressPool>>, // Nodes only (and required for nodes)
+
+    pub config: std::sync::RwLock<config::Config>,
```
cpacejo: `RwLock` is fine but better would be an `RcuBox` (from rcu.rs).  RCU boxes have the benefit that writers do not block readers, so we can safely read their contents from the fastpath and update from mgmt without risking priority inversion (mgmt blocking fastpath).  I _think_ it shouldn't give you trouble but let me know if it does.

cpacejo (dismissed):
This looks good but can you try with an `RcuBox` instead of an `RwLock`?  That would allow us to access the config in the fastpath too.

adapter/ph/src/assembly.rs
owenrhysthomas: Oh, I thought that was the case for RwLock, I misremembered. I'll change that

cpacejo: RwLock does have the advantage over Mutex that readers don't conflict with each other, but the writer will still hold them at bay (necessarily, because there's only one "thing" under the lock), and vice-versa.

RCU mechanisms have the further advantage that writers don't block readers at all (or vice-versa!) because writers just swap in a new "version" of the "thing" in them, and readers just keep a reference to the version they were looking at (which later gets garbage collected), so no-one blocks anyone.  They also (depending on implementation) can be _very_ fast for the readers (as in, zero overhead).  So they're fantastic for things like network configuration that written rarely and read millions of times per second.

owenrhysthomas:
A downside of the config being in the assembly, is that with the RcuBox it is not the most straightforward (if possible) to write to a single element in a struct (at least from what I can see), without simply creating a new struct, changing one of the elements, then writing that new struct.

cpacejo:
> A downside of the config being in the assembly, is that with the RcuBox it is not the most straightforward (if possible) to write to a single element in a struct (at least from what I can see), without simply creating a new struct, changing one of the elements, then writing that new struct.

Yes that is 100% true!  It is the main downside of RCU performance-wise, but not a problem in this case since config updates are relatively rare.

Ergonomics-wise, the Rust `{ .. }` syntax helps with this, you should be be able to say e.g. `box.update(move |cfg| { field = newvalue .. cfg.clone() })`

svjaffer (approved):

PR #971, opened 2025-08-20
MSG_CONFIRM for echo response

### LINKED ISSUE #758: set MSG_CONFIRM in echo replies
This prevents the kernel from repeatedly querying via ARP our peer's HW address.  (We can also send it for other replies, but echo reply seems both necessary and sufficient.)

owenrhysthomas:


owenrhysthomas:
I updated the libc version to try and fix the macos version, that didn't end up being the solution, should I decrement it back?

cpacejo:
No need to roll back the libc version it's good we keep things vaguely up-to-date anyway.

adapter/ph/src/packet.rs
```diff
@@ -237,6 +240,19 @@ impl PacketMetadata {
...
+        match self.flags & flags::CONFIRM != 0 {
+            true => libc::MSG_CONFIRM,
+            false => 0,
+        }
+    }
+
+    #[cfg(not(target_os = "linux"))]
+    pub fn get_confirm_flag(&self) -> libc::c_int {
+        0
+    }
```
cpacejo: I think these would be better placed in `fastpath_io`, since that's where we interface `Packet`s with OS syscalls.  (`Packet` is just a container type so shouldn't really "know" about OS syscall details.)

These can then just be private functions in `fastpath_io` (e.g. `fn get_confirm_flag(pkt: &Packet) -> libc::c_int`).

(& Note that `flags` is public so it can still be accessed directly from `fastpath_io`.)

cpacejo (changes requested):
This looks good thanks!  Only change is let's move the `get_confirm_flag()` method to keep `Packet` OS-agnostic.

cpacejo (approved):
Looks good thanks!

PR #972, opened 2025-08-22
created bash completions for CLI

### LINKED ISSUE #579: Generate CLI completion scripts
https://docs.rs/clap_complete/latest/clap_complete/

owenrhysthomas:
The README is not completely necessary and definitely not formatted well, would love feedback on that


owenrhysthomas:
This currently only generates a completion script in Bash, but could easily be expanded to create creation scripts for other shells

svjaffer (approved):

PR #976, opened 2025-08-25
updated error message in km_noise::tick

### LINKED ISSUE #908: Better error message if adapter is unable to initially contact node
Right now, if an adapter is pointed at a non-existent node, it eventually fails with a keying timeout error (after saying the tether was "successfully started").

We should produce an error message stating that we couldn't in fact contact the node at all, so users know what to address.

owenrhysthomas:
Here is the new error, the only thing I changed was adding the portion about node-adapter connection failure.

image

I'm not quite sure if this was what you had in mind. I don't know if there is a way to tell if the node actually never made a successful connection to the node versus a node that made a connection and then lost it.


mkolehmainen (approved):
Great! That's helpful!

PR #978, opened 2025-08-26
Create generation command in makefile, add generation for all supported shells

### LINKED ISSUE #977: create all completion scripts for CLI and put generation in makefile

owenrhysthomas:


owenrhysthomas:
I looked into automatically installing the completions, but I couldn't find a way for the Makefile to detect the shell the user is running, it's built in SHELL variable only detects the shell of the Makefile itself. I then thought I could just make individual commands for each shell, but not all the shells had straightforward ways of installing completions (with some requiring restarting the shell) so I wanted feedback on that aspect of the PR before I moved forward with that. It is worth doing its own command then for example printing to the user that they have to restart the shell to complete the installation?

mkolehmainen:
I feel like in other projects I've seen there is just a note saying, eg "to install command line completions add xyz to your profile".  I think we can leave it to users to add the completions if they want them.

owenrhysthomas:
@mkolehmainen Ok that sounds good, I'll add some phrasing in the makefile

adapter/cli/src/main.rs
```diff
@@ -479,15 +478,36 @@ fn serialize(program: &str) -> String {
...
 
-fn generate_completion<W: std::io::Write>(mut writer: BufWriter<W>) {
-    generate(
-        Bash,
-        &mut CmdlineArgs::command(),
-        CmdlineArgs::command().get_name().to_string(),
-        &mut writer,
-    );
+fn generate_completion(path: String) -> std::io::Result<()> {
+    let exts: Vec<&str> = Vec::from(["sh", "elv", "fish", "ps1", "zsh"]);
```
mkolehmainen: No big deal, but why not just make a list of the Shell::xxx things here instead of these extensions? It looks like you just use these extensions to look up a Shell::xxx.

mkolehmainen (dismissed):
Looks good to me! One minor comment.

adapter/cli/src/main.rs
owenrhysthomas: Good point. I initially wanted to make an enum map of both so I could have both, but the Shell enum doesn't support an enum map, so I just made an array, but I should've just made an array of tuples

owenrhysthomas:
> > I looked into automatically installing the completions, but I couldn't find a way for the Makefile to detect the shell the user is running, it's built in SHELL variable only detects the shell of the Makefile itself. I then thought I could just make individual commands for each shell, but not all the shells had straightforward ways of installing completions (with some requiring restarting the shell) so I wanted feedback on that aspect of the PR before I moved forward with that. It is worth doing its own command then for example printing to the user that they have to restart the shell to complete the installation?
> 
> This probably won't work for windows, but should work for linux/mac: you can access the `${SHELL}` environment variable from the makefile. @cpacejo might have more thoughts on best practices about this.

I talked to Chris and we don't think it's possible to detect the shell within the Makefile. As I said in my original comment, the SHELL variable detects the shell that Make is using, which is set to bash by default. You can change the SHELL variable yourself and have Make use a different shell, but neither Chris or I could find a way to use it to detect the shell that Make is being run on due to it being set by default. Chris said it is unique among environment variables in that way.

svjaffer:
> I talked to Chris and we don't think it's possible to detect the shell within the Makefile. As I said in my original comment, the SHELL variable detects the shell that Make is using, which is set to bash by default. You can change the SHELL variable yourself and have Make use a different shell, but neither Chris or I could find a way to use it to detect the shell that Make is being run on due to it being set by default. Chris said it is unique among environment variables in that way.

I'm surprised by this.  The Make documentation says that the `SHELL` variable is set by default to whatever `/bin/sh` is, which is usually bash, but can be changed by the user to change their default shell.  The Makefile itself can set a different shell, which will override that, but we're in control of that and can choose not to do that.

Anyway, not a big deal.  I'm probably just misunderstanding something.

svjaffer (approved):

PR #982, opened 2025-08-27
generate completion scripts for node and adapter

### LINKED ISSUE #973: Completion scripts for node and adapter
Generated completion scripts for CLI in #579, could also be useful for the node and adapter and is pretty straightforward. Probably not the most urgent since we are mostly using the integration tests at this time, but could be helpful in the future.

A good example of using completion scripts with the clap Derive API (could also see our CLI): https://github.com/clap-rs/clap/blob/master/clap_complete/examples/completion-derive.rs

owenrhysthomas:


adapter/ph/src/main_args.rs
```diff
@@ -110,6 +110,10 @@ pub struct CommonArgs {
     /// Which packet I/O engine to use
     #[arg(long, value_parser = clap::builder::PossibleValuesParser::new(std::iter::once(batch_io::AUTO_ENGINE_NAME).chain(batch_io::engine_names())), default_value_t = batch_io::AUTO_ENGINE_NAME.to_owned())]
     pub io_engine: String,
+
+    // Path to the generations file you want to create
+    #[arg(long, short = 'g')]
+    pub generate: Option<String>,
```
cpacejo: Rather than add this as an argument to the existing binary, let's create a new binary whose sole job is to create the bindings.  (This maintains separation of concerns and also avoids linking in `clap_complete` into the main binary.)  You should be able to simply add a new file under `src/bin/`.  Then you can add `clap_complete` as a dependency _only_ of this new binary by adding a `[[bin]]` section as described [here](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#binaries).

adapter/ph/completions/ph-cli.elv
cpacejo, line 1: My default stance is not to check these generated files in (else it's easy to forget about them and they become outdated).  I think we'll just (eventually) want to autogenerate them when building a release package.  Otherwise I think it's fair to ask developers to just run `make` to create these as needed.

cpacejo (changes requested):

cpacejo (approved):
Looks great thanks!

PR #997, opened 2025-09-04
remove local zpr addresses from assembly

### LINKED ISSUE #993: remove local_zpr_addresses from assembly
final variable that is in both the config and separately in the assembly in its own RcuBox

owenrhysthomas:


cpacejo (approved):
Hum… I'm still torn whether the `config` should only ever reflect user-specified values (and the assembly should have separate fields for settings pushed down by the node / visa server).  E.g. considering a future world where we allow the user to update the config via the debug tool… I think we wouldn't want changes to that to override the values sent by the node.

I think this change is fine on technical merits tho, if we ever do have an issue with the above we can revisit it so I'm fine with checking it in!

PR #999, opened 2025-09-05
pid goes in temp file

### LINKED ISSUE #990: [BLOCKED] Use temp file for Visa Service PID file in integ tests
The integration test tries writing the Visa Service PID file into ~/.local which would conflict with any other visa service running on the machine.  Let's instead specify a temp file in the integration tests.

Blocked on:
* org-zpr/zpr-visaservice/issues/28

owenrhysthomas:


.github/workflows/adapter.yml
```diff
@@ -30,7 +30,7 @@ concurrency:
 env:
   CARGO_TERM_COLOR: always
   CARGO_NET_GIT_FETCH_WITH_CLI: true
-  VISA_SERVICE_RELEASE: v0.5.0
+  VISA_SERVICE_RELEASE: v0.6.0
```
owenrhysthomas, line 34: I rebased onto a branch in another PR #1003 so that this PR would run successfully. I expected that when I merged that branch into main, the diff for this branch would reflect the updated main, but it seems it is still diffing with the main from before the merge? Strange

svjaffer: No, because of how we're doing the squash and merge.  The history no longer matches.  If you want to avoid that problem you need to do some history-rewriting in this branch to remove the redundant commits.

svjaffer (approved):

.github/workflows/adapter.yml
owenrhysthomas: Ah I thought it might be something to do with the squashing.

PR #1006, opened 2025-09-09
fix integ test error handling multiple fastpath counter blocks

cpacejo:
I think splitting the fastpath counters into multiple groups broke this, easy fix!


owenrhysthomas (approved):
Great!

PR #1010, opened 2025-09-11
Move internal workflows to v1

svjaffer:
Use version numbers for official GH actions instead of hashes


mkolehmainen (approved):
I like this a lot!

PR #1012, opened 2025-09-12
null encryption subseet of noise handshake

### LINKED ISSUE #998: add option to enable+force NULL encryption for debug

owenrhysthomas:


owenrhysthomas:
I ended up changing approach in this PR because I realized that the noise protocol is more integral to the functioning of the KM than I had initially thought. As I discovered, and Mathias pointed out, the CN is passed through the peer cert. I thought about creating a paired down handshake that just passed a CN, but the message sending in the handshake of the noise protocol uses snow, so I was unsure how to approach creating a similar protocol with another method, which I assume I would need thrift or cap'n proto if we switch to that. This solution -- just switching out the decrypt and encrypt functions -- seemed like a good temporary solution until I could implement a more robust and permanent one.

adapter/ph/src/km_multiplexor.rs
```diff
@@ -277,10 +286,10 @@ fn add_noise_link(
     let spawn_sig_tx = asm.km_state.km_sig_tx.clone();
 
     let (km_tx, km_rx) = mpsc::channel(asm.topology_config.km_link_queue_size);
-
+    let km_impl = asm.config.get().km_impl.clone();
```
cpacejo: `km_impl` is just an integer right?  I think the `clone()` shouldn't be needed in that case?

adapter/ph/src/mgmt/dispatch.rs
```diff
@@ -126,37 +126,41 @@ fn handle_key_management(asm: &Arc<Assembly>, pkt: &mut Packet) {
...
-            "KeyManagement packet not using NOISE - type is {}",
-            km_hdr.message_type
-        );
-        core::count_event(asm, pkt, ManagementCounterType::OtherError);
-        return;
-    }
+    // TOOD change this
+    if (km_hdr.is_noise() && asm.config.get().km_impl == zpr::KM_ID_NOISE)
+        || (km_hdr.is_null() && asm.config.get().km_impl == zpr::KM_ID_NULL)
+    {
```
cpacejo: `TOOD` → `TODO` so `grep` can find this later :)

Also if you invert the condition (just throw a `!(...)` around it) you could leave this in the order "early error return // happy path".  Stylistically I find that helps prevent the happy path from becoming overly indented.

adapter/ph/src/km_noise.rs
```diff
@@ -358,6 +358,45 @@ impl Codec for NoiseCodec {
...
+        self: &Self,
+        payload: &[u8],
+        message: &mut [u8],
+    ) -> Result<usize, EncryptionError> {
+        if message.len() < payload.len() {
+            return Err(EncryptionError::ParseError);
+        }
+        for i in 0..payload.len() {
+            message[i] = payload[i];
+        }
```
cpacejo: `message.copy_from_slice(payload)` is a shorter way to write this :)  (same for `decrypt` below)

cpacejo (dismissed):
Awesome!  Just a few style nits

adapter/ph/src/mgmt/dispatch.rs
owenrhysthomas: I just removed the TODO because I actually had just put it in for myself during development and had made the change I had meant to

cpacejo (approved):
Perfect!

PR #1013, opened 2025-09-15
sigint vservice instead of sigterm

### LINKED ISSUE #1000: vservice.pid does not get deleted in integration test
It seems to me that somehow the Remove function on the PidFIle struct is not getting called, despite being called with `defer`, meaning it should supposedly be called regardless of program exit conditions

not sure if it does not get deleted when running visa service independently from one-node-test - requires more investigation

owenrhysthomas:
Found that vservice cleanly exits with SIGINT ([prototype README](https://github.com/org-zpr/zpr-prototype?tab=readme-ov-file#to-shut-things-down)), so changed how the vservice is killed


cpacejo:
Ah good find!  This is fine but I think a simpler fix is to simply use SIGINT across the board (the PH treats the first SIGINT same as a SIGTERM anyway).

owenrhysthomas:
> Ah good find! This is fine but I think a simpler fix is to simply use SIGINT across the board (the PH treats the first SIGINT same as a SIGTERM anyway).

Ah, I wasn't sure if that was the case, so I erred on the side of caution, but will make that change.

cpacejo (approved):
Perfect thank you!

PR #1015, opened 2025-09-17
create separate binary for generating completions in cli

### LINKED ISSUE #981: factor out shared code between completion scripts in cli and ph/adapter
shared func `generate_completion`

also convert CLI generation to separate binary to align with adapter/ph methodology

owenrhysthomas:


owenrhysthomas:
This change means there is alignment between the method of generating completions between the ph and the ph-cli. Also does a bit of cleaning of the CLI by factoring out the structs for the command line args.

owenrhysthomas:
I initially wanted to factor out the shared code between the two generate files, especially now that there are two almost identical files. I thought about putting it in the ZPR lib, but then I realized it didn't make sense because this is not a ZPR concept. It seems like a lot to make a whole lib just for the one function, but I'm not sure, interested in input.

cpacejo (approved):
Nice looks good!

PR #1020, opened 2025-09-24
implement ZDPR logic

cpacejo:
This is the core ZDPR state machine.  It manages the lifetime of each packet sent on a link.

Note this code does _not_ do anything "active": it does not send packets or start timers.  It expects the caller to handle those things.  (That code comes later!)

Documentation is inline but please let me know any questions!

More tests are forthcoming.

### DIFF
--- a/adapter/ph/src/zdpr.rs
+++ b/adapter/ph/src/zdpr.rs
@@ -1,0 +1,709 @@
+#![allow(dead_code)]
+
+//! Reliable ZDP logic.
+//!
+//! `Sender` and `Receiver` encapsulate the sending and receiving logic
+//! respectively of a reliable ZDP session.  Each side of a ZDP link will
+//! have both its own `Sender` and `Receiver` state corresponding to the two
+//! directions of traffic.
+//!
+//! These structures encapsulate only the logic involved in maintaining
+//! session state.  The mechanics of sending/receiving packets and managing
+//! timeouts live elsewhere.
+
+use enum_map::{Enum, EnumMap};
+use std::collections::VecDeque;
+use std::task::{Context, Poll, Waker};
+use zpr::SeqNum;
+
+/// Reify the truncated sequence number into an offset relative to the
+/// reference sequence number, under the assumption that it is within a
+/// window centered on the highest seen value thus far.
+fn reify_seq_num_relative(reference: SeqNum, sn: u16) -> i64 {
+    // We operate under the assumption that the difference between the
+    // true sequence number and `reference` is in the range [-2^15, 2^15).
+
+    // Under that assumption, we can subtract the truncated versions
+    // of both to produce a 16-bit 2s-complement value representing
+    // this difference.
+    let diff = sn.wrapping_sub(reference as u16);
+
+    // Convert the 16-bit 2s-complement value into a 64-bit signed value.
+    (diff as i16) as i64
+}
+
+/// Truncate a sequence number to 16 bits for sending over the ether.
+pub fn truncate_seq_num(sn: SeqNum) -> u16 {
+    (sn & 0xFFFF) as u16
+}
+
+/// ID of a packet which was queued for sending (but maybe not yet sent).
+pub type QueuedPacketId = u64;
+
+pub enum EnqueueResult<'a, Pkt> {
+    Sent(SeqNum, &'a mut Pkt),
+    Queued(QueuedPacketId),
+}
+
+/// State of the sending side of a reliable ZDP session.
+///
+/// `Pkt` is any abstract representation of a packet which the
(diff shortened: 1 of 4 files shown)


adapter/ph/src/zdpr.rs
```diff
@@ -0,0 +1,709 @@
...
+
+use enum_map::{Enum, EnumMap};
+use std::collections::VecDeque;
+use std::task::{Context, Poll, Waker};
+use zpr::SeqNum;
+
+/// Reify the truncated sequence number into an offset relative to the
+/// reference sequence number, under the assumption that it is within a
+/// window centered on the highest seen value thus far.
+fn reify_seq_num_relative(reference: SeqNum, sn: u16) -> i64 {
```
mkolehmainen, line 22: This I don't fully comprehend, but I think will be clearer when I can see SeqNum.

adapter/ph/src/zdpr.rs
```diff
@@ -0,0 +1,709 @@
...
+    /// upon receiving a packet).  We represent it anyway to simplify the implementation.
+    recvd: WindowBitset,
+
+    /// Receiver statistics.
+    stats: EnumMap<ReceiverStat, u64>,
+}
+
+impl Receiver {
+    /// The maximum window size the receiver can be configured with.
+    pub const MAX_WINDOW_SIZE: usize = WindowBitset::BITS as usize;
```
mkolehmainen, line 608: I assume this means max window size in receiver is 64?  In sender is usize::MAX.... those are very different values.  What's the expected typical window size?

adapter/ph/src/zdpr.rs
```diff
@@ -0,0 +1,709 @@
...
+
+    fn oldest_unrecvd_offset(&self) -> i64 {
+        -(WindowBitset::BITS as i64) + (self.recvd.leading_ones() as i64) + 1
+    }
+
+    /// Process the full sequence number of a received packet.
+    ///
+    /// The return value indicates whether and how to process
+    /// the associated packet.
+    pub fn process_packet(&mut self, sn: SeqNum) -> Disposition {
```
mkolehmainen, line 642: I'm having a little trouble visualizing how these offsets work, but I believe you!!

mkolehmainen (approved):
Thanks for the comments!

adapter/ph/src/zdpr.rs
cpacejo: They're all referenced to the "newest" (rightmost/backmost) packet in the queues.  So offset 0 is the back of the queue.  Offset -1 is the second-to-last.  Etc.

(The reason is because I track the sequence number / packet ID of the backs of the queues rather than the fronts.  And the reason for that is because the "clean up" code which removes dead entries from the fronts of the queues doesn't also have to fixup these pointers.)

adapter/ph/src/zdpr.rs
cpacejo: Correct!  These values are limits of these particular implementations.  I use 32 as the default (in `config`).

adapter/ph/src/zdpr.rs
cpacejo: Essentially on the wire to save bits we only send the rightmost 16 bits of the 64-bit sequence number.  But that means we need to "reconstitute" the leftmost 48 bits on the other end of the wire.  So we fill those in such that the resulting sequence number is within ±2^15 of our notion of the "current" sequence number.

As an example say the "current" sequence number is 0x12345678.  And we want to send packet 0x12345675 (say it's a retransmit or ack).  We'd send it with truncated sequence number 0x5675 which would get reconstituted as 0x12345678.  Likewise if we sent packet 0x12345679 (say it's the next packet) we'd send 0x5679 which would get reconstituted as 0x12345679.

This all works even if our notion of "current" sequence number has advanced a little bit because our active window is orders of magnitude smaller than 2^15.

The tricky cases are "very" old or new sequence numbers relative to our "current" sequence number.  E.g. 0x1233D678.  This is -0x8000 behind the current SN 0x12345678.  We'd send 0xD678 and reconstitute it correctly as 0x1233D678.  Likewise 0x1234D677 is +0x7FFF ahead of the current SN.  We'd send 0xD677 and reconstitute it correctly.  But 0x1233D677 is -0x8001 behind… we'd send 0xD677 which would get incorrectly reconstituted as 0x1234D677.

Which REMINDS me… we need to include the full/reconstituted sequence number in the checksum, not the truncated sequence number.  Because that's exactly what prevents us from processing "really old" packets which got stuck on the wire/retransmitted for some reason.  32k packets is not that many if we're sending them quickly.  We can run through those in 1 second with an RTT of 1 ms and window size of 32 which is within the lifespan of a wayward packet.  Will address that in a future commit.

PR #1021, opened 2025-09-25
ZDPR futures (and a couple bugfixes)

cpacejo:
These futures allow mgmt code to asynchronously wait on the stages of a ZDPR packet.  Requesting a mgmt packet to be sent (e.g. via `send_non_flow_mgmt()`) will return a `Sent` future, representing that the packet has been entered into the ZDPR machinery.

The caller can do one of four things with `Sent`.  Either:
1) `.await` it.  This completes only when the packet has made its way into the send window (and therefore actually sent on the wire).  This enables callers to participate in basic flow control by preventing the caller from queueing up more than a single packet while the window is full.  On completion, this returns an `Acked` future (see below).
2) Drop the future, or `.try_cancel()` it.  If the packet is currently in the blocked queue, it is cancelled (and returned in the case of `.try_cancel()`).  This allows callers to choose to no longer wait in the blocked queue (e.g. because of a timeout).  (If the packet was already sent, an `Acked` future is returned instead.)
3) `.enqueue()` it.  This returns immediately, leaving the packet on the blocked queue (if it was still/ever there).  This allows non-async code (such as the link state machine) to send mgmt packets, in exchange for taking on the burden of ensuring that they do not flood the blocked queue.
4) Call `.acked()` to transform this into a `SentAndAcked` future (see below) which waits for the packet to be acknowledged.

When `Sent` completes, an `Acked` future is returned.  (Or, an error is returned if the link died.)   This future can only be `.await`ed on, indicating that the remote peer acknowledged receipt of the packet.  (Dropping an `Acked` does nothing; the packet continues to participate in ZDPR until it is acknowledged.)

A `SentAndAcked` future (as returned by `Sent::acked()`) combines the stages of waiting for sending and acking into one.  This is helpful in contexts such as `select!()` statements which cannot handle the "chained" futures of Sent->Acked.  It provides the same affordances as a `Sent` future, except that completion indicates the packet has been acknowledged by the peer.  (Notably, we'll use this for echo requests.)

An upcoming PR will switch over the `mgmt::core` methods such as `send_non_flow_mgmt()` to use ZDPR and return these futures.  For now they are just dead code.

------

ALSO, this PR contains a couple bugfixes to `zdpr`.  First, cancellation is fixed not to remove the packet from the sent queue (this was an optimization I had made which breaks an assumption used by these futures, namely that we can continue to refer to a packet which we tried and failed to cancel).  Second, there was an arithmetic error in packet ID -> sequence number lookup.  (This is my reminder to get unit tests in place!)

### DIFF
--- a/adapter/ph/src/mgmt/core.rs
+++ b/adapter/ph/src/mgmt/core.rs
@@ -10,6 +10,7 @@ use crate::packet::Packet;
 use crate::seq_nums;
 use crate::zdp;
 use crate::zdpr;
+use std::task::{Context, Poll};
 use thiserror::Error;
 use tokio::time::sleep;
 use tracing::*;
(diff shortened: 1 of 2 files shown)


adapter/ph/src/mgmt/core.rs
```diff
@@ -85,6 +87,191 @@ pub fn send_per_flow_mgmt_response(
...
+    packet_id: PacketId,
+    // NOTE: if we add anything without a trivial Drop implementation,
+    // modify `enqueue()` appropriately!
+}
+
+impl<'a> Sent<'a> {
+    /// If sending would block, queue this packet instead/
+    pub fn enqueue(self) {
+        std::mem::forget(self)
+    }
```
mkolehmainen, line 155: Do we really want a pub function that just calls forget? How does preventing the drop from running manage to enqueue the packet?

mkolehmainen (dismissed):
Neato!

adapter/ph/src/mgmt/core.rs
cpacejo: The trick is, the packet is already enqueued :) The drop handler actually _dequeues_ the packet.

(I couldn't think of a better name for this method.  But I do think of this as handing ownership/lifetime of the packet over to the queue, so to me it reflects the intention of calling this method.  It's clear that calling this has the potential to build the queue in an unbounded manner.)

I'll clarify with some comments.

cpacejo: (Also note that `Sent` has no members allocated on the heap or with their own Drop implementations.  So `forget()` here doesn't leak anything; the only effect is not to run our own custom Drop implementation.)

mkolehmainen: Ah ok!

mkolehmainen (approved):
LGTM!

PR #1022, opened 2025-09-25
ZDPR Acknowledgement packet type

cpacejo:
Adds the Acknowledgement packet type needed by ZDPR.  Also adds the notion of a "tiny packet" which we use for sending Acknowledgements (since they are so tiny), and a "small packet" which we use for all other management packets (since they are not large), in order to reduce pressure on the memory allocator of allocating 3 pages for every management packet.


owenrhysthomas (approved):
Cool!

PR #1023, opened 2025-09-25
ZDPR retries

cpacejo:
Here is the first "active" part of ZDPR.  It is a per-link worker whose sole job is to retry unacknowledged ZDPR packets.

Its logic is very simple.  Every 400 ms, if there are any unacknowledged packets in the ZDPR sender state, it retransmits them.

When notified by the ZDPR retry-timer-reset `Notify`, it resets the timer to execute in 400 ms (and every 400 ms repeatedly after that).  This notification will be sent (in a future PR) whenever the ZDPR sender state transitions from having "no" unacknowledged packets to "some" unacknowledged packets.  The net effect is that the first packet sent in a "burst" starts the timer, which runs continually until there are no outstanding unacknowledged packets.

This matches the behavior of the [Reliable UDP](https://datatracker.ietf.org/doc/html/draft-ietf-sigtran-reliable-udp-00/) proposal.

### DIFF
--- a/adapter/ph/src/assembly.rs
+++ b/adapter/ph/src/assembly.rs
@@ -23,6 +23,7 @@ use crate::tun_ctl::TunCtl;
 use crate::visa_table;
 use crate::vs_types::AuthServicesList;
 use crate::zdp::TerminateReason;
+use crate::zdpr_worker;
 use km_noise::NoiseKeypair;
 use std::collections::HashMap;
 use std::net::IpAddr;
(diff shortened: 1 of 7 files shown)


svjaffer (approved):
I have one real concern about the code here, which is that build_and_egress_packets doesn't count the number of packets dropped due to the `MgmtSubstrateEgress` being full.  I understand that you are treating it as if the substrate dropped the packet, but is there a good reason not to have a counter to indicate that this happened?

cpacejo:
Oh good idea!  I'll add such a counter in a new PR.

PR #1024, opened 2025-09-26
counters in two columns!

mkolehmainen:
Save some vertical screen space.

### DIFF
--- a/adapter/ph/src/signal_worker.rs
+++ b/adapter/ph/src/signal_worker.rs
@@ -3,6 +3,7 @@
 use crate::assembly::Assembly;
 use crate::counters::*;
 use crate::logging::targets::STARTUP;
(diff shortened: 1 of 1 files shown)


cpacejo (changes requested):
Nice!  I have one concern, which is that this makes automated parsing (e.g. with awk) a bit more complicated.  Could you add a "|" character between the two columns, so that a `tr '|' '\n'` can be used to break into one stat per line?

The integration test as an example parses this, in the section "Check stats".  So you'll need to adjust that as well (the `tr` trick above should work fine if you make the above change).

mkolehmainen:
Ah I did wonder if anything was reading those.

mkolehmainen:
Since turns out the output is consumed by at least one test, I'm going to leave this not done for now.

PR #1027, opened 2025-09-29
created .capnp file with specification for protocol

owenrhysthomas:


cli-proto/cli.capnp
```diff
@@ -0,0 +1,29 @@
+@0xdd33f90089b3556f;
+
+# I considered making success also take text, for future extensibility, but
+# none of the current commands need info on success
```
cpacejo, line 4: In that case we can always add a new "success" type.  Or genericize this one as `SuccessOrError(SuccessValue, ErrorValue)` (which is explicitly backward-compatible in Cap'n Proto).  I think this is fine for now though.

cli-proto/cli.capnp
```diff
@@ -0,0 +1,29 @@
...
+interface CmdLineInter {
+    echo @0 () -> ();
+    resetCounters @1 () -> ();
+    counters @2 () -> (counts: Text);
+    setCaptureFile @3 (filePath: Text) -> (result: SuccessOrError);
+    closeCaptureFile @4 () -> ();
+    flushCaptureFile @5 () -> ();
+    setCaptureProgram @6 (program: Text) -> (result: SuccessOrError);
+    deleteCaptureProgram @7 () -> ();
+    perfSample @8 (duration: UInt64, frequency: UInt64) -> (result: Text);
```
cpacejo, line 21: Let's suffix the argument names with the unit of time they represent (e.g. "_ms" if they're milliseconds or "_s" if they're seconds).

cli-proto/cli.capnp
```diff
@@ -0,0 +1,29 @@
...
+    showLinkSummary @9 () -> (summary: Text);
+    showLink @10 (id: UInt32) -> (result: Text);
+    configureLink @11 (id: UInt32) -> (); # Currently unsupported in PH
+    startLink @12 (id: UInt32) -> (result: SuccessOrError);
+    stopLink @13 (id: UInt32) -> (restult: SuccessOrError);
```
cpacejo, line 26: Typo: `restult`

cli-proto/cli.capnp
```diff
@@ -0,0 +1,29 @@
...
+    resetLink @14 (id: UInt32) -> ();
+    changeLogging @15 (logs: Text) -> (result: SuccessOrError);
```
cpacejo, line 28: This is fine for now (keeping these 1:1 with what's currently there).  After we get everything working let's consider making `logs` a richer type (e.g. a list or map).

Same for `counters` above, I think that could be later changed to return a list/map of counters.  (But again I'd keep it as-is for now.)

owenrhysthomas: Yes, my plan was to get all this working first, then go back through the server and do better formatting for the result to take better advantage of the fact that we don't just have to use strings.

cpacejo:
Looks good thanks for posting!  Just a couple naming nits and future design suggestions.

Feel free to make a new repo (maybe call it `zpr-admin-api`? since it corresponds to the `admin_worker`) and push this into it.  You can follow https://github.com/org-zpr/zpr-policy and/or https://github.com/org-zpr/zpr-vsapi as examples.

PR #1029, opened 2025-09-29
no more seq nums, add transaction IDs

cpacejo:
Gets rid of sequence numbers from our internal API.

There are two commits here.  The first eliminates sequence numbers from the API, the second eliminates the sequence number generator and tracker.

This leaves us briefly vulnerable to duplicate packets, and out-of-order echo responses.  These issues will be remedied by an upcoming PR which enables ZDPR.

For request/response pairs (right now, just Bind Request), we add a Transaction ID header (part of the latest ZDP spec), which is used to pair requests with responses.  For now, some parts of that mechanism are stubbed out pending an upcoming PR which reworks the sync request mechanism in the face of ZDPR.  (It still functions though.)

There's now a new field in packet metadata which holds the sequence number for debug purposes.  It's set to 0 pending the PR which activates ZDPR.

Key Management messages are excepted from all this processing.


mkolehmainen (approved):
Looks good!

PR #1032, opened 2025-10-03
switch to ZDPR

cpacejo:
Switch to using ZDPR to send packets, and process received packets with ZDPR.


mkolehmainen (approved):
Let's do this!

PR #1035, opened 2025-10-06
get rid of unused pkt argument to count_event()

cpacejo:
We don't use this argument and it makes an upcoming PR clunky (preventing easy use of the `?` operator to propagate send failures).


owenrhysthomas (approved):
🤩

PR #1036, opened 2025-10-06
ZDPR futures

cpacejo:
Now we can finally use the futures we built as returns from calls to the core mgmt send functions.

This change does that, and litters the rest of the codebase with `.await` and `.enqueue()` calls as appropriate to the context (async or non-async).

Also I replaced the unit error type of the futures with a new `MgmtSendResult` which allows for easy error propagation via `?`.  (The only error which can occur is link closure.)

Also removed a misleading comment about memory order races which can not in fact occur.

### DIFF
--- a/adapter/ph/src/mgmt/core.rs
+++ b/adapter/ph/src/mgmt/core.rs
@@ -109,6 +109,10 @@ pub fn send_acknowledgement(asm: &Assembly, link_id: zpr::LinkId, sequence_numbe
         .try_enqueue_packet(link_id, &mut packet);
 }
 
+pub enum MgmtSendError {
+    LinkClosed,
+}
+
 #[allow(dead_code)]
 enum PacketId {
     Sent(zpr::SeqNum),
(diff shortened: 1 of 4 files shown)


adapter/ph/src/mgmt/core.rs
```diff
@@ -326,7 +330,11 @@ fn send_mgmt_helper(
     hdr.packet_type = packet_type;
 
     let Some(peer_state) = asm.peer_table.get(link_id) else {
-        return;
+        return Sent {
```
mkolehmainen, line 333: Seems odd to return a "Sent" here.  But maybe this just helps to clean up the allocation?

mkolehmainen (approved):
Looks good!

adapter/ph/src/mgmt/core.rs
cpacejo: It's a `Sent` which immediately resolves as an error.  This avoids having to make the return type `Result<Sent, MgmtSendError>`.  So the error path is the same whether the link is already gone, or if it disappeared while we were waiting in the queue.

PR #1037, opened 2025-10-07
Switched CLI to capnp protocol

### LINKED ISSUE #1026: convert CLI to use capnp

owenrhysthomas:
Still to do

- take more advantage of capnp - less reliance on Text type in capnp
-  FD in capnp
-  Get perf sample working again? \

Not sure if the way I added the capture socket to the config was the best - it probably should have been optional, but I decided to do it this way since it was quicker and this is a temporary fix. If you think it should be changed before this is merged, just let me know. 

~~NOTE this PR will not compile until~~
~~1. https://github.com/org-zpr/zpr-admin-api/pull/4 is approved~~
~~2. The corresponding PR in [org-zpr/zpr-admin-api-rs](https://github.com/org-zpr/zpr-admin-api-rs) is generated and merged~~
~~3. Cargo.toml for cli and ph are updated in this PR~~


adapter/cli/src/main.rs
```diff
@@ -4,53 +4,49 @@
...
-    };
-    ($comm:expr, $socket:ident, $arg1:tt, $arg2:tt) => {
-        basic_call_response_2($comm, $socket, $arg1, $arg2)
-    };
-}
+#[allow(dead_code)]
+const ANCILLARY_BUFFER_SIZE: usize = 128;
 
-fn main() -> std::io::Result<()> {
+#[tokio::main]
```
cpacejo, line 34: Since we're using Tokio here just for semantic reasons (and not performance reasons) it might be better to use the [current-thread runtime](https://docs.rs/tokio/latest/tokio/attr.main.html#current-thread-runtime) (and then we can drop the `rt-multi-thread` feature flag as well).

(Nothing wrong _per se_ with using the multi-thread runtime, it's just a lot of extra mechanism and code under the hood which the CLI won't benefit from.)

adapter/ph/src/config.rs
```diff
@@ -228,6 +230,16 @@ impl Config {
                 Err(e) => return Err(e),
             }
         }
+        if self.capture_path.to_str().unwrap().is_empty() {
```
cpacejo, line 233: I think instead here we can do `.as_os_str().is_empty()` (to avoid the potential panic on non-UTF-8 characters).

adapter/cli/src/main.rs
```diff
@@ -4,53 +4,49 @@
...
+            .await
+            .map(|_| {})
     } else {
-        run_cli(&socket)
+        run_cli(&socket, cap_socket).await
     }
 }
 
-fn run_cli(socket: &str) -> std::io::Result<()> {
+async fn run_cli(socket: &str, cap_socket: Option<String>) -> Result<(), capnp::Error> {
```
cpacejo, line 49: For consistency you could make `cap_socket` an `Option<&str>`, which when calling this function you can easily get one via `cap_socket.as_ref()`.  (Same for `parse_and_exec()` and `process_command()`.

adapter/cli/src/main.rs
```diff
@@ -4,53 +4,49 @@
...
-    ($comm:expr, $socket:ident, $arg1:tt, $arg2:tt) => {
-        basic_call_response_2($comm, $socket, $arg1, $arg2)
-    };
-}
+#[allow(dead_code)]
+const ANCILLARY_BUFFER_SIZE: usize = 128;
 
-fn main() -> std::io::Result<()> {
+#[tokio::main]
+async fn main() -> Result<(), capnp::Error> {
```
cpacejo, line 35: I see four different sources of errors: `capnp` itself, the OS (when opening a file), the CLI parsing code, and the RPC errors themselves.

Rather than repurpose `capnp::Error` for all of these, let's define our own error type.  Something like:
```
enum CliError {
  ProtocolError(capnp::Error),
  OsError(std::io::Error),
  ParseError(String),
  PeerError(String),
}
```

Then e.g. when checking for a no-such-file error, you can match against `OsError` and check its `kind()` rather than comparing against a string.

& particularly, you can use [thiserror](https://docs.rs/thiserror/latest/thiserror/) to generate `From` and `Error` implementations automatically:

```
use thiserror::Error;

#[derive(Error, Debug)]
enum CliError {
  #[error("protocol error: {0}")]
  ProtocolError(#[from] capnp::Error),
  #[error("OS error: {0}")]
  OsError(#[from] std::io::Error),
  #[error("parse error: {0}")]
  ParseError(String),
  #[error("RPC error: {0}")]
  RpcError(String),
}
```

By deriving `From`, you can continue to use `?` to propagate errors and they will be automatically converted from `capnp::Error` or `std::io::Error` into `CliError`.

adapter/cli/src/main.rs
```diff
@@ -60,178 +56,582 @@ fn run_cli(socket: &str) -> std::io::Result<()> {
...
                         if quit {
                             return Ok(());
                         }
                     }
-                    Err(err) => {
-                        println!("Failed to parse command \"{}\".  Error: {}", line, err);
-                    }
+                    Err(err) => match err.extra {
+                        // Quits the program if the specified port is not open
+                        ref val if *val == "No such file or directory (os error 2)".to_string() => {
```
cpacejo, line 77: See my suggestion above about creating a custom error type.  String matching errors like this can be fragile (e.g. this will likely differ on Windows).

adapter/cli/src/main.rs
```diff
@@ -60,178 +56,582 @@ fn run_cli(socket: &str) -> std::io::Result<()> {
...
 
-fn basic_call_response_0(comm: RpcCommands, socket: &str) -> std::io::Result<()> {
-    let command_str = format!("{}\n", comm);
-    basic_call_response_impl(&command_str, socket)
+// TODO combine no arg tasks into one func w/ closure
+async fn echo_task(
+    service: svc::Client,
+    system: capnp_rpc::RpcSystem<VatId>,
+) -> Result<(), capnp::Error> {
+    // TODO perhaps make the localset around the whole match?
```
cpacejo, line 174: Yes, I think you can/should do this at the toplevel!  Also there is an [upcoming enhancement to Tokio](https://github.com/tokio-rs/tokio/issues/6739) which makes the current-thread runtime effectively "automatically" start with a `LocalSet`, so you'd match that behavior.

Also the `spawn_local(system)` can/should then occur once at the toplevel (so no longer needed as an argument here).  So you effectively create the RPC link once, and then issue one (or more) commands on it via each of these tasks.

adapter/ph/src/admin_worker.rs
```diff
@@ -29,293 +32,422 @@ use tokio::net::{UnixListener, UnixStream};
...
-                    Err(_e) => {
-                        error!(target: RPC, "Connection failed");
-                    }
+pub async fn launch_capnp(
+    asm: Arc<Assembly>,
+    listener: UnixListener,
+) -> Result<(), Box<dyn std::error::Error>> {
+    // let listener = tokio::net::UnixListener::bind(path)?;
+
+    tokio::task::LocalSet::new()
```
cpacejo, line 49: No need to create a new `LocalSet` here or `.run_until()`, we create one already in `main()` and are already running in a local task (we were `spawn_local()`ed).

adapter/ph/src/admin_worker.rs
```diff
@@ -29,293 +32,422 @@ use tokio::net::{UnixListener, UnixStream};
...
-            Ok(RpcCommands::Echo) => {
-                buf_writer.write_all(echo(&asm).await.as_bytes()).await?;
-                buf_writer.write_all("OK\n".as_bytes()).await?
+impl svc::Server for AdminServiceImpl {
+    fn echo(
+        &mut self,
+        _: svc::EchoParams,
+        _: svc::EchoResults,
+    ) -> ::capnp::capability::Promise<(), ::capnp::Error> {
+        capnp::capability::Promise::from_future(async move {
```
cpacejo, line 94: So for this and many of the methods which "just" return a value (don't do any .awaiting or such), you can use `capnp::capability::Promise::ok(...)` to return the result directly.  So e.g. here would be `capnp::capability::Promise::ok(Ok(()))`.  This would eliminate the need for cloning `asm` in those methods also.

One "gotcha" though is that the `?` operator no longer works in those cases (since the return type of the methods is not itself a `Result`).  It works in the `from_future` blocks because the async block creates a new context, and the future's return type is a `Result`.  (`capnp` ought to, but I think does not, have an error propagation macro for this.)  So `Promise::ok()` is mostly just useful for the simpler methods like `echo` here.

adapter/ph/src/admin_worker.rs
```diff
@@ -29,293 +32,422 @@ use tokio::net::{UnixListener, UnixStream};
...
+                    success_builder.set_none(());
                 }
-            },
-            _ => buf_writer.write_all("ERR\n".as_bytes()).await?,
-        };
+                _ => {
+                    let error_builder = results_builder.init_error();
+                    error_builder
+                        .init_txt("Invalid program".len() as u32)
+                        .push_str(&"Invalid program");
```
cpacejo, line 236: You should be able just write `.set_txt("Invalid program")` here, which does these two steps for you.

adapter/ph/src/admin_worker.rs
```diff
@@ -29,293 +32,422 @@ use tokio::net::{UnixListener, UnixStream};
...
+        &mut self,
+        _: svc::PerfSampleParams,
+        mut results: svc::PerfSampleResults,
+    ) -> ::capnp::capability::Promise<(), ::capnp::Error> {
+        capnp::capability::Promise::from_future(async move {
+            let results_builder = results.get();
+            let response = format!("Not currently supported");
+            results_builder
+                .init_result(response.len() as u32)
+                .push_str(&response);
```
cpacejo, line 267: Same for these, I think you can just `.set_result(&response)`.

cpacejo (approved):
Overall this looks good!  Mostly some comments to streamline the Cap'n Proto stuff a bit, and to make the error type a bit semantically richer.  This is fine to check-in as-is though if you want to make those changes in a separate PR.

adapter/cli/src/main.rs
owenrhysthomas: Yes! This was going to be one of my future enhancements, I have a TODO about it somewhere in there

PR #1040, opened 2025-10-07
remove requirement for explicit link noise keys

### LINKED ISSUE #836: Once bootstrap is working, rethink link keys
Once we have bootstrap working we can probably rethink the requirement for pre-generated link keys.  Seems to me that only certain adapters -- those that need to boostrap -- would need pregenerated keys with well-known CN values.

Other adapters could just generate a new link key at startup and don't need a CN and can just perform normal ZPR authentication with existing authentication services.

Obviously this is all very hand wavey.  But it feels like if we have real auth we shouldn't need the link keys to be handed out by an admin.  An adapter and node can establish a link using Noise pattern that does a key exchange with no requirement for a priori knowledge on either side.  Once the link is up, it will be the auth step that decides id the actor/adapter is able to connect or not.

mkolehmainen:
A plain old adapter no longer needs to have a configured noise key.  If you drop the `certificate_file` and `private_key_file` configuration args, then the PH will just create a self-signed cert using a randomly generated noise key.

Since we use CNs so much, you do have to set the CN and that is set in the new `adapter.name` config.

Special adapters like "vs.zpr" still need a signed key.  And I've assumed that node<->node links will require signed keys.

### DIFF
--- a/adapter/ph/src/pki.rs
+++ b/adapter/ph/src/pki.rs
@@ -188,9 +188,6 @@ pub fn generate_self_signed_noise_cert(
 
     // Set the issuer name (for a self-signed cert, this is also the subject)
     let mut name = X509Name::builder()?;
-    name.append_entry_by_text("C", "US")?; // Country
-    name.append_entry_by_text("ST", "MA")?; // State or Province
-    name.append_entry_by_text("O", "ZPR")?; // Organization
     name.append_entry_by_text("CN", cn)?; // Common Name (e.g., domain name or IP)
     let name = name.build();
     builder.set_subject_name(&name)?;
@@ -223,6 +220,7 @@ pub fn generate_self_signed_noise_cert(
 
 #[cfg(test)]
 mod test {
+
     use super::*;
 
     #[test]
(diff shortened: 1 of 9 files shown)


adapter/ph/src/pki.rs
```diff
@@ -157,6 +163,64 @@ pub fn get_cn_from_cert(cert: &X509) -> Option<String> {
...
+    let mut serial = BigNum::new()?;
+    serial.rand(128, MsbOption::MAYBE_ZERO, true)?;
+    let serial_asn1 = serial.to_asn1_integer()?;
+    builder.set_serial_number(&serial_asn1)?;
+
+    // Set the issuer name (for a self-signed cert, this is also the subject)
+    let mut name = X509Name::builder()?;
+    name.append_entry_by_text("C", "US")?; // Country
+    name.append_entry_by_text("ST", "MA")?; // State or Province
+    name.append_entry_by_text("O", "ZPR")?; // Organization
```
cpacejo: I think we should leave these out of the subject name.  I.e. the full subject name should be only exactly the CN.

(Also we should be accepting certificates only if their full subject name is exactly just the CN we expect.  I think we don't have this code yet though?)

My reasoning being – otherwise we have the situation of having certificates for two different subjects (which are treated as different identities by most X.509 stuff) which we treat as the same subject (by ignoring everything but the CN).

cpacejo (changes requested):
Looks good!  Just one change I think we should make, self-signed certs (all of our certs actually) should have a subject name which only consists of a CN, no other hierarchical bits.

(Or alternatively, let's use and require the same hierarchical bits on all certificates.  I am just trying to avoid both the confused-deputy problem of accidentally only matching part of the subject name, and also the usability issue of having to specify a full distinguished name for every cert in config files.)

adapter/ph/src/pki.rs
mkolehmainen: Ah ok.  I think you are saying that to X509 the "identity" is concat of all those crazy components (include CN). I'll also add a ticket to check that we are using the full "identity" where appropriate.  I will patch up the self signed cert code.

cpacejo (approved):

PR #1045, opened 2025-10-09
remove retransmits from link state machine

cpacejo:
Since ZDPR handles retransmits we no longer need these!  I think this also solves the out-of-order init auth issue we're sometimes seeing too.  At least for me that happened when the LSM timed out and retransmitted, which resulted in multiple messages which eventually came through due to ZDPR.

We _do_ still need the timeouts, because we don't start `run_active()` which performs echos until after the link becomes `ACTIVE`.  So I've modified them all to simply terminate the link, and the timeouts which were previously named `RETRY_TIMER` I've multiplied by 3 (our old retry count) and renamed to `TIMEOUT`.  Also `TERMINATE_RESPONSE_TIMER` I've renamed to `TERMINATE_TIMEOUT` for consistency, though it didn't retransmit.

Also renamed `DEFAULT_REQUEST_RETRY_TIMER` to `DEFAULT_ZDPR_RETRY_TIMER` to reflect its sole new use.

### DIFF
--- a/adapter/ph/src/config.rs
+++ b/adapter/ph/src/config.rs
@@ -40,16 +40,17 @@ pub const TINY_MESSAGE_HEADROOM: usize = 64;
 
 pub const DEFAULT_ZDPR_RECEIVE_WINDOW_SIZE: usize = 32;
 
-pub const DEFAULT_REQUEST_RETRY_COUNT: usize = 3;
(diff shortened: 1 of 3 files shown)


mkolehmainen (approved):
Great!

PR #1047, opened 2025-10-09
changes from feedback

### LINKED ISSUE #1026: convert CLI to use capnp

owenrhysthomas:


adapter/cli/src/main_args.rs
```diff
@@ -22,7 +22,7 @@ pub struct CliCommand {
     pub command: Commands,
 }
 
-#[derive(Subcommand, Debug)]
+#[derive(Subcommand, Debug, PartialEq)]
```
cpacejo: Do you know where the requirement for `PartialEq` here came from?  I'm surprised we'd be comparing these for equality anywhere.

cpacejo (dismissed):
Sweet looks good!  Just a nit about the `PartialEq` derives I'm curious why we need them.

adapter/cli/src/main_args.rs
owenrhysthomas: It relates to the comment I left on cli/src/main.rs line 131 about quitting immediately if quit is the relevant command. Since we now do the bind within the `process_command` function, if the user wants to quit after the socket disconnects, say if the integration test ends and they want to quit, they would get an error that the port was not found then it would quit. It is not strictly necessary if there is some reason to not have the `PartialEq`, but it seemed more graceful to me than erroring out in that case.

cpacejo: Aha OK I see it!  So in this case what you can write instead is:
```
if matches!(command, Commands::Quit) { ... }
```
and that avoids the need to derive `PartialEq` on everything.

(The reason to avoid `PartialEq` is just the generic reason of avoiding having unnecessary mechanisms in the code -- in that case the compiler builds a bunch of code which knows how to compare two `Commands` values of any kind, but we only use it to check whether the enum tag is `Quit`.  Also as can be seen below it places the requirement that everything _inside_ of `Commands` in turn needs to derive `PartialEq`, which here turned out to be possible but in the general case isn't always, hence why it's not the default.)

owenrhysthomas: Right that makes sense, I didn't realize you could use that syntax, I will update that.

cpacejo (approved):
Perfect thanks!

PR #1049, opened 2025-10-09
remove authentication ACK messages

cpacejo:
Finally, addressing what first inspired ZDPR!

These three commits (maybe helpful to view separately) remove the explicit acknowledgement messages associated with the three authentication messages.  In detail:

* `InitAuthenticationResponse` goes away.  It always carried a successful status so we don't need to change anything there.  The node no longer transitions into `WaitForInitAuth`; instead it goes directly to `WaitForAcquireZprAddress`.  However, since that may be a long time, and it is useful that we provide quick feedback if the link is actually dead, we still set a short initial timeout, which is reset to a longer timeout upon receiving a ZDPR ACK for `InitAuthenticationRequest`.  (This mirrors the behavior of `WaitForInitAuth` without the extra machinery of actually transitioning to and from that state.)

* `AcquireZprAddressResponse` goes away.  It carried an error status in the case the node couldn't parse the corresponding request, which in turn caused the adapter to bring the link down.  Instead now, the node simply brings the link down itself.

* `GrantZprAddressResponse` goes away.  It carried an error status in the case the adapter couldn't parse the corresponding request, which in turn caused the node to bring the link down.   Instead now, the adapter simply brings the link down itself.  (The node therefore now also now transitions to `Active` from `RegisterAA` as soon as the visa service authorizes the connection, rather than waiting further for the adapter to acknowledge the grant.)

Also a couple trivial cleanups mixed in.

### DIFF
--- a/adapter/ph/src/link_state.rs
+++ b/adapter/ph/src/link_state.rs
@@ -145,7 +145,7 @@ pub enum LinkEvent {
     SentHelloResponse,
 
     ReceivedInitAuth((bool, Option<auth::ZdpInitAuthenticationPayload>)), // (bootstrap_flag, challenge)
-    ReceivedInitAuthResponse(ResponseCode),
+    ReceivedInitAuthAck,
 
     ReceivedAcquireZprAddressRequest(Option<Vec<IpAddress>>, String), // (requested_addrs, auth_blob)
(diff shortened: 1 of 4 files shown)


mkolehmainen (approved):
Ok great. Can't wait to try it.

PR #1052, opened 2025-10-14
Ort/logging

### LINKED ISSUE #1048: logging in admin worker

owenrhysthomas:
We didn't have any logging in this file before other than for errors relating to accepting connections to the socket, so I didn't want to go too overboard, but if we want I could definitely add more debug/trace logging.

### DIFF
--- a/adapter/ph/src/admin_worker.rs
+++ b/adapter/ph/src/admin_worker.rs
@@ -44,12 +44,9 @@ pub async fn launch_capnp(
     asm: Arc<Assembly>,
     listener: UnixListener,
 ) -> Result<(), Box<dyn std::error::Error>> {
-    // let listener = tokio::net::UnixListener::bind(path)?;
-
     loop {
         let (sock, _addr) = listener.accept().await?;
-        // TODO add connect info to log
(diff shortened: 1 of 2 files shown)


adapter/ph/src/admin_worker.rs
```diff
@@ -85,7 +82,7 @@ impl svc::Server for AdminServiceImpl {
         _: svc::EchoParams,
         _: svc::EchoResults,
     ) -> ::capnp::capability::Promise<(), ::capnp::Error> {
-        // TODO add logging here and in the other commands
+        info!(target: RPC, "Echo procedure initiated");
```
cpacejo: I think the echo & counters routines would be better at `debug` level, since they don't make any changes (and `counters` will be very chatty anyway when viewing live counters).  (`reset_counters` I'd keep at `info` though.)

cpacejo (changes requested):
Nice!  Just an ask to lower the level of the echo and counters RPCs because they might get chatty otherwise.

cpacejo (approved):
Thanks!

PR #1053, opened 2025-10-15
replaced five tuple parsing in handlers with classifier

### LINKED ISSUE #892: Use classifier as our one true place to parse packets
New code in the mgmt/handler parses the packets in a way that may already be done with more exactitude in our classifier.  Can we unite these two?

https://github.com/org-zpr/zpr-core/blob/07413661db69758c4c8f7c05135c203144c00717/adapter/ph/src/mgmt/handlers.rs#L584

owenrhysthomas:
This ends with the same result for the five_tuple, but there are a few warnings that do get lost by using the classifier

### DIFF
--- a/adapter/ph/src/mgmt/handlers.rs
+++ b/adapter/ph/src/mgmt/handlers.rs
@@ -8,7 +8,7 @@ use crate::config;
 use crate::counters;
 use crate::link_state::{LinkEvent, LinkStateError};
 use crate::logging::targets::{FLOW_MGMT, REPORTING, ZDP};
-use crate::net_defs::{ip_number, IpAddress};
+use crate::net_defs::IpAddress;
 use crate::packet::Packet;
 use crate::tlv::{self, TlvEncoding};
 use crate::zdp;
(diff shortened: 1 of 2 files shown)


adapter/ph/src/mgmt/handlers.rs
```diff
@@ -704,116 +704,39 @@ pub async fn handle_bind_actor_address_request(
...
-            }
-            let icmp_type = pkt.get_u8();
-            src_port = icmp_type.into();
-            dst_port = icmp_type.into();
-        }
-        _ => {
-            warn!(target: ZDP, "Link {}: unsupported IP protocol {}", pkt.metadata().ingress_link_id, ip_protocol);
-            return Err(HandleMgmtError::BadStructure);
+            let metadata = pkt.metadata_mut();
+            metadata.set_dst_port(metadata.get_src_port_hbo());
```
mkolehmainen: Not sure if this would break the prototype visa service, but the v2 visa service will want ICMP type in source_port and ICMP code in dest_port.

mkolehmainen (dismissed):
LGTM

adapter/ph/src/mgmt/handlers.rs
owenrhysthomas: It works fine with the icmp_type and icmp_code in the src and dst ports respectively. I suspect it doesn't make a difference for the current implementation, so it was easier to just put the type in both, but since it's more steps here to put the type into the dst instead of leaving the code in there, I will remove it.

owenrhysthomas:
The test was failing locally, so I had to rebase to main, hence all the extra commits

adapter/ph/src/mgmt/handlers.rs
mkolehmainen: Ah ok, i just checked the v1 VS and it uses the copy of the raw packet we make so the 5-tuple in this case has no bearing.  In V2 we will use the 5-tuple and (eventually) we will want the icmp-code in dst_port.

mkolehmainen:
I'm hoping @svjaffer  can help you restore git sanity here.

owenrhysthomas:
> I'm hoping @svjaffer can help you restore git sanity here.

Had Sarah walk me through fixing it, should be good now - I should've merged instead of rebasing.

adapter/ph/src/mgmt/handlers.rs
owenrhysthomas: Yes, with the change I made in the most recent commit, the icmp code will be in the dst port

svjaffer (approved):

mkolehmainen (approved):
Excellent!

Issue #1055, opened 2025-10-15
KM-Noise packet retransmits


cpacejo:
KM messages are by design not protected by ZDPR – they are meant to simply directly tunnel KM protocol messages.  However the KM-Noise backend is not resilient to lost packets, which causes flaky-network testing to fail often.  Let's add retransmits to KM-Noise to remedy this.

PR #1056, opened 2025-10-15
add (but do not enable) lossy-network integration test

cpacejo:
Adds a test which runs with 10% packet loss to exercise ZDPR functionality.  Has already helped find a bunch of ZDPR bugs.  Unfortunately KM-Noise isn't resilient to packet loss so the test often fails during handshaking.  (See <https://github.com/org-zpr/zpr-core/issues/1055>.)  Hence it's not enabled as part of our checkin suite.


owenrhysthomas (approved):
Cool!

PR #1059, opened 2025-10-15
ZDPR loose ends

cpacejo:
Tiny ZDPR things I've missed.

`GRANT_REQUEST_TIMEOUT` is no longer used (since nodes now immediately transition to Active at that point) and in fact causes error logs when it eventually goes off with no-one handling it.

I keep forgetting to bump the PH version to reflect these protocol changes, so here we go.

Echos no longer need the sequence number field.  It was only used in echo responses, and we don't have those anymore.


owenrhysthomas (approved):
🔥

PR #1061, opened 2025-10-20
initial fix of lua dissector - includes fixes to align with current p…

### LINKED ISSUE #415: Lua dissector: expert info malformed fetch unsigned integer when parsing Transit Packet

### LINKED ISSUE #418: Lua dissector: display MACs as hex, not decimal

### LINKED ISSUE #1054: EPIC: Update Lua dissector

owenrhysthomas:
…rotocol

### DIFF
--- a/tools/README
+++ b/tools/README
@@ -5,12 +5,13 @@ Installation instructions for Lua dissectors
 - When the new window pops up, click folders
 - Find where it says 'Personal Lua Plugins' and copy the path it shows under Location
 - Copy both the zdp dissector and the link layer header dissector into that folder
+- Restart Wireshark
 - In the Wireshark menu bar again, click Edit -> Preferences
 - In the preferences window, expand the Protocols section
 - Scroll until you find DLT_USER
 - Edit the Encapsulations Table
 - Click the plus button to add a new user
-- Set the Payload protocol to zdp, the Header size to 1, adn the Header protocol to zdp_link_p2p
+- Set the Payload dissector to zdp, the Header size to 1, and the Header protocol to zdp_link_p2p
 - Set the trailer size to 0 and leave the Trailer protocol blank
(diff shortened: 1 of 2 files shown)


owenrhysthomas:
@cpacejo Here's a list of changes that I noticed in the current ZDP proto compared to the RFC
- HMAC is added onto the end of the packet, after the a2a_mac, and it seems to be removed before packets are captured in the capture worker
- the nonce gets added and then removed again during encryption, so is also not visible during capture
- I don't think there is a transaction ID on any of the management packets except Bind actor request/response. I am not positive about this one because in mgmt::dispatch.rs any packet that requires processing (i.e. any that have Disposition::AckAndProcess and therefore reach dispatch::handle_response need to have a transaction ID, but on line 92 of mgmt_processor_workers it seems like only BindActorAddressRequest packets are expected to have a transaction ID, and from what I saw in the dissector, at least hello, echo, and I believe terminate did not have transaction IDs
- Potentially more TK as I decode more of the different types of management packets.

cpacejo (approved):
Nice!  Will try this out later today.

With my latest changes a couple things will break (but feel free to address in a separate PR):

1) The sequence number field is removed from Transit, ARP, and Key Management packets.
2) The sequence number field in all other packets grows from 2 to 8 bytes.

owenrhysthomas:
> Nice! Will try this out later today.
> 
> With my latest changes a couple things will break (but feel free to address in a separate PR):
> 
>     1. The sequence number field is removed from Transit, ARP, and Key Management packets.
> 
>     2. The sequence number field in all other packets grows from 2 to 8 bytes.

Sounds great! I'll throw up another PR with those fixes. There is also another PR coming soon with more handling for different kinds of management messages, which will come with more discrepancies between the current protocol and the RFC.

PR #1063, opened 2025-10-20
use full 8 bytes for sequence numbers on the wire

cpacejo:
In order to prevent replay attacks, we need to incorporate the full sequence number of ZDPR packets into the AEAD data somehow.  Since the sequence number is now only present on mgmt packets, Sarah suggested simply including the full sequence number in the packet body and doing away with the truncated sequence number mechanism.  Since management packets are mostly small (and the large ones are very large), this doesn't cause any problems and is architecturally simple, so this patch does that.

Note that this is a deviation from RFC 6/17.

Also a drive-by micro performance enhancement in ZDPR as suggested in <https://doc.rust-lang.org/std/task/struct.Waker.html#impl-Clone-for-Waker>.


mkolehmainen (dismissed):
Great!

mkolehmainen (approved):
Can't believe I didn't catch that missing negation! Great!

PR #1064, opened 2025-10-21
changed which packets are expected to have seq num

### LINKED ISSUE #1054: EPIC: Update Lua dissector

owenrhysthomas:


cpacejo (approved):
Thanks!  Both these changes are now checked in so the dissector and PH should be in sync again.

PR #1066, opened 2025-10-23
Ort/enhance mgmt dissection

### LINKED ISSUE #413: Lua dissector: display IP addresses as IPs, not hex

### LINKED ISSUE #414: Lua dissector: range out of bounds zdp_dissector.lua:180 when parsing Bind Agent Address Request

owenrhysthomas:


tools/zdp_dissector.lua
```diff
@@ -425,6 +414,12 @@ function Dissector(init_tree, init_buffer)
             self.tree:add(field, self.buffer(self.pos, len))
             self.pos = self.pos + len
         end,
+        add_field_and_return = function(self, field, len)
+            local val = self.buffer(self.pos, len)
+            self.tree:add(field, val)
+            self.pos = self.pos + len
+            return val:uint()
+        end,
```
owenrhysthomas, line 429: I suppose you could just make the regular `add_field` function do this work, but not sure if there would be a performance hit for returning and converting to uint for every value you want to add to the tree

owenrhysthomas:
More differences I noticed between the RFC and the current code

- Bind request does not have packet count (nor does it appear to contain multiple actor packets), or endpoint packet length it has also been renamed from bind endpoint to bind actor. It also has a compression mode
- Several ZDP type numbers are used differently in the RFC than in the Code. See the comments in the `type_name_table` for specifics
  - 5, 6, 9, 10, 15, 16, 17, 18 , 127, 132, 133, 134, 142, 143, 144, 147, 149 all diverge
- Acquire ZPR address packets have the order of the blob and address fields swapped in the RFC and the implementation
- Grant ZPR address packets do not have a blob or DHCP info in the implementation

owenrhysthomas:
Also - in terms of dissecting the Blobs for the AcquireZprAddressRequest, I know there are two types of blobs, `ZdpSelfSignedBlob` and `ZdpAuthCodeBlob`, but I don't think there is a way to differentiate between them, or even to distinguish between the fields since they are variable length strings. I suppose that's why it's called a blob.

tools/zdp_dissector.lua
cpacejo: Hmm yeah this might be a micro-optimization, but I also think it's fine to leave as-is!

cpacejo (approved):
Nice!  Parsing the TLVs is a great help!

PR #1068, opened 2025-10-25
Replace sync_req mechanism with transaction manager

cpacejo:
The sync_req mechanism was used both to link responses to requests, and also to manage sequence IDs, transmit windows, and retries.

Since ZDPR now performs all of those latter functions, and we also want to move away from the request+response model, we replace sync_req (which has served bind requests admirably for over a year!) with a simple transaction ID manager which operates entirely on the sender side.

Pending org-zpr/zpr-core#906 (which will break our sole implemented synchronous request+response method, BindActorAddress, into separate async steps), we also temporarily add state to the peer table which allows BindActorAddressResponses to be forwarded to the correct awaiting code.

This is the "final" piece in the ZDPR journey!  (I feel like I said that before but this is actually the last thing in my TODO!)

Suggest viewing with whitespace changes disabled (lots of indentation changes).


mkolehmainen (dismissed):
Looks good.  I would have gpt write you some unit tests for the txn code!

cpacejo:
> Looks good. I would have gpt write you some unit tests for the txn code!

Oh I meant to add some!  Lemme do that.

mkolehmainen (approved):
Excellent!

PR #1079, opened 2025-10-30
Created five tuple lookup table

### LINKED ISSUE #889: More efficient visa table search
The adapters new match_traffic function in visa_table.rs uses a simple linear search over all the visas.  This story is about creating a traffic index to make the search more efficient.

owenrhysthomas:
I think the best way to review this will be commit by commit, which are

1. the first commit is about constructing the table and all of the tests associated with that
2. the second commit adds the lookup function that takes in a five tuple and returns a visa, if it is in the table
3. the third commit integrates the `FiveTupleLookupTable` into the visa_table - makes `visa_table::match_traffic` essentially a wrapper for `FiveTupleLookupTable::find_match`

Major TODO is to add a better way to update the lookup table. Currently every time a visa is added or removed, the whole table is remade from scratch. That is fine currently because the visa table is relatively small, but will become problematic quickly.

The first improvement will likely use some sort of cloning and `Arc`s, as suggested by Chris. The step after that would be some more specific editing.

Editing the table also is likely easier right now because of the lack of wildcards, lists, etc, will become more complicated as allowed values increase.

### DIFF
--- a/adapter/ph/src/five_tuple_lookup_table.rs
+++ b/adapter/ph/src/five_tuple_lookup_table.rs
@@ -1,0 +1,886 @@
+use crate::net_defs::{IpAddress, IpProtocol};
+use crate::rcu::RcuBox;
+use crate::visa_table::Visa;
+
+use ip_network_table_deps_treebitmap::IpLookupTable;
+use range_set_blaze::RangeMapBlaze;
+use std::collections::HashMap;
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+use zpr::{L3Type, VisaId};
+
+pub type FiveTupleLookup = HashMap<
+    IpAddress,
+    IpLookupTable<Ipv6Addr, RangeMapBlaze<u16, RangeMapBlaze<u16, Vec<(IpProtocol, VisaId)>>>>,
+>;
+
+pub struct FiveTupleLookupTable {
+    pub table: RcuBox<FiveTupleLookup>,
+}
+
+impl FiveTupleLookupTable {
+    // TODO change how construction is done once visas move away from being based on a FiveTuples
+    pub fn new(visa_table: HashMap<VisaId, Visa>) -> Self {
+        let mut hash_table: FiveTupleLookup = HashMap::new();
+        for (visa_id, visa) in visa_table.iter() {
+            let five_tuple = match visa.ftuple {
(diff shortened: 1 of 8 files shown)


adapter/ph/src/five_tuple_lookup_table.rs
```diff
@@ -0,0 +1,1576 @@
...
+                _ => None,
+            };
+
+            // TODO This is quite inefficient, improve
+            // Try to add to hash table, if there is a collision, combine the tables, then add the combined table
+            match hash_table.insert(five_tuple.dst_address, ip_table) {
+                None => (),
+                Some(removed_src_addrs) => {
+                    let in_table_src_addrs = hash_table.get_mut(&five_tuple.dst_address).unwrap();
+                    for (og_src_addr, og_mask_len, og_dst_ports) in removed_src_addrs.iter() {
```
svjaffer, line 78: Could we break some of this up into functions?  Even if this isn't code we can reuse, I think it would be more readable.

adapter/ph/src/five_tuple_lookup_table.rs
```diff
@@ -0,0 +1,1576 @@
...
+    fn test_construction_one_visa() {
+        let src_addr = [1u8; 16];
+        let dst_addr = [2u8; 16];
+
+        let l4proto = vsapi::PEPIndex::TCP;
+        let src_port = 10;
+        let dst_port = 11;
+        let src_dst =
+            vsapi::PEPArgsTCPUDP::new(Vec::new(), Vec::new(), src_port, dst_port, None, None);
+        let visa: vsapi::Visa = vsapi::Visa::new(
```
svjaffer, line 188: Since you don't care about most of these fields, you can save some repetition and visual space by making a test function that constructs a visa from only the fields you care about in these tests.  NBD if you'd rather not, though.

adapter/ph/src/five_tuple_lookup_table.rs
owenrhysthomas: resolved in later PR

adapter/ph/src/five_tuple_lookup_table.rs
owenrhysthomas: will resolve in later PR

svjaffer (approved):

PR #1080, opened 2025-10-31
updated cli to rust 2024

owenrhysthomas:


cpacejo (approved):
Nice thanks!  Reminder to myself to do this for the PH also.

PR #1081, opened 2025-10-31
PH: bump Rust edition to 2024

cpacejo:
Beside the formatting differences, main changes were:

- Using `.as_ref()` / `.as_mut()` instead of `ref` bindings when matching an `&Option<_>`
- Adding `+ use<>` to one function signature
- Replace `gen()` with `r#gen()`

Also I did a drive-by change of a match to an if/let to reduce indentation in one spot.


adapter/ph/src/main_argparse.rs
```diff
@@ -768,7 +768,9 @@ mod test {
         let args = vec!["ph", "adapter", "-c", tmpfile.get_path().to_str().unwrap()];
 
         let noise_key = "GExPGh5RE/nKo8WoN8EqknDDNIEjWBL6PZm08Uhvn0w=";
-        env::set_var("NOISE_PRIVATE_KEY", noise_key);
+        unsafe {
```
owenrhysthomas, line 771: Do we want to add a `SAFETY` comment here? Or is it okay since it's just in a test? I see other `unsafe` blocks in the test don't have `SAFETY` comments, so I assume it's fine, just wanted to confirm

owenrhysthomas (dismissed):
Nice!

adapter/ph/src/main_argparse.rs
cpacejo: Oh good catch!  Yes without serialization (which is in place) there's the potential to cause a crash during testing.

owenrhysthomas (approved):

PR #1090, opened 2025-11-03
Put enums in the table

### LINKED ISSUE #1083: improve five tuple lookup table

owenrhysthomas:
The new commits for this branch start at `added dst port level enum`

`added dst port level enum` and `added src port level enum` are the two major commits on this branch. `added wildcard insertion tests` and `added function to make visas w/ relevant info in tests` are more housekeeping/cleanup

I think that going commit-by-commit would be helpful to get a sense of what happened at each stage, but I would not spend too much time looking at the tests commit by commit because they change a lot, and the final commits clean the tests up, so they are easier to read. 

Also - a lot of the code that is written in `added dst port level enum` is changed in added `src port level enum`, so I wouldn't spend too much time separately. As I said, I think it could be helpful to skim each commit separately, but I don't think it is super helpful overall since so many of the changes were then changed again


adapter/ph/src/five_tuple_lookup_table.rs
```diff
@@ -11,15 +11,29 @@ use zpr::{L3Type, VisaId};
...
+// generic variables in place of the specific values inside
+#[derive(Clone)]
+pub enum DstPortLevel {
+    Wildcard(SrcPortLevel),
+    MultiVal(RangeMapBlaze<u16, SrcPortLevel>),
+    SingleVal((u16, SrcPortLevel)),
+}
+
+#[derive(Clone, Eq, PartialEq)]
+// Need eq for RangeMapBlaze, and need partialeq for eq - should i explicitly define them myself?
```
svjaffer, line 30: As discussed this morning: Nope!  Derive should be correct here.

adapter/ph/src/five_tuple_lookup_table.rs
```diff
@@ -142,30 +113,251 @@ impl FiveTupleLookupTable {
...
+            }
+        }
+
+        intersection
+    }
+
     pub fn find_match(&self, ft: FiveTuple) -> Option<VisaId> {
+        // NOTE I didn't make a subfunction for finding the match for the src_level, even though it is essentially repeated three times
+        // becuase I know this func is all about speed and sometimes passing to another function can cause minor slowdown, not sure
+        // how the rust compiler handles such things or if it is significiant enough to matter, if not I will make a helper func
```
svjaffer, line 303: For now, you should assume the compiler will do the right thing.  If not, we should be able to see it down the line with perf testing and correct it at that point.

svjaffer (approved):

PR #1091, opened 2025-11-05
updated to capnp v23, no longer need promises

### LINKED ISSUE #1085: update cli rpc to use capnp v0.23

owenrhysthomas:
NOTE: tests will not pass until https://github.com/org-zpr/zpr-admin-api-rs/pull/12 is approved and merged.

### DIFF
--- a/adapter/cli/Cargo.toml
+++ b/adapter/cli/Cargo.toml
@@ -15,11 +15,11 @@ shlex = { version = "1.3.0" }
 zpr = { path = "../../zpr" }
 zpr-ext = { path = "../../zpr-ext", features = ["bytes", "tokio", "zerocopy"] }
 clap_complete = { version = "4.5.57", optional = true}
-capnp = { version = "0.21" }
-capnp-rpc = { version = "0.21" }
+capnp = { version = "0.23" }
+capnp-rpc = { version = "0.23" }
 tokio = { version = "1.37.0", features = ["macros", "rt", "io-util", "time"] }
 tokio-util = { version = "0.7.11", features = ["compat"] }
-cli-proto = { git = "https://github.com/org-zpr/zpr-admin-api-rs.git", tag = "v0.2.0" }
+cli-proto = { git = "https://github.com/org-zpr/zpr-admin-api-rs.git", tag = "v0.3.0" }
 thiserror = { version = "2.0.17" }
 
 [features]
--- a/adapter/cli/Cargo.toml
+++ b/adapter/cli/Cargo.toml
@@ -19,7 +19,7 @@ capnp = { version = "0.23" }
 capnp-rpc = { version = "0.23" }
 tokio = { version = "1.37.0", features = ["macros", "rt", "io-util", "time"] }
 tokio-util = { version = "0.7.11", features = ["compat"] }
-cli-proto = { git = "https://github.com/org-zpr/zpr-admin-api-rs.git", tag = "v0.3.0" }
+cli-proto = { git = "https://github.com/org-zpr/zpr-admin-api-rs.git", tag = "v0.6.0" }
 thiserror = { version = "2.0.17" }
 
 [features]
--- a/adapter/ph/src/admin_worker.rs
+++ b/adapter/ph/src/admin_worker.rs
@@ -23,6 +23,7 @@ use std::fmt::Write;
 use std::io::Error;
 use std::io::IoSliceMut;
 use std::path::PathBuf;
+use std::rc::Rc;
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
@@ -78,20 +79,20 @@ struct AdminServiceImpl {
 
 impl svc::Server for AdminServiceImpl {
     async fn echo(
-        self: ::capnp::capability::Rc<Self>,
+        self: Rc<Self>,
         _: svc::EchoParams,
         _: svc::EchoResults,
-    ) -> Result<(), ::capnp::Error> {
+    ) -> Result<(), capnp::Error> {
         debug!(target: RPC, "Echo procedure initiated");
 
         Ok(())
     }
 
     async fn reset_counters(
-        self: ::capnp::capability::Rc<Self>,
+        self: Rc<Self>,
         _: svc::ResetCountersParams,
         _: svc::ResetCountersResults,
-    ) -> Result<(), ::capnp::Error> {
+    ) -> Result<(), capnp::Error> {
         info!(target: RPC, "Reset counters procedure initiated");
         for value in self.asm.counters.management.values() {
             value.reset();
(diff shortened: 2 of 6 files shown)


adapter/cli/Cargo.toml
```diff
@@ -15,11 +15,11 @@ shlex = { version = "1.3.0" }
...
 zpr-ext = { path = "../../zpr-ext", features = ["bytes", "tokio", "zerocopy"] }
 clap_complete = { version = "4.5.57", optional = true}
-capnp = { version = "0.21" }
-capnp-rpc = { version = "0.21" }
+capnp = { version = "0.23" }
+capnp-rpc = { version = "0.23" }
 tokio = { version = "1.37.0", features = ["macros", "rt", "io-util", "time"] }
 tokio-util = { version = "0.7.11", features = ["compat"] }
-cli-proto = { git = "https://github.com/org-zpr/zpr-admin-api-rs.git", tag = "v0.2.0" }
+cli-proto = { git = "https://github.com/org-zpr/zpr-admin-api-rs.git", tag = "v0.3.0" }
```
cpacejo: Humm I think this should be at v0.6.0 to match what's in the bindings repo (which might need a proper tag applied if the automation isn't working there)?

adapter/ph/src/admin_worker.rs
```diff
@@ -77,20 +77,21 @@ struct AdminServiceImpl {
 }
 
 impl svc::Server for AdminServiceImpl {
-    fn echo(
-        &mut self,
+    async fn echo(
+        self: ::capnp::capability::Rc<Self>,
```
cpacejo: This is just an alias for `std::rc::Rc`, which you can/should just `use` so you can write this simply as `Rc<Self>`.

adapter/ph/src/admin_worker.rs
```diff
@@ -77,20 +77,21 @@ struct AdminServiceImpl {
...
         _: svc::EchoParams,
         _: svc::EchoResults,
-    ) -> ::capnp::capability::Promise<(), ::capnp::Error> {
+    ) -> Result<(), ::capnp::Error> {
```
cpacejo: The initial `::` before `capnp` isn't necessary; I think the generated code just does that to avoid name conflicts.  (It's not wrong of course, just trying to keep consistent with our other fully-qualified names where we don't use an initial `::`.)

adapter/ph/src/admin_worker.rs
```diff
@@ -149,123 +150,112 @@ impl svc::Server for AdminServiceImpl {
...
 
     fn set_capture_file(
-        &mut self,
+        self: ::capnp::capability::Rc<Self>,
         _: svc::SetCaptureFileParams,
         _: svc::SetCaptureFileResults,
-    ) -> ::capnp::capability::Promise<(), ::capnp::Error> {
-        info!(target: RPC, "Set capture file procedure initiated");
-        ::capnp::capability::Promise::err(::capnp::Error::unimplemented(
+    ) -> impl ::core::future::Future<Output = Result<(), ::capnp::Error>> + 'static {
```
cpacejo: This should be able to become an `async fn` too also I think?  Which just immediately returns `Err(capnp::Error::unimplemented(...))`.

adapter/ph/src/admin_worker.rs
```diff
@@ -279,81 +269,75 @@ impl svc::Server for AdminServiceImpl {
...
+        Ok(())
     }
 
     fn stop_link(
-        &mut self,
+        self: ::capnp::capability::Rc<Self>,
         params: svc::StopLinkParams,
         mut results: svc::StopLinkResults,
-    ) -> ::capnp::capability::Promise<(), ::capnp::Error> {
+    ) -> impl ::core::future::Future<Output = Result<(), ::capnp::Error>> + 'static {
```
cpacejo: Same here I think this one should be able to become an `async fn`?

adapter/ph/src/admin_worker.rs
```diff
@@ -377,27 +361,24 @@ impl svc::Server for AdminServiceImpl {
...
+        Ok(())
     }
 
     fn change_logging(
-        &mut self,
+        self: ::capnp::capability::Rc<Self>,
         params: svc::ChangeLoggingParams,
         mut results: svc::ChangeLoggingResults,
-    ) -> ::capnp::capability::Promise<(), ::capnp::Error> {
+    ) -> impl ::core::future::Future<Output = Result<(), ::capnp::Error>> + 'static {
```
cpacejo: One more which should be able to become an `async fn`

cpacejo (changes requested):
A few minor changes but looks good!

adapter/cli/Cargo.toml
mkolehmainen: The generated repos have their own, independent version numbers!

owenrhysthomas: Yes, the version number is 6 because the this.version only changes the minor version, but some of the tags had patch numbers, so the current tag is at v0.2.0, and the next one will be v0.3.0. I could change that to bring the tag in line with the version number, though, if it is confusing that they diverge.

mkolehmainen: Also want to bring up that maybe the schema shared by the ph and cli should just live in core and be referenced by path.  Seems like extra complexity here to separate it out that maybe isn't needed.  This is just me ruminating.

cpacejo: > The generated repos have their own, independent version numbers!

Yes!  But this generated repo is at 0.6.0: https://github.com/org-zpr/zpr-admin-api-rs/blob/main/this.version  (unless I'm missing something??)

cpacejo: > Also want to bring up that maybe the schema shared by the ph and cli should just live in core and be referenced by path. Seems like extra complexity here to separate it out that maybe isn't needed. This is just me ruminating.

This was my thought when we created this but then Owen & I decided it should be split out for some reason that now escapes me.

cpacejo (dismissed):
Sweet thanks!

adapter/cli/Cargo.toml
owenrhysthomas: > > Also want to bring up that maybe the schema shared by the ph and cli should just live in core and be referenced by path. Seems like extra complexity here to separate it out that maybe isn't needed. This is just me ruminating.
> 
> This was my thought when we created this but then Owen & I decided it should be split out for some reason that now escapes me.

I believe it was because Sarah thought that we shouldn't have generated code checked into core.

mkolehmainen: Seems like the Cap'n Proto way (and maybe the Rust way) is to have the build.rs do the generation. So the cargo file makes sure that you get the expected versions of things.

mkolehmainen: But also I think it does make sense to expose a lib where we have a bunch of potential clients for it.  Although in the case of rust the "build-bindings" workflow doesn't really do anything -- except copy over the latest schema file.

mkolehmainen: Also @cpacejo the this.version thing has nothing to do with the tag applied to the language repo.

mkolehmainen (approved):
LGTM. In future we may want to consider bringing the API repo back into core.

PR #1092, opened 2025-11-05
Cslab iterator

### LINKED ISSUE #435: iterator over Cslab / PeerTable

cpacejo:
Adds iterator functionality to the Cslab / RcuCslab and PeerTable.  This replaces our ad-hoc list of peer IDs we maintained under a Mutex.

The iterator implementation simply scans the bitmap of allocated entries (using bitcounting for performance).  To avoid walking/paging in the entire bitmap if (as is almost always the case) only a small prefix has been used, we also track the maximum entry which has ever been allocated (the "frontier") and stop there.  The frontier is updated prior to marking the entry as in-use to ensure consistency if a scan is performed after a lookup (explained in-depth in the comments).  The frontier is never decreased (reasoning also given in the comments).

Also simplified some of the logic in `assembly` regarding iterating over peers.

I can't figure out a good way to export an iterator from the peer table, so it just has `for_each()` and `find()` methods directly on it for now.  (I suspect I'd need to add a `MappedRcuGuard` type but that would be blocked on [`MappedRwLockReadGuard`](https://doc.rust-lang.org/std/sync/struct.MappedRwLockReadGuard.html).)

Closes #435.

### DIFF
--- a/adapter/ph/src/admin_worker.rs
+++ b/adapter/ph/src/admin_worker.rs
@@ -268,7 +268,13 @@ impl svc::Server for AdminServiceImpl {
     ) -> ::capnp::capability::Promise<(), ::capnp::Error> {
         info!(target: RPC, "Show link summary procedure initiated");
         {
-            let peer_ids = self.asm.peer_ids.lock().unwrap();
+            let mut peer_ids = Vec::new();
+            self.asm.peer_table.for_each(|(id, _)| {
+                if id.get() != zpr::LOCAL_ACTOR_LINK_ID {
+                    peer_ids.push(id.get())
+                }
+            });
+
             let mut results_builder = results.get().init_summary(peer_ids.len() as u32);
 
             for (i, id) in peer_ids.iter().enumerate() {
--- a/adapter/ph/Cargo.lock
+++ b/adapter/ph/Cargo.lock
@@ -485,7 +485,7 @@ dependencies = [
 
 [[package]]
 name = "cslab"
-version = "0.1.1"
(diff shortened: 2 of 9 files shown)


cpacejo:
For some reason `five_tuple_lookup_table` formatting is failing, PR here: https://github.com/org-zpr/zpr-core/pull/1093

adapter/ph/src/admin_worker.rs
```diff
@@ -268,7 +268,13 @@ impl svc::Server for AdminServiceImpl {
     ) -> ::capnp::capability::Promise<(), ::capnp::Error> {
         info!(target: RPC, "Show link summary procedure initiated");
         {
-            let peer_ids = self.asm.peer_ids.lock().unwrap();
+            let mut peer_ids = Vec::new();
+            self.asm.peer_table.for_each(|(id, _)| {
+                if id.get() != zpr::LOCAL_ACTOR_LINK_ID {
```
mkolehmainen, line 273: I guess we used to show the local-actor-link-id peer, but now we don't.  I guess that's fine?

mkolehmainen (approved):
Glad to see the tests!  Looks good to me!

adapter/ph/src/admin_worker.rs
cpacejo: We didn't add it to the `peer_ids` list!  I didn't realize this at first and things were crashing (since we were trying to send shutdown messages to it).  So by skipping it when iterating I retain the original behavior.

PR #1093, opened 2025-11-05
five tuple lookup table formatting

cpacejo:
For some reason my `cargo fmt` formats this differently from yours?  What compiler version are you on (`rustup show`)?


owenrhysthomas:
This was because the PR was created before we changed the edition to rust 2024, but then it merged into main which was running rust 2024. I had another PR related to the table that I fixed it in, but good to do it here so it's sooner.

owenrhysthomas (approved):

PR #1094, opened 2025-11-06
arc-swap backend for Rcu

cpacejo:
Adds arc-swap as a backend for Rcu.  (I hadn't found it in my initial survey of the landscape!)

Also includes a bugfix for a locking oversight I noticed in the aarc backend (which is not currently used).  Unfortunately I don't know a good way to unit-test this.


mkolehmainen (approved):
neato!

PR #1099, opened 2025-11-07
prints counts on failure

### LINKED ISSUE #1067: counters not dumped when integ test fails
If the integration test ends in FAILURE the counters are not dumped.  They should be as that's when they are most useful.

owenrhysthomas:


integration-test/capture-test.sh
```diff
@@ -189,14 +189,23 @@ set_program "$ADAPTER1_SOCK" "$TMPDIR/cap_test1.pcap" 'link[0] == 1'
...
+PASS=0
 echo "Wait for TUN carrier..."
-wait_for 5 check_carrier zpr-node tun0 || { echo "FAILURE"; exit 1; }
-wait_for 5 check_carrier zpr-vs tun0 || { echo "FAILURE"; exit 1; }
-wait_for 5 check_carrier zpr-a tun0 || { echo "FAILURE"; exit 1; }
-wait_for 5 check_carrier zpr-b tun0 || { echo "FAILURE"; exit 1; }
-if [[ "$NUM_ACTORS" -ge 3 ]]; then
-  wait_for 5 check_carrier zpr-c tun0 || { echo "FAILURE"; exit 1; }
+wait_for 15 check_carrier zpr-node tun0 || { echo "FAILURE"; PASS=1; }
+if [[ "$PASS" == 0 ]] then
```
owenrhysthomas, line 195: you don't really need to check PASS before every line, but if you don't then when there is a FAILURE it will wait for all of them to fail before moving on, this way once one fails it fails immediately

cpacejo (changes requested):
Ah!  I forgot I had all of these as early exits.  Good catch!

I think you can probably also get rid of the `echo "FAILURE"` calls here, since now we should reach the common code at the end which prints "FAILURE" if PASS is 1.

(If I'm wrong just ignore bc otherwise this looks good!)

owenrhysthomas:
> Ah! I forgot I had all of these as early exits. Good catch!
> 
> I think you can probably also get rid of the `echo "FAILURE"` calls here, since now we should reach the common code at the end which prints "FAILURE" if PASS is 1.
> 
> (If I'm wrong just ignore bc otherwise this looks good!)

Ah yes, true!

cpacejo (approved):
Nice thanks!

PR #1104, opened 2025-11-10
aligned mismach between rfc and code

### LINKED ISSUE #1086: Fix up ZPR message type numbers to match RFC
Many of these differ from what is now in the RFC; see the Lua dissector for list of which ones differ.

owenrhysthomas:
All other differences noted in the Lua dissector are different because the message type is different in the RFC, but all the message types that are in both the RFC and the code are the same.


cpacejo (approved):
Nice thank you!

PR #1105, opened 2025-11-10
allow for setting of auth_private_key in prep for libnode2

mkolehmainen:
Add a way to pass in an RSA private key through command line and config file.

This is not yet used but will be needed in libnode2 and I'm adding it in now to help with testing libnode2.


cpacejo (approved):
Nice!

PR #1107, opened 2025-11-11
Independent classifier

### LINKED ISSUE #1089: make classifier independent of Packet
The classifier should operate on any `&[u8]` and return a `FiveTuple` rather than writing it into a `Packet`.

owenrhysthomas:
On success, the classifier now returns a tuple of the `ClassifierResult` and the `FiveTupleOptional` which has all the same fields as a `FiveTuple`, except that they are all `Option`s. This allows us to maintain the same functionality that the existing classifier has where it can set parts of the FiveTuple in the packet metadata before (potentially) erroring out. I created the new struct because if we used `FiveTuple` we would have to designate some dummy values (like 0) that are not allowed to mean that the value was not set, which is not ideal.

If the classifier returns an error, it still returns the FiveTupleOptional along with the error message, so that the user can apply the parts of the `FiveTuple` that were parsed (if any) if they want.

Once the buffer is classified, the user must simply call `write_ft_optional` on the packet and pass in the `FiveTupleOptional` that was returned from the call to `classify` and they will have the same functionality the classifier previously had.

If the user wants a regular `FiveTuple` they can simply call `FiveTuple::try_from` on a FiveTupleOptional; to use the try_from, the `FiveTupleOptional` must be fully specified, i.e. none of the fields can be `None`, otherwise there will be an error. I can add another function that will convert regardless and just put in dummy values to the new `FiveTuple` if that would be helpful.

### DIFF
--- a/adapter/ph/src/classifier.rs
+++ b/adapter/ph/src/classifier.rs
@@ -1,6 +1,5 @@
 #![allow(dead_code)]
 use crate::net_defs::*;
-use crate::packet;
 use arrayref::array_ref;
 use std::mem::size_of;
 use zerocopy::byteorder::network_endian::*;
@@ -93,6 +92,16 @@ pub struct ClassifierOptions {
     ignore_truncated_packets: bool,
 }
 
+#[derive(Default, Debug)]
+pub struct FiveTupleOptional {
+    pub src_address: Option<IpAddress>,
+    pub dst_address: Option<IpAddress>,
+    pub l3_type: Option<zpr::L3Type>,
+    pub l4_protocol: Option<IpProtocol>,
+    pub src_port: Option<u16>,
+    pub dst_port: Option<u16>,
+}
+
 impl ClassifierOptions {
     pub fn ignore_truncated_packets(mut self, value: bool) -> Self {
         self.ignore_truncated_packets = value;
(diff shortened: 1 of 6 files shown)


adapter/ph/src/classifier.rs
```diff
@@ -889,53 +935,53 @@ mod tests {
...
-            0x00, 0x00, 0xff, 0xff, 0x01, 0x02, 0x03, 0x04,
-            0xfc, 0x00, 0x00, 0x02, 0x00, 0x00, 0x00, 0x01,
-            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01,
-            0xa9, 0xa0, 0x1f, 0x90, 0x02, 0x1b, 0x63, 0x8c,
-            0x00, 0x00, 0x00, 0x00, 0xa0, 0x02, 0x67, 0x5c,
-            0x8e, 0xb9, 0x00, 0x00, 0x02, 0x04, 0x0b, 0x7c,
-            0x04, 0x02, 0x08, 0x0a, 0x80, 0x1d, 0xa5, 0x22,
-            0x00, 0x00, 0x00, 0x00, 0x01, 0x03, 0x03, 0x07
-        ];
+            let packet_data = [
```
cpacejo: Did you mean to indent these extra?

cpacejo (commented):
I don't think the FiveTupleOptional is a bad idea _per se_, but I'd suggest the simpler approach of accepting a `&mut FiveTuple` as an argument (essentially in place of the `&mut PacketMetadata` for most of the internal functions, and as an additional input to the public functions), and just writing into that directly.

You can then just make the `PacketMetadata::five_tuple` member `pub` (and get rid of the `five_tuple()` method on it), and pass a reference to it directly into the calls here.

This will preserve the behavior of returning "partial" five-tuples in cases where we error out partway, without complicating any of the return types or requiring the extra mechanism to copy over a "partial" five-tuple.  (At the risk of prematurely optimizing, the "partial" copy mechanism involves a lot of branching which would happen on every packet through the system, which in general is good to avoid.)

adapter/ph/src/classifier.rs
owenrhysthomas: I did not, and since you pointed this out, I actually found a few place when extra indentation had been added at a previous time.

cpacejo (approved):
Looks good, thanks!

PR #1108, opened 2025-11-11
fixed set_program function

### LINKED ISSUE #1103: play around with capture worker
capture worker currently sometimes fails because there is an address already in use

https://slack-files.com/T083SPP50-F09S2US4H8U-61e54315bc

owenrhysthomas:
Not 100% sure this will fix all the problems with the capture worker running, but I ran it locally many times and it did not fail. Regardless, it will fix a logic problem with the set_program function and ensure there is  no overlap in capture sockets.

I also added in a sleep and extended another one to align with the one node test, not sure if this is necessary.

### DIFF
--- a/integration-test/one-node-test.sh
+++ b/integration-test/one-node-test.sh
@@ -47,12 +47,15 @@ if [ ! -e "$VS_ADMIN_BIN" ]; then
 fi
 
 NODE_SOCK=node.sock
-CAP_SOCK=cap.sock
 VS_SOCK=vs.sock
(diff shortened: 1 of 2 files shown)


integration-test/one-node-test.sh
owenrhysthomas, line 115: the capture test does not have this. I am not exactly sure what it is used for in the VS, but I could go digging if we think it could be causing problems.

cpacejo (approved):
OK nice!!  Hopefully this will squash the issue :)

PR #1109, opened 2025-11-11
Mk/1073 - a start on libnode2 that can connect to v2vs

### LINKED ISSUE #1073: node connect to new visa service
Implement enough of libnode2 to get connect working.

Maybe do in parallel with -> https://github.com/org-zpr/zpr-core/issues/1110

### LINKED ISSUE #1110: Node connect
Implement enough of the VS/VSS APIs to permit a node to connect.  Node will run through the VS API steps:
- connect()
- challenge() to get a challenge then,
- authenticate() to have challenge authenticated
- Finally calls registerVss which for now can return "unimplemented"

Note
- Implies we can load a policy and ascertain that the node is allowed to connect.
- Does not mean we necessarily need to keep any state (yet).
- Bonus if ping() and disconnect() work.

I am working in parallel on the node side too => https://github.com/org-zpr/zpr-core/issues/1073

mkolehmainen:
This is a start on libnode2 -- a replacement for libnode that uses the Capn Proto scheme and our new VS-API and VSS-API.

So far this only has enough logic to do a node connect to the visa service (and a disconnect).

There is a toy dev tool this can also build named "ln2" that you can use to try a connect against a running visa service.


libnode2/src/bin/ln2.rs
svjaffer, line 1: I would probably call this something like "mock-node" instead of "ln2".  Something to be more clear in the name that this is a test tool

svjaffer (approved):
One comment on the naming, which doesn't need to be fixed in this PR.

libnode2/src/bin/ln2.rs
mkolehmainen: I was expecting this. ;)

PR #1114, opened 2025-11-14
New internal visa

### LINKED ISSUE #1106: Create serialization-agnostic internal Visa type
In libnode, we simply re-export the `vsapi::Visa` type from Thrift, and use that throughout the PH to hold visas.

We should instead create our own internal `Visa` type (which probably will look similar to the Thrift-generated one in libnode/vsapi/mod.rs, albeit with richer data types e.g. `SystemTime` for `expires`), and serialization/deserialization methods to/from Thrift, Cap'n Proto, and maybe JSON for debugging.

owenrhysthomas:
Chris and Sarah suggested that with the move from thrift to capnp, it might be a good time to create an internal visa type instead of relying on the thrift struct. I based the `Visa` mostly on the current thrift struct with some modifications based on what is in the capnp specification. The `VisaResponse` is mostly based on how the capnp `VisaResponse` is structured. 

Next I am going to make `TryFrom` implementations for the capnp version of the structure.


libnode/src/net_defs.rs
```diff
@@ -0,0 +1,245 @@
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
```
owenrhysthomas: The contents of this file have not been changed from when it was in `net_defs` in the PH, it was simply copy pasted so it could be shared between the libnode and ph

libnode/src/visa.rs
```diff
@@ -0,0 +1,398 @@
...
+    VisaParseError(u64, &'static str),
+    #[error("{0}")]
+    VisaHopError(&'static str),
+}
+
+#[derive(Debug)]
+pub enum VisaResponse {
+    Allow(Visa),
+    Deny(Denied),
+    Error(VisaResponseError),
```
mkolehmainen: This is really more of an API call error, not specific to visa-request.  Maybe better as "VsApiError" or something.

libnode/src/visa.rs
```diff
@@ -0,0 +1,398 @@
...
+    }
+}
+
+// TODO figure out which of these need to stay once we switch to capnp
+#[derive(Debug, Clone)]
+pub struct Visa {
+    pub issuer_id: u64, // i32 in thrift, u64 in capnp
+    pub config: i64,
+    pub expires: SystemTime,
+    // pub source: Vec<u8>,
```
mkolehmainen: `source` commented out?

libnode/src/visa.rs
```diff
@@ -0,0 +1,398 @@
...
+    pub bw: bool,
+    pub bw_limit_bps: i64,
+    /// empty/None means no data cap
+    pub data_cap_id: String,
+    pub data_cap_bytes: i64,
+    /// tether address of service actor
+    pub data_cap_affinity_addr: Vec<u8>,
+}
+
+pub enum EndpointT {
```
mkolehmainen: This endpoint thing is used in the tcp/udp pep.  But it's commented out there. Is it used somewhere else?

libnode/src/logging.rs
```diff
@@ -1,4 +1,7 @@
 pub mod targets {
     pub const VS_RPC: &str = "vs_rpc";
     pub const VSS_RPC: &str = "vss_rpc";
+    pub const V_STRUCTURE: &str = "v_structure";
```
mkolehmainen: I don't agree with this logging stuff.  The functions that log this should return errors and let the caller log it.

libnode/src/visa.rs
```diff
@@ -0,0 +1,398 @@
...
+}
+
+impl TryFrom<vsapi::VisaHop> for Visa {
+    type Error = VisaError;
+
+    fn try_from(hop: vsapi::VisaHop) -> Result<Self, Self::Error> {
+        match hop.visa {
+            Some(visa) => Visa::try_from(visa),
+            None => {
+                error!(target: VH_STRUCTURE, "No visa in VisaHop");
```
mkolehmainen: I don't think logging here (or in most places in this file) is appropriate.

libnode/src/visa.rs
```diff
@@ -0,0 +1,398 @@
...
+
+// Could also implement a TryFrom instead of picking arbitarty values
+impl TryFrom<vsapi::Visa> for Visa {
+    type Error = VisaError;
+
+    fn try_from(thrift_visa: vsapi::Visa) -> Result<Self, Self::Error> {
+        let issuer_id = match thrift_visa.issuer_id {
+            Some(val) => val as u64,
+            None => {
+                error!(target: V_STRUCTURE, "No issuer id");
```
mkolehmainen: I'm not going to flag the rest of these, but I would remove all this logging.

libnode/src/visa.rs
```diff
@@ -0,0 +1,398 @@
...
+            },
+            None => IpAddress::UNSPECIFIED,
+        };
+        let dock_pep = match thrift_visa.dock_pep {
+            Some(val) => {
+                match val {
+                    vsapi::PEPIndex::UDP => ip_number::UDP,
+                    vsapi::PEPIndex::TCP => ip_number::TCP,
+                    vsapi::PEPIndex::ICMP => ip_number::ICMP,
+                    _ => ip_number::UDP, // Not sure what default here should be, perhaps we want to make a UNSET ip number
```
mkolehmainen: This should return an error if the PEP is not an expected value.

libnode/src/visa.rs
```diff
@@ -0,0 +1,398 @@
...
+            None => ip_number::UDP, // Not sure what default here should be
+        };
+        let tcp_udp_pep = match thrift_visa.tcpudp_pep_args {
+            Some(val) => Some(TcpUdpPep::from(val)),
+            None => None,
+        };
+        let icmp_pep = match thrift_visa.icmp_pep_args {
+            Some(val) => Some(IcmpPep::from(val)),
+            None => None,
+        };
```
mkolehmainen, line 327: It should be an error if a PEP is defined -- which it must be, and the corresponding args are NOT defined.

libnode/src/visa.rs
```diff
@@ -0,0 +1,398 @@
...
+                error!(target: V_STRUCTURE, "No dest in visa with issuer id {issuer_id}");
+                return Err(VisaError::VisaParseError(issuer_id, "No dest"));
+            }
+        };
+        let src_addr = match thrift_visa.source_contact {
+            Some(val) => match IpAddress::try_from(val) {
+                Ok(addr) => addr,
+                Err(_) => IpAddress::UNSPECIFIED,
+            },
+            None => IpAddress::UNSPECIFIED,
```
mkolehmainen: Should be an error if these addresses are not set.

libnode/src/visa.rs
```diff
@@ -0,0 +1,398 @@
...
+            icmp_antecedent,
+        }
+    }
+}
+
+impl From<vsapi::KeySet> for KeySet {
+    fn from(thrift_key_set: vsapi::KeySet) -> Self {
+        let format = match thrift_key_set.format {
+            Some(val) => val,
+            None => 0,
```
mkolehmainen: Format must be set.

libnode/src/vss.rs
```diff
@@ -99,8 +100,16 @@ impl VisaSupportSyncHandler for VisaSupportHandlerImpl {
     fn handle_install_visas(&self, vh: Vec<vsapi::VisaHop>) -> thrift::Result<()> {
         debug!(target: VSS_RPC, "handle_install_visas, count={}", vh.len());
         for v in vh {
+            let visa = match Visa::try_from(v) {
+                Ok(visa) => visa,
+                Err(_) => {
```
mkolehmainen: Probably going to want to keep this error to log it here.

libnode/src/visa.rs
```diff
@@ -0,0 +1,398 @@
...
+                    Self::Deny(Denied::new(DenyCode::Fail, thrift_visa_response.reason))
+                }
+                val => Self::Error(VisaResponseError::new(
+                    ErrorCode::UnknownStatusCode,
+                    format!("Status code: {val:?}"),
+                    0,
+                )),
+            },
+            None => {
+                error!(target: VR_STRUCTURE, "No code in VisaResponse");
```
mkolehmainen: Avoid logging here I think

adapter/ph/src/compress.rs
```diff
@@ -5,6 +5,7 @@ use crate::defs::FiveTuple;
 use crate::net_defs;
 use crate::packet::Packet;
 use bytes::Buf;
+use libnode::net_defs as libnode_net_defs;
```
mkolehmainen: hmmm... Now we have two net_defs?  Those will be distinct types to rust. Which is sure to lead to headaches.  Maybe put netdefs into zpr crate and both libnode and ph can use that??

adapter/ph/src/mgmt/dock.rs
```diff
@@ -6,7 +6,7 @@ use crate::tc;
 use crate::visa_mgmt;
 use crate::visa_table::VisaTableError;
 
-use libnode::{vsapi, vsconn};
+use libnode::{visa, vsconn};
```
mkolehmainen: Just reading this I wonder if "visa" module should be called "vsapitypes" or something like that.

adapter/ph/src/mgmt/dock.rs
```diff
@@ -98,39 +98,37 @@ pub async fn bind_actor_address(
...
-                );
-            }
-
-            Ok(resp) => {
-                asm.counters.management[ManagementCounterType::VisaRequestDenied].increment();
-                debug!(target: FLOW_MGMT, "visa request rejected: {resp:?}");
-                return Err(BindActorAddressError::PolicyError);
-            }
+            Ok(visa_response) => match visa_response {
+                visa::VisaResponse::Allow(visa) => {
```
mkolehmainen: Again on the naming.  Above I suggested "vsapitypes", but maybe it should just be called "vsapi".  That would read nicely, `vsapi::VisaResponse`.

mkolehmainen: Ah dang, we already have vsapi in there for the generated api.  Ok, I'm back to "vsapitypes" or "vstypes". Or ?

adapter/ph/src/mgmt/dock.rs
```diff
@@ -98,39 +98,37 @@ pub async fn bind_actor_address(
...
+                    );
+                }
+                visa::VisaResponse::Deny(denied) => {
+                    asm.counters.management[ManagementCounterType::VisaRequestDenied].increment();
+                    debug!(target: FLOW_MGMT, "visa request denied: {:?}", denied.reason);
+                    return Err(BindActorAddressError::PolicyError);
+                }
+                // Not implemented as part of thrift visas
+                visa::VisaResponse::Error(error) => {
+                    asm.counters.management[ManagementCounterType::VisaRequestDenied].increment();
```
mkolehmainen: This is porbably not a VISA-DENIED error.  This is likely a visa parsing problem. A DENY is a valid VisaResponse.

adapter/ph/src/visa_mgmt.rs
```diff
@@ -180,27 +180,9 @@ pub async fn actor_disconnect(asm: Arc<Assembly>, addr: IpAddress) {
 /// Figure out egress link and insert visa into table.
 pub fn parse_visa(
```
mkolehmainen: This function is poorly named.  It is not a visa parsing routine.

adapter/ph/src/visa_table.rs
```diff
@@ -98,68 +98,37 @@ impl Visa {
...
-        };
-
-        let dst_addr = match &visa.dest_contact {
-            Some(dst) => match IpAddress::try_from(dst.clone()) {
-                Ok(addr) => addr,
-                Err(_) => IpAddress::UNSPECIFIED,
-            },
-            None => IpAddress::UNSPECIFIED,
-        };
+    pub fn extract_five_tuple(visa: &visa::Visa) -> FiveTuple {
```
mkolehmainen: I feel like five_tuple() or get_five_tuple() should be a function on Visa itself.  ?

adapter/ph/src/visa_table.rs
```diff
@@ -425,7 +388,7 @@ fn make_tcp_visa(
     dest_port: u16,
     configuration: i64,
     expiration_ms: i64,
-) -> vsapi::Visa {
+) -> visa::Visa {
     let pepargs = vsapi::PEPArgsTCPUDP {
```
mkolehmainen: Why are we using `vsapi` way down here?

mkolehmainen (changes requested):
Good first pass.  I do think needs a little more attention.

In general should be very inflexible w/ respect to parsing of the thrift.  Just return error if values we need are missing.

Need to deal with net_defs.

Possibly need a better module name.

libnode/src/visa.rs
owenrhysthomas: I left a comment but I just realized I started it as part of a review so it didn't show up. Most of the values that are commented out are values that are in the Thrift Visa but not used in the capnp version. The only exception is the EndpointT, which is in the capnp but not in the Thrift.

owenrhysthomas:
If you want to look at this commit-by-commit, which might be helpful, I will list the major commits and what happened at each stage:

1. **address comments:** the first round of changes I made, put the net_defs into zpr-utils, removed logging from Visa implementations, added in more errors
2. **moved custom Visa type to zpr crate:** as the name implies, this is where I moved the visa.rs file from libnode into vsapi_types in the zpr crate
3. a**ddress rest of feedback:** final round of addressing comments from the original review, moved extract_five_tuple to be a method on Visa, removed dest from Visa and replaced use of dest with dst_addr
4. **merge:** self explanatory
5. **formatting:** self explanatory
6. **fixed cargo lock:** fixed error that arose from merge
7. **removed IpAddress from five tuple lookup:** swapped it for IpAddr, created VsapiFiveTuple and associated methods
8. **merge:** self explanatory
9. **updated path in tun_pi:** macos only file needed path for net_defs updated
10.  **format tun_pi:** self explanatory
11. **update workflow files:** tests whether updated workflow worked with specific commit hash
12. **v1.1 in workflow instead of commit hash:** after determining that the workflow worked, it was merged, and tagged, then I updated the workflow file to use the tag
13. **removed IpProtocol from five_tuple_lookup:** created VsapiIpProtocol to eliminate five_tuple_lookup_table reliance on zpr-utils

mkolehmainen (approved):
Excellente!

PR #1115, opened 2025-11-14
bind request non sync work

cpacejo:
This is the first step of the final phase of work to move away from the async-request model for bind requests (#906), and toward an explicit state-machine model.  The changes herein are:

* We add a dummy link representing the internal dock on nodes.  We use the same fixed ID (2) we use on adapters to represent the dock link.  This is needed so that the internal adapter of a node can open a transaction to request a bind of the internal dock same as a standalone adapter connected to the external dock.  The transaction ID is the key to reference the state of an outstanding bind request in the adapter tables.
* We add a means to try to open a transaction without blocking.  This is useful when initiating a bind request, since we never want to wait for a free transaction; we always just want to drop the initial packet and give up if there's no free transaction.
* Tracking of pending bind requests moves from the peer state to the ALT.  (We still use the peer state to track egress tether requests, see TODO below.)
* Adapter-initiated bind requests are no longer async functions.  They now open a transaction, add it to the ALT, send a message, and return immediately.  Later, the bind response then looks up the state in the ALT by transaction ID, and installs the tether at that point.

The benefits of the explicit state-machine model are:
* better match to the unidirectional message model of ZDP
* no longer a need for every transaction to make temporary "one-shot" channels for waiting for results
* much easier to get insight into a running system to see what state everything is in (rather than just a giant bag of opaque Tokio-tasks)

Maybe easier to review commit-by-commit.

Work for future PRs includes:
* Either adding a wrapper layer so bind requests from internal adapters can also be state-machine-based, or reworking the internal dock bind code (which is currently called directly) so that they are so.  (This will then allow bind requests to be issued concurrently)
* Breaking up egress tether requests similarly.  (This will then allow us to remove the global transaction→one-shot-channel table.)

THEN we'll finally be able to tackle implementing the full routing flow correctly!

### DIFF
--- a/adapter/ph/src/adapter_manager_worker.rs
+++ b/adapter/ph/src/adapter_manager_worker.rs
@@ -1,4 +1,4 @@
-use crate::adapter_tables::{AltEntry, AltPep};
+use crate::adapter_tables::{self, AltPep};
 use crate::assembly::{Assembly, PhMode};
 use crate::counters::ManagementCounterType;
 use crate::logging::targets::FLOW_MGMT;
@@ -33,28 +33,10 @@ pub async fn launch(
 
 // RFC 6.5 § 6.3.11
 async fn do_request_tether_id(asm: &Arc<Assembly>, pkt: Packet) {
-    let five_tuple = pkt.metadata().five_tuple();
-
-    // if there's already an entry, this is a duplicate request
-    // (NOTE: we should be the only ones modifying this table!)
-    if asm.alt.get(five_tuple).is_some() {
-        mgmt::core::count_event(asm, ManagementCounterType::DroppedAwaitingBind);
-        return;
-    }
+    let dock_link_id = zpr::DOCK_LINK_ID;
 
     // copy out five tuple so we can give away packet
-    let five_tuple = *five_tuple;
-
-    let dock_link_id = match asm.ph_mode {
-        PhMode::Adapter => zpr::DOCK_LINK_ID,
-        PhMode::Node => zpr::LINK_ID_UNKNOWN,
-    };
-
-    if dock_link_id != zpr::LINK_ID_UNKNOWN && !asm.is_link_ready(dock_link_id) {
-        debug!(target: FLOW_MGMT, "Link {dock_link_id} is not ready to receive traffic yet");
-        mgmt::core::count_event(asm, ManagementCounterType::DroppedNoSA);
-        return;
-    }
+    let five_tuple = *pkt.metadata().five_tuple();
 
     if five_tuple.dst_address.is_v6_unicast_link_local()
         || five_tuple.src_address.is_v6_unicast_link_local()
@@ -63,11 +45,42 @@ async fn do_request_tether_id(asm: &Arc<Assembly>, pkt: Packet) {
         return;
     }
 
+    let Some(peer_state) = asm.peer_table.get(dock_link_id) else {
+        // Link was dropped
+        return;
+    };
+
+    if !peer_state.link_state_machine.is_ready() {
+        debug!(target: FLOW_MGMT, "Link {dock_link_id} is not ready to receive traffic yet");
+        mgmt::core::count_event(asm, ManagementCounterType::DroppedNoSA);
+        return;
+    }
+
+    // try open a transaction on the link; if we can't due to backpressure, don't wait, just give up
+    let Some(txn) = peer_state.txn_mgr.try_open() else {
+        debug!(target: FLOW_MGMT, "link {dock_link_id}: backpressure on link prevents issuing bind request for {five_tuple}");
+        mgmt::core::count_event(asm, ManagementCounterType::QueueBackpressure);
+        return;
+    };
+
+    // copy out packet body
     let packet_body = pkt.body().to_owned();
 
     // mark ALT entry as pending to attempt to (i.e. racily) prevent
     // fastpath from issuing multiple requests
-    asm.alt.insert(five_tuple, AltEntry::Pending(pkt));
+    match asm.alt.insert_pending(five_tuple, pkt, &txn) {
+        Ok(()) => (),
+
+        Err(adapter_tables::InsertPendingError::AlreadyPending(_pkt)) => {
+            // there's already an entry; this is a duplicate request
+            mgmt::core::count_event(asm, ManagementCounterType::DroppedAwaitingBind);
+            return;
+        }
+
+        Err(adapter_tables::InsertPendingError::DuplicateTransaction(_pkt)) => {
+            panic!("duplicate transaction")
+        }
+    }
 
     debug!(target: FLOW_MGMT, "link {dock_link_id}: Issuing bind request for {five_tuple} (is now set PENDING)");
 
@@ -103,33 +116,19 @@ async fn do_request_tether_id(asm: &Arc<Assembly>, pkt: Packet) {
                     .five_tuple()
             {
                 error!(target: FLOW_MGMT, "Bind of {five_tuple} falied: node supplied TC incompatible with initial packet: {tc}");
-                asm.alt.remove(&five_tuple);
+                asm.alt.remove(&five_tuple).unwrap();
                 return;
             }
 
             // Bind succeeded; add to ALT.
             debug!(target: FLOW_MGMT, "Bind of {five_tuple} succeeded: {tether_id}");
 
-            let AltEntry::Pending(initial_packet) = asm
-                .alt
-                .alter(&five_tuple, |entry| {
-                    assert!(
-                        matches!(entry, AltEntry::Pending(_)),
-                        "coding error: race to activate pending ALT entry"
-                    );
-
-                    std::mem::replace(
-                        entry,
-                        AltEntry::Active(AltPep {
-                            compression_mode: tc.compression_mode(),
-                            tether_id,
-                        }),
-                    )
-                })
-                .unwrap()
-            else {
-                unreachable!();
+            let pep = AltPep {
+                compression_mode: tc.compression_mode(),
+                tether_id,
             };
+
+            let initial_packet = asm.alt.set_active(&five_tuple, pep).unwrap();
 
             // now send out initial packet
             match asm.actor_output_requeue.try_enqueue_packet(initial_packet) {
@@ -144,7 +143,7 @@ async fn do_request_tether_id(asm: &Arc<Assembly>, pkt: Packet) {
         Err(err) => {
             // Bind failed; remove pending entry from ALT.
             debug!(target: FLOW_MGMT, "Bind of {five_tuple} failed: {err}");
-            asm.alt.remove(&five_tuple);
+            asm.alt.remove(&five_tuple).unwrap();
         }
     }
 }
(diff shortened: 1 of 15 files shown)


adapter/ph/src/adapter_tables.rs
```diff
@@ -22,12 +24,13 @@ pub struct AltPep {
 }
 
 pub enum AltEntry {
+    Pending(Packet, TxnHandle),
     Active(AltPep),
-    Pending(Packet),
 }
 
 pub struct ActorLookupTable {
```
mkolehmainen, line 39: Maybe outside the scope of this PR, but I would like to know what this actor-lookup-table is for.  What does "pending" mean in this context?

adapter/ph/src/adapter_tables.rs
```diff
@@ -40,13 +43,103 @@ impl std::ops::Deref for AltEntryGuard<'_> {
...
+    NotFound,
+    NotPending,
+}
+
+#[derive(Debug)]
+pub enum RemoveError {
+    NotFound,
+}
+
+pub enum SetActiveError {
```
mkolehmainen, line 79: What is "active" here?

mkolehmainen: Is it bind requests? If so why are bind-requests logically in an Actor Lookup Table?

adapter/ph/src/adapter_manager_worker.rs
```diff
@@ -33,118 +32,91 @@ pub async fn launch(
...
 
+    // copy out packet body
     let packet_body = pkt.body().to_owned();
 
+    let txn_id = txn.id();
+
     // mark ALT entry as pending to attempt to (i.e. racily) prevent
     // fastpath from issuing multiple requests
-    asm.alt.insert(five_tuple, AltEntry::Pending(pkt));
+    match asm.alt.insert_pending(five_tuple, pkt, &txn) {
```
mkolehmainen, line 72: Ah, ok. This answers an earlier question. So these are pending bind requests.

adapter/ph/src/main.rs
```diff
@@ -581,23 +581,34 @@ fn main() -> ExitCode {
...
+                )
+                .unwrap();
+
+            assert_eq!(dsid.get(), zpr::DOCK_LINK_ID);
+        }
 
-        assert_eq!(dsid.get(), zpr::DOCK_LINK_ID);
+        PhMode::Node => assert_eq!(
+            asm.peer_table.insert_internal_peer().get(),
+            zpr::DOCK_LINK_ID
```
mkolehmainen, line 610: hmm. Does DOCK_LINK_ID make sense for the node here? Why do we care so much what the link ID is now that there seems to be  "type" floating around of LinkType::Internal?

adapter/ph/src/mgmt/adapter.rs
```diff
@@ -0,0 +1,95 @@
...
+use crate::tc;
+use tracing::*;
+use zpr::StreamId;
+
+#[derive(Debug)]
+pub enum InstallTetherError {
+    NoSuchTransaction,
+}
+
+/// Install the given tether into the ALT.
```
mkolehmainen, line 16: I wish there was more documentation of the ALT.  I don't get why "tethers" belong in an "actor" lookup table. More info about what the purpose of the ALT is and what sorts of things it holds would be helpful for understanding things here.  This is probably because I've been working with visa service where Actors are different I guess.

adapter/ph/src/mgmt/adapter.rs
```diff
@@ -0,0 +1,95 @@
...
+    tc: tc::Ip5TupleTc,
+) -> Result<(), InstallTetherError> {
+    let five_tuple = asm
+        .alt
+        .lookup_pending(txn)
+        .map_err(|_| InstallTetherError::NoSuchTransaction)?;
+
+    // Confirm this TC matches our initial packet.
+    // TODO: use the TC itself to do this, once we have that code in place.
+    if tc.five_tuple()
```
mkolehmainen, line 34: This seems like a funny check. Why don't we take the nodes word for it here?

adapter/ph/src/mgmt/adapter.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    // Bind succeeded; add to ALT.
+    debug!(target: FLOW_MGMT, "Bind of {five_tuple} succeeded: {tether_id}");
+
+    let pep = AltPep {
+        compression_mode: tc.compression_mode(),
+        tether_id,
+    };
+
+    let Ok(initial_packet) = asm.alt.set_active(&five_tuple, pep) else {
```
mkolehmainen, line 50: Why do all errors from set_active imply a "duplicate bind request" ... and then we end up returning a "no-such-transaction". Feels like we are dropping information here.

adapter/ph/src/mgmt/handlers.rs
```diff
@@ -1056,13 +1057,54 @@ pub async fn handle_bind_actor_address_response(
...
-    let Some(sender) = peer_state.bind_req_state.lock().unwrap().remove(&txn) else {
-        return Err(HandleMgmtError::UnknownTransaction(txn_id));
+    let Ok(hdr) = zdp::ZdpBindActorAddressResponseHeader::read_from_buf(&mut pkt) else {
+        return Err(HandleMgmtError::BadStructure);
     };
 
-    let _ = sender.send(pkt);
+    match hdr.status_code {
+        zdp::ResponseCode::Success => {
+            let stream_id = pkt.metadata().ingress_stream_id;
```
mkolehmainen, line 1067: I'm just going to yell into the void here that "Tcst" is a terrible name for a type.

zpr/src/lib.rs
```diff
@@ -38,7 +38,8 @@ pub const LINK_ID_UNKNOWN: LinkId = 0;
 /// Link ID used to refer to a node or adapter's local actor.
 pub const LOCAL_ACTOR_LINK_ID: LinkId = 1;
 
-/// Link ID used on an adapter to refer to the dock to which it's connected.
+/// Link ID used on an adapter to refer to the dock to which it's connected,
+/// or on a node to refer to the node's internal dock.
 pub const DOCK_LINK_ID: LinkId = 2;
```
mkolehmainen, line 43: I did comment on this earlier.  
Doesn't LOCAL_ACTOR_LINK_ID feel like the better ID for a node to use to refer to its own internal dock?

mkolehmainen (dismissed):
Various comments included! Nothing blocking merging of this PR. I think using state machine is the right way to go.

adapter/ph/src/adapter_tables.rs
cpacejo: This is what adapters use to map a 5-tuple to a stream ID.  "Pending" here means, we've asked for a stream ID but haven't gotten one yet.

adapter/ph/src/adapter_tables.rs
cpacejo: Yes, "active" means we've acquired a stream ID and can now send packets.

These are outbound bind requests from the adapter; so every stream is either "pending" (waiting for the bind request to complete) or "active" (we have a stream ID).

adapter/ph/src/adapter_manager_worker.rs
cpacejo: Correct!  I should sprinkle some comments in here, I always forget to!

adapter/ph/src/main.rs
cpacejo: I deliberated on this for a while but this ended up working out well architecturally.

The problem is the "node" is really, a node _and also_ an (internal) adapter.  Nodes have peers (other nodes and adapters) and adapters also have peers (the dock and the actor pseudo-peer).  So on a node we now have this problem of sharing the peer table between the node and the internal adapter.

Note that peer IDs do double duty, they're used both for indexing peer state in the peer table, and also for forwarding.

We could (and maybe ultimately should?) solve this problem on the node by having two separate tables, one for the node proper, and one for its internal adapter, and treating the internal adapter like an external one for forwarding purposes.  But assigning DOCK_LINK_ID as an internal peer also solves this problem as follows:
* The node proper continues to use LOCAL_ACTOR_ID for forwarding traffic to the internal adapter.
* The node proper ignores DOCK_LINK_ID.
* The node+internal adapter continue to act as a single unit for forwarding purposes.  (Actor traffic is forwarded between external peers, and the LOCAL_ACTOR_ID.  DOCK_LINK_ID does not enter into forwarding decisions on the node.)
* The change: **the internal adapter can now use the DOCK_LINK_ID peer state to store transactions which it opens with the internal dock.**

We care what the link ID specifically is for two reasons:
1. So we can directly access the one-and-only instance of the dock peer in the peer table without doing, say, a special-peer lookup (though, we could in fact do this).
2. To keep the code consistent between adapter and node.

Again, in the future it might make sense to in fact architecturally separate the internal adapter to a greater degree (which is in fact suggested in RFC 6).  But that's a much larger change.

adapter/ph/src/mgmt/adapter.rs
cpacejo: It's a concept from RFC 6, which is where the naming comes from.  I notice now in RFC 17 it's been changed to Endpoint Lookup Table (same function though).  I'll open a ticket to change it (and also add some documentation in the code).

adapter/ph/src/mgmt/adapter.rs
cpacejo: Just the general principle of not trusting external processes to maintain internal invariants.  (We want to make sure we're not stomping some other flow's forwarding tables… there could even be race conditions or such if we did.  So this is a quick & easy check to ensure that.)  Makes the code easier to vet in isolation.

adapter/ph/src/mgmt/adapter.rs
cpacejo: Duplicate bind _response_, not request.  The only way for a transaction to have exited pending (i.e., either error from `set_active()`) at this point (after our earlier lookup of the 5t) is because we got some other response to this same transaction which caused this code to be run twice concurrently.  So from a serialization perspective, the first response ended the transaction, and now this 2nd response is for a non-existent transaction.  So the error message is essentially inferring that a duplicate response must be the cause.

I'll add a comment with the above reasoning.

adapter/ph/src/mgmt/handlers.rs
cpacejo: I don't disagree :)  Just mirroring what's in the RFC.

zpr/src/lib.rs
cpacejo: I think my other response should cover this – but TLDR, this is from the point of view of the internal adapter.  `DOCK_LINK_ID` refers to the "dock" end of an adapter, and  `LOCAL_ACTOR_LINK_ID` refers to the "actor" end of an adapter.  (These are both now true on both node-internal and standalone/external adapters.)

The real issue is that we use a single link ID namespace and tablespace for both the node-proper and its internal adapter.  From the point of view of the node-proper, it does in fact use `LOCAL_ACTOR_LINK_ID` to refer to the internal dock (or, at least, to forward packets in that direction), and never references `DOCK_LINK_ID`.

We could/maybe should revisit this in the future, to separate the internal adapter from the node-proper more cleanly.

mkolehmainen (approved):
Excellent!

PR #1136, opened 2025-11-25
Mk/1128 ln2 vreq - add ability to issue a visa-request over the VSAPI using testing tool

mkolehmainen:
Updated the test tool to allow user to issue a visa request to a visa service.  Also renamed the test tool from 'ln2' to 'lntest'.

Includes yet another version of vsapi_types. New story to fix this and put the types into the zpr crate, see
https://github.com/org-zpr/zpr-core/issues/1137


libnode2/src/bin/lntest.rs
```diff
@@ -0,0 +1,293 @@
...
+    );
+
+    let handle = vsc.handle();
+
+    let local_set = tokio::task::LocalSet::new();
+    let _local_set_guard = local_set.enter();
+    let mut js = JoinSet::new();
+
+    js.spawn_local(async move {
+        info!("starting the VSConn");
```
owenrhysthomas, line 129: In the PH and other modules, we have targets on the logging, maybe that would be helpful here, too

owenrhysthomas (approved):

libnode2/src/bin/lntest.rs
mkolehmainen: I do use that in the vs.  This is just a cli tool.

owenrhysthomas: Ah, right!

PR #1139, opened 2025-12-02
Mk/1137 types

mkolehmainen:
Move all the various api type code that had been writen for libnode2 and visa service into the zpr crate.

Will require a new tag (zpr-0.3.0) once this goes through.  ( https://github.com/org-zpr/zpr-visaservice/pull/91 ).

Note I feel like we need a cleanup pass on this soon.  For one thing, the use of "source", "src", "addr", "address" is inconsistent. ( https://github.com/org-zpr/zpr-core/issues/1141 ).


zpr/src/vsapi_types.rs
```diff
@@ -80,6 +101,139 @@ pub enum ErrorCode {
...
+}
+
+impl TryFrom<v1::packet_desc::Reader<'_>> for PacketDesc {
+    type Error = VsapiTypeError;
+    fn try_from(reader: v1::packet_desc::Reader<'_>) -> Result<Self, Self::Error> {
+        let src_ip = reader.get_source_addr()?;
+        let source = match src_ip.which().unwrap() {
+            v1::ip_addr::V4(ipv4) => {
+                let octets: [u8; 4] = ipv4?
+                    .try_into()
```
owenrhysthomas: The `try_into` here produces `Result<[u8; 4], TryFromSliceError>`. `VsapiTypeError` implements has `TryFromSliceError(#[from] std::array::TryFromSliceError)`, so I think you can replace the `map_err` with just a `?`, unless you want it to stay as a serialization error

zpr/src/vsapi_types.rs
```diff
@@ -343,6 +523,18 @@ impl Visa {
...
+    pub fn get_expiration_timestamp(&self) -> u64 {
+        match self.expires.duration_since(UNIX_EPOCH) {
+            Ok(dur) => dur.as_millis() as u64,
+            Err(_) => 0,
+        }
+    }
+}
+
+pub fn visa_expiration_timestamp_to_system_time(timestamp: u64) -> SystemTime {
+    UNIX_EPOCH + Duration::from_millis(timestamp)
```
owenrhysthomas, line 530: There are several places that this exact calculation is done in this file (see lines 546, 607, and more) this func could be used instead. I could also make that change when I add doc comments (see #1140)

zpr/src/vsapi_types.rs
```diff
@@ -367,7 +559,9 @@ impl TryFrom<v1::visa::Reader<'_>> for Visa {
 
         let dock_pep = DockPep::try_from(reader.get_dock_pep()?)?;
         let session_key = KeySet::try_from(reader.get_session_key()?)?;
-        let cons = Constraints::default();
+
+        // TODO: constaints not yet implemented.
```
owenrhysthomas: Typo: const**r**aints

owenrhysthomas (dismissed):
Nice! Just a few small notes, could be addressed in a later PR, especially since both #1140 and #1141 are doing cleanup on this file.

zpr/src/vsapi_types.rs
mkolehmainen: Oh right, we haven't implemented "constraints" or "constaints"!

owenrhysthomas (approved):
Nice! :fire:

PR #1142, opened 2025-12-02
refactor a bit

mkolehmainen:
Broke apart the vsapi_types file into a bunch of smaller ones, keeping public interface the same.  

I think this is a small improvement.  

In future I'd like to add convert/thrift & convert/capnp to separate the serialization code.

Possibly should move vsapi_types_writers into convert aswell.


owenrhysthomas (dismissed):
This looks good. The one thing is I think it is a little bit confusing at this point to have `connection.rs` be separate from `response.rs`. I understand that in the capnp version the `ConnectionResponse` is removed and replaced with `Connection`, which is why I named it that here, but it functions as a response essentially. Alternatively, if you think connection should be kept separate, then maybe `ConnectRequest` should be with it too, instead of in it's own separate file.

It just seems a bit confusing to me to have `request.rs` and `ressponse.rs` but then not have the corresponding "response" to `ConnectRequest` in a different file

owenrhysthomas:
Also - how come the github runners didn't run

mkolehmainen:
Ok thanks for looking. I will look to organize connection better.  I was more or less on auto-pilot doing the refactor, just trying to group things.  I'll look at the workflow file too to see if changing the paths has stopped something from triggering.

owenrhysthomas (approved):
👍

PR #1157, opened 2025-12-05
zpr cleanup

### LINKED ISSUE #1141: clean up of zpr-core/zpr
One annoyance is inconsistent use of terms like "source", "src", "dest", "dst", "addr", "address", etc.

Also the lib.rs file is full of oddities and should be split into meaningful modules.

Once this is done (either are part of this or create another story) we must add a new tag and update the usage of the repo.  As far as I know usage includes:
- zpr-core/adapter/ph
- zpr-core/libnode2
- zpr-visaservice/vs
- zpr-visaservice/libeval
- zpr-visaservice/zpt

owenrhysthomas:


owenrhysthomas:
I changed all the src/dst/addr/address to source/dest_addr in zpr. The FiveTuple in adapter/ph still uses src/dst_address, should we change those too?

mkolehmainen (approved):
Looks great! Thanks!

PR #1158, opened 2025-12-05
Mk/update vsapi 0.8.0

mkolehmainen:
Previously updated zpr-vsapi to fix a typo.

I have gone in here and changed HEAD of the submodule (I think).
And then changed the old references to the typo.

Once this goes through we can add a new zpr-xxx tag and use it from other crates.


cpacejo (approved):
Technically this requires a major version bump on `zpr`.  But since the only uses of `set_enpoint()` are also addressed in this changeset this is fine!

PR #1163, opened 2025-12-09
Ort/remove panic from zpr

### LINKED ISSUE #1145: remove (or document as appropriate) all panics in zpr-core/zpr
Make sure there are no panics in zpr-core/zpr.

- Get rid of all unwraps unless panic is really only option and if so document that in function docs, or rename with an `_unchecked` or similar.
- Consider `Result<>` instead of panic

owenrhysthomas:
Removed all unwraps from zpr, other than the unwraps in the unit tests in mod, because we necessarily know the result (and would want the test to panic if for some reason we were wrong).

I removed unwraps that were unchecked and could have potentially caused a panic, and I also removed unwraps that came after `if val.is_some()` and replaced with either with an `if let`, or a `match` just to avoid any confusion when scanning the code that there may be an unchecked unwrap.

TODO: rebase -i once #1157 has been approved and merged to eliminate unnecessary commits

### DIFF
--- a/zpr/src/vsapi_types/error.rs
+++ b/zpr/src/vsapi_types/error.rs
@@ -1,3 +1,5 @@
+use std::net::AddrParseError;
+
 use thiserror::Error;
 
 /// Error type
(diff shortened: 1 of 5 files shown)


mkolehmainen (approved):
Amazing!

PR #1165, opened 2025-12-09
Mk/fixln2

mkolehmainen:
Broken with recent changes to zpr.  Patched up and now should run its test when zpr changes.


owenrhysthomas (approved):
Will this put - zpr/** into .github/workflows/libnode.yml twice? Since #1164 was merged in and added them?

PR #1167, opened 2025-12-11
removed option Visa and ftuple from visa_table

### LINKED ISSUE #1147: visa_table::Visa::visa doesn't need to be Option
Seems like it's never None.  Per @svjaffer: "It might be left over from bootstrap visa stuff"

owenrhysthomas:


cpacejo (approved):
Nice hooray simplification!

PR #1168, opened 2025-12-12
update to zpr-policy 0.6.7

mkolehmainen:
I believe this now pulls in v0.6.7 of zpr-policy.  Once this goes in I'll tag zpr crate so that I can use the new features in the compiler and visa service.

This also now includes "policy_types" which are a bunch of type code from the compiler that will also be useful in the visa service.

### DIFF
--- a/zpr/src/policy_types/attribute.rs
+++ b/zpr/src/policy_types/attribute.rs
@@ -236,7 +236,7 @@ impl Attribute {
         } else {
             write!(f, "{}", key).unwrap();
             if self.is_multi_valued() {
-                write!(f, "{}", "{}").unwrap();
+                write!(f, "{{}}").unwrap();
             }
             if self.optional {
                 write!(f, "?").unwrap();
@@ -270,7 +270,7 @@ impl Attribute {
     ///
     /// ## Panics
     /// - if passed `name` does not start with `zpr`.
-    pub fn must_zpr_internal_attr<S: Into<String>>(name: S, value: S) -> Self {
+    pub fn must_zpr_internal_attr<S: Into<String>, T: Into<String>>(name: S, value: T) -> Self {
         if let Some(name_without_domain) = name
             .into()
             .strip_prefix(&format!("{}.", ATTR_DOMAIN_ZPR_INTERNAL))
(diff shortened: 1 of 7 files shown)


zpr/src/policy_types/attribute.rs
```diff
@@ -0,0 +1,580 @@
...
+                write!(f, "{key}:").unwrap();
+            } else if v.len() == 1 {
+                write!(f, "{key}:{}", v[0]).unwrap();
+            } else {
+                write!(f, "{key}:{{{}}}", v.join(", ")).unwrap();
+            }
+        } else {
+            write!(f, "{}", key).unwrap();
+            if self.is_multi_valued() {
+                write!(f, "{}", "{}").unwrap();
```
cpacejo: `write!(f, "{{}}")` does what you want here!  Or `.push_str()` like you do elsewhere.

zpr/src/policy_types/attribute.rs
```diff
@@ -0,0 +1,580 @@
...
+            write!(f, "{}", key).unwrap();
+        }
+        f
+    }
+
+    /// Special constructor for ZPR internal attributes with a single value.
+    ///
+    /// ## Panics
+    /// - if passed `name` does not start with `zpr`.
+    pub fn must_zpr_internal_attr<S: Into<String>>(name: S, value: S) -> Self {
```
cpacejo: I don't understand this function name?

cpacejo: Maybe want separate types S and T for the two parameters?  (E.g. so `name` can be a `String` and `value` can be some other thing)

cpacejo (dismissed):
Nice looks good!

zpr/src/policy_types/attribute.rs
mkolehmainen: I use must_ to indicate it will panic.  Maybe "_unchecked" is more rust-y?  I'll clean this up in another PR.

mkolehmainen: Oh good idea.

cpacejo (approved):
Looks good!

PR #1179, opened 2026-01-12
bind request rework

### LINKED ISSUE #741: Make visa requests non-blocking
Currently, when the management path needs to make a visa request, it blocks on it.  This means that no other management packets can be handled, blocking further visa requests or keep-alives or link management messages.

### LINKED ISSUE #1119: Rework route-request process as state machine
The route-request process is currently a couple large async blocks (`mgmt::dock::bind_actor_address()` and `assembly::Assembly::add_route()`).  We want to transform these into a state machine whose state is kept per-link per-bind-request likely in the Forwarding Table.

For ingress bind requests from adapters, there are essentially three states:

* Awaiting Visa Response
* Awaiting Next-Hop Bind (next-hop could be egress adapter, or next-node)
* Active

We should keep space to support bind requests from nodes (though those currently don't exist).  They would have an additional state, Awaiting Visa Push, which is used instead of Awaiting Visa Response.

Note that visa requests themselves will remain as async tasks for the forseeable future.

Converting egress binds from async tasks to state machines is a separate PR, blocked by this one.

### LINKED ISSUE #1120: Rework egress bind requests as state machine
Like #906, but for egress binds.  Blocked by #1119.

cpacejo:
OK!!!  This has been a long time coming.  These changes completely rework how bind requests work internally, they're now a state machine instead of a random sequence of things happening in an async block.

My two main goals with this rewrite are (a) make everything legible and easy to understand, (b) allow us to easily integrate multi-hop binds, and (c) fix shortcomings of the current bind request flow.

There's a couple of changes that still need to be made but those will come in separate commits.

Sorry this one is so big!  I will try my best to explain the changes below.

Minor changes:

* The PFT (in `forwarding_tables`) was given the ability to reserve an entry (to reference the ID) before adding it.
* The bind request handler now validates that the requested source address actually belongs to the adapter making the request.
* The node's internal adapter link now knows its own address (needed so we don't need to special-case the code which validates bind request source addresses).
* Inserting a visa into the visa table now just returns the visa ID, not visa ID + egress link.
* Added a function to get the traffic classifier of a visa.  This "should" be part of the visa itself, but for now we synthesize it from the visa's 5-tuple.
* Added some utility functions for converting to/from references to different IP address formats.
* The ZDP bind request send function now requires the caller to open a transaction on the link.
* The ZDP bind response send function is now broken into separate success and error variants.

Major changes:

* The ZDP bind request and response handlers now just immediately forward work on to the appropriate functions in the `mgmt::dock` and `mgmt::adapter` modules.  They no longer block, mess about with tables, or contain long complicated sequences of async logic.
* The logic from the ZDP bind actor address handler and the `assembly::add_route()` method are now restructured into a state machine defined by functions in the `mgmt::dock` module, which now contains all the logic for managing internal dock state.
* The `bind_req_state` table in the peer table is replaced with the dock state, which in turn contains a table containing active bind-request state machines.
* The logic from the ZDP bind egress stream handler is moved into `mgmt::adapter::bind_egress_stream()`.  So `mgmt::adapter` now contains all the logic for managing internal adapter state.

### DIFF
--- a/adapter/ph/src/mgmt/adapter.rs
+++ b/adapter/ph/src/mgmt/adapter.rs
@@ -1,12 +1,124 @@
-use super::txn_mgr;
-use crate::adapter_tables::EltPep;
-use crate::assembly::Assembly;
+//! Adapter API.
+//!
+//! These functions operate on the local adapter state at a high level.
+//!
+//! They are meant to be invoked either from [super::handlers] by an
+//! adapter in response to a message from a node, or directly by a node
+//! managing its local adapter.
+
+use super::{dock, txn_mgr};
+use crate::adapter_tables::{DltPep, EltPep};
+use crate::assembly::{Assembly, PhMode};
 use crate::counters::ManagementCounterType;
 use crate::logging::targets::FLOW_MGMT;
 use crate::queues;
 use crate::tc;
+use std::num::NonZero;
+use std::sync::Arc;
 use tracing::*;
-use zpr::packet_info::StreamId;
+use zpr::packet_info::{LOCAL_ACTOR_LINK_ID, LinkId, StreamId};
+
+/// Request to bind a stream which egresses from the local adapter.
+///
+/// `dock_link_id` identifies the link of the dock making this request.
+/// (This is currently always [zpr::packet_info::DOCK_LINK_ID]; on adapters,
+/// this represents the remote dock; on nodes, this represents the internal
+/// dock.)
+///
+/// `txn_id` is used when issuing responses to the dock to link them to
+/// this request.
+///
+/// `tc` is the traffic classifier for which the request is being made.
+///
+/// The bind is performed simply by inserting a PEP corresponding to the
+/// given TC into the DLT, generating a tether ID.
+///
+/// A success response (containing the tether ID) or error response (in
+/// the event that the DLT is full) is then issued to the dock, where these
+/// responses are handled by [`dock::install_tether()`] and [`dock::deny_tether()`]
+/// respectively.
+pub fn bind_egress_stream(
+    asm: &Arc<Assembly>,
+    dock_link_id: NonZero<LinkId>,
+    txn_id: txn_mgr::TxnId,
+    tc: tc::Ip5TupleTc,
+) {
+    debug!(
+        target: FLOW_MGMT,
+        "bind_egress_stream(dock_link_id={dock_link_id}, txn_id={txn_id}, tc={tc})");
+
+    // form PEP
+    let pep = DltPep {
+        compression_mode: tc.compression_mode(),
+        five_tuple: *tc.five_tuple(),
+    };
+
+    // TODO: reverse path
+
+    // attempt to insert into DLT and respond to requestor
+    match asm.dlt.insert(pep) {
+        Ok(tid) => {
+            // TODO: maybe tick a counter somewhere?
+            match asm.ph_mode {
+                PhMode::Node => {
+                    // we're a node operating on our internal adapter; "respond" directly to local dock
+                    let dock_peer_state = asm.peer_table.get(LOCAL_ACTOR_LINK_ID).unwrap();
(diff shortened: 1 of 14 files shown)


adapter/ph/src/mgmt/adapter.rs
```diff
@@ -1,12 +1,124 @@
...
+    // attempt to insert into DLT and respond to requestor
+    match asm.dlt.insert(pep) {
+        Ok(tid) => {
+            // TODO: maybe tick a counter somewhere?
+            match asm.ph_mode {
+                PhMode::Node => {
+                    // we're a node operating on our internal adapter; "respond" directly to local dock
+                    let dock_peer_state = asm.peer_table.get(LOCAL_ACTOR_LINK_ID).unwrap();
+                    let Some(txn) = dock_peer_state.txn_mgr.get(txn_id) else {
+                        // adapter no longer has this transaction; ignore
```
owenrhysthomas: At some point, we would presumably want to edit the TxnMgr::get() function to no longer return an option if this transaction no longer exists

adapter/ph/src/mgmt/dock.rs
```diff
@@ -1,165 +1,708 @@
...
+pub fn bind_actor_address(
     asm: &Arc<Assembly>,
     ingress_link_id: NonZero<LinkId>,
-    five_tuple: &FiveTuple,
+    txn_id: TxnId,
     packet_body: &[u8],
-) -> Result<(StreamId, tc::Ip5TupleTc), BindActorAddressError> {
-    let mut forwarding_decision: Option<ForwardingDecision> = None;
-
+) {
```
owenrhysthomas, line 115: Would it be helpful to have a return type on all of these functions, or would it just be too confusing?

adapter/ph/src/mgmt/handlers.rs
```diff
@@ -743,9 +764,14 @@ fn parse_grant_zpr_address_request(
 /// handle a Bind Actor Address Request (RFC 6.5 § 6.3.11)
 pub async fn handle_bind_actor_address_request(
     asm: &Arc<Assembly>,
-    txn_id: u16,
+    txn_id: txn_mgr::TxnId,
```
owenrhysthomas: you import `super::txn_mgt::TxnId` at the top, I think you can just use `TxnId` here, as you do in `handle_bind_egress_stream_request`

owenrhysthomas (dismissed):
Nice! The state machine is great

adapter/ph/src/mgmt/dock.rs
cpacejo: The intent is that they "cannot fail", in that they are signals to the state machine, many of which come from a remote system.  So hence the return type is implicitly the unit type `()`.

If the state transition fails, we signal it by sending an error message to the requestor, and entering the failure state (i.e. terminating the state machine).  But the act of telling the state machine that this signal arrived always succeeds.

adapter/ph/src/mgmt/adapter.rs
cpacejo: I'm not sure what you mean.  `get()` returns `Some(txn)` if it finds the transaction, and `None` if the transaction no longer exists.  In general we have to expect we might see `None` and handle it somehow, since the transaction ID is coming from a remote system and could be "wrong" for reasons outside of our control.

BUT, this particular case is internal adapter replying to the local dock, so there is no "remote system" to be "wrong".  If we do see `None` here, _we've_ made a mistake, and should probably panic rather than ignore it.  Lemme change this (and elsewhere) to do that…

adapter/ph/src/mgmt/handlers.rs
cpacejo: Ah good catch!

owenrhysthomas (approved):

svjaffer (approved):
Appreciate the TODOs!

PR #1189, opened 2026-01-22
Cleanup handlers.rs

### LINKED ISSUE #1171: clean up PH mgmt::handlers
Let's move all "smarts" out of this module.  These handlers should just decode the packet and call out to corresponding APIs elsewhere in the PH.  See `handle_bind_actor_address_request()` and `handle_bind_egress_stream_request()` and their corresponding `response()` functions as examples of what we should do.  `handle_hello_request()` is an example of a handler which is probably doing too much.

cpacejo:
These three changes finalize my push to move all the "smarts" out of handlers.rs and into modules associated with the things being controlled.  We do three things here:

1) Clean up link state error handling (it was done a few different ways and also redundantly)
2) Drop the link instead of replying with an error if we got a malformed HelloRequest (making this consistent with handling other protocol errors).
3) Move HelloRequest processing into link_state.rs & simplify by removing an intermediate state which wasn't actually waiting on anything.

Maybe easiest to review each commit separately.


mkolehmainen (approved):
Looks good to me!

PR #1196, opened 2026-01-30
Mk/1194 blobs

### LINKED ISSUE #1194: visa_mgmt::build_connect_request is buggy
This function is dropping blob information that must be passed on to the visa service.

First of all it is assuming the blob (which is a base64 encoding of a json encoded struct) is always SS type, which is wrong.  

Also it is then just creating an SS type blob and only setting the challenge field.   

WHAT SHOULD HAPPEN-
- The correct blob should be communicated in the connect request.  So maybe caller to this function should pass the real blob.
- We need all the data at the VS.  
- The prototype VS decodes the json -- the new VS wants a strongly typed thing.

mkolehmainen:
Fix a bug in the handling of connect-request payloads.

This required some clean up in zpr-common first.  Now propagated into zpr-core.

Becuase it was incredibly confusing in libnode, I renamed the "vsapi" import of the thrift to "vsapi_thrift".

### DIFF
--- a/adapter/ph/src/visa_mgmt.rs
+++ b/adapter/ph/src/visa_mgmt.rs
@@ -1,4 +1,5 @@
 use crate::assembly::Assembly;
+use crate::auth::AuthBlob;
 use crate::counters::ManagementCounterType;
 use crate::link_state::{LinkEvent, LinkStateError};
 use crate::logging::targets::VISA_MGMT;
@@ -9,6 +10,7 @@ use std::num::NonZero;
 use std::sync::Arc;
 use std::time::UNIX_EPOCH;
 use tracing::*;
+
 use zpr::dn::VISA_SERVICE_CN;
 use zpr::packet_info::{LinkId, VisaId};
 use zpr::vsapi_types::{self, AuthServicesList};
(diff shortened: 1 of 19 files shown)


adapter/ph/src/visa_mgmt.rs
```diff
@@ -65,18 +67,28 @@ pub fn build_connect_request(
...
 
-    // The visa service expects to find the BLOBs in the challenge response buffers.
-    let mut ss = vsapi_types::ZprSelfSignedBlob::default();
-    ss.challenge = blob.as_bytes().to_vec();
-    let crbufs: Vec<vsapi_types::AuthBlob> = vec![vsapi_types::AuthBlob::SS(ss)];
+    // In v2 the blobs are transported to the VS in a typed structure.
+    // In v1 (prototype) the blobs are b64 encoded json.
+    let mut crbufs: Vec<Vec<u8>> = Vec::new();
+    match blob {
+        AuthBlob::SelfSigned(ss_blob) => {
```
owenrhysthomas, line 82: `zpr-common` doesn't differentiate between the two types of blob in the V1 Connect Request, does it need to?

owenrhysthomas (approved):

adapter/ph/src/visa_mgmt.rs
mkolehmainen: No. Over the wire they both look like strings.

PR #1197, opened 2026-01-30
Ort/tls for capnp

### LINKED ISSUE #200: implement TLS on the Cap'n Proto connection node<->visa_service
The visa service channel setup between the visa service and the node is using thrift on TCP but without TLS.  Need to add TLS.

In this scenario, if it matters, the visa service is the "server" side of the connection.  The visa service should use it's own cert as set in config file.  The visa service should confirm that the cert from the client side is signed by the authority.  So the visa service will need new config option for the root-ca cert.

The node, in turn, should confirm that the cert from the visa service is similarly signed.

owenrhysthomas:


libnode2/src/vsconn.rs
```diff
@@ -94,8 +99,11 @@ impl VSConn {
 
         let sock = tokio::net::TcpStream::connect(self.vs_addr).await?;
         sock.set_nodelay(true)?;
+        println!("VS ADDR {}", self.vs_addr.ip());
```
mkolehmainen: stray println!

mkolehmainen (changes requested):
Looks good - I just noted a stray println that ought to be removed.

mkolehmainen (approved):
Great!

PR #1205, opened 2026-02-05
Ort/update ph deps

owenrhysthomas:
Initially upgraded just bytes to mitigate [this security error.](
https://github.com/org-zpr/zpr-core/security/dependabot/55) I don't even think we use BytesMut::reserve, so this is probably not necessary, but can't hurt to update anyway.

Figured I would update other deps while I was at it. 

Did not upgrade the following dependencies: socket2, tokio-tun, and nix. Each of these would require a rework of the code that, while not majorly intensive, would require more time than this PR was meant to do. 

Also [this security error](https://github.com/org-zpr/zpr-core/security/dependabot/60)

### DIFF
--- a/adapter/ph/src/km_noise.rs
+++ b/adapter/ph/src/km_noise.rs
@@ -416,11 +416,11 @@ impl KeyManagerStateMachine for KmNoise {
         self.send_hmac_key = None;
         if self.initiate {
             let rpk = self.peer_pub_key.as_ref().unwrap();
-            // TODO do we want ? here, or nested match statements with custom error message?
-            // Or do we want to replace the match completely and but a ? on build_initiator()
             let mut initiator = match snow::Builder::new(np)
-                .local_private_key(self.local_keypair.private.as_ref())?
-                .remote_public_key(rpk)?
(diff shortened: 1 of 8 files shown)


adapter/ph/src/km_noise.rs
```diff
@@ -416,9 +416,11 @@ impl KeyManagerStateMachine for KmNoise {
         self.send_hmac_key = None;
         if self.initiate {
             let rpk = self.peer_pub_key.as_ref().unwrap();
+            // TODO do we want ? here, or nested match statements with custom error message?
+            // Or do we want to replace the match completely and but a ? on build_initiator()
```
cpacejo: Since the only errors raised by these methods is if we have a coding error (specifically, if we've called the methods multiple times), I think the right thing here to do is `.unwrap()` the errors.  (There's nothing the caller can do to recover from our coding error.)

(Same for the later call to Builder as well.)

cpacejo (changes requested):

cpacejo (approved):
Looks good!

PR #1214, opened 2026-02-18
Mk/1202 port to ln2

### LINKED ISSUE #499: Make Vsconn::run nonblocking
Though async, this method blocks on Thrift calls, making it incompatible with the PH's mgmt thread (it must run on its own thread).

Now that we have async Cap'n Proto, we can trivially make this method nonblocking.

### LINKED ISSUE #1202: Switch PH to libnode2
Hook the libnode2 up to the ph, getting ride of old libnode.

Probably also requires redoing (or disabling and creating a ticket to redo...)  the core integration test thing that talks to a visa service, and just doing this may expose some bugs in vsv2.  Note that vsv2 requires val-key.

Follow on tasks:
- Prepare zpr-visaservice to deprecate "core" and "vs-confirm" -> https://github.com/org-zpr/zpr-visaservice/issues/121
- Remove old (thrift) output format on the ZPL compiler. -> https://github.com/org-zpr/zpr-compiler/issues/91

mkolehmainen:
Switch PH to use libnode2. Update integration tests to use vs (new visa service).  Also added some lifecycle event emitting for libnode2.

**Does not** remove the old libnode directory, nor clean out old vs and thrift.  Will add another task for those.

Still needs: 
- Probably tests will fail since new tests require  vs (not vservice) -> https://github.com/org-zpr/zpr-visaservice/issues/139
- Need valkey in the test environment -> https://github.com/org-zpr/zpr-dev-tools/pull/38 -- Actually we **don't** need this since the integration tests here run directly out of the yml files.

### Core: libnode2 Integration (primary theme)

  - adapter/ph ported from libnode → libnode2: Updated Cargo.toml/Cargo.lock to depend on libnode2. All relevant imports and call sites in main.rs, visa_mgmt.rs,
  mgmt/dock.rs, five_tuple_lookup_table.rs, and vss_worker.rs have been updated accordingly.
  - libnode2/src/vsconn.rs — Expanded significantly (+78 lines) with lifecycle improvements (lifecycle management, etc.).
  - libnode2/src/claims.rs — New file added.
  - libnode2/src/lib.rs — Exports claims module.

### Refactor: vs_worker split out

  - adapter/ph/src/vs_worker.rs — The VS worker logic was extracted from main.rs into its own module and substantially expanded (~76 lines of new logic vs. the 21-line
  stub it replaced).

### Fix: SS-blob signature

  - libnode2 / adapter — SS-blob signature is now treated as raw bytes rather than base64 text.

### Fix: Attribute keys

  - Added two missing attribute keys needed by the node.

### Integration Tests reworked

  - integration-test/ scripts (capture-test.sh, one-node-test.sh, common_funcs.sh, parse_arguments.sh) significantly updated.
  - integration-test/pregen/:
    - Added RSA key pair (node-rsa-key.pem, node-rsa-pubkey.pem).
    - IPv6 pregenerated test binaries replaced (.bin → .bin2 with updated content).
    - IPv4 pregenerated test binaries updated (larger, reflecting new format).
    - .zplc config files updated.
    - Removed README.md.

### Minor

  - .gitignore updated.
  - adapter/ph/src/config.rs — 6 lines added.


integration-test/capture-test.sh
```diff
@@ -66,6 +107,10 @@ function close_program() {
   "$PH_DEBUG_BIN" -p "$SOCKET" capture close-file
 }
 
+function check_vs_valkey_port() {
+  sudo ip netns exec zpr-vs bash -lc 'exec 3<>/dev/tcp/127.0.0.1/6379'
+}
+
```
cpacejo: Would be at home in `common_funcs.sh` I think

libnode2/src/vsconn.rs
```diff
@@ -95,20 +110,62 @@ pub struct VSConnHandle {
...
+    ///
+    /// `buffer_size` determines how many commands can be buffered to send to the run loop before
+    /// [VSConnHandle] starts blocking.
     pub fn new(
         buffer_size: usize,
         vs_addr: SocketAddr,
         node_cn: String,
         node_private_key: PKey<Private>,
-    ) -> Self {
+    ) -> (Self, mpsc::Receiver<VSConnLifecycleEvent>) {
```
cpacejo: I wonder if a [broadcast](https://docs.rs/tokio/latest/tokio/sync/broadcast/index.html) channel would be a better fit for the lifecycle events (allows multiple independent listeners).  Also then, rather than returning the receiver from `new()`, you can just add a new method maybe call it `subscribe_lifecycle_events(&self)` which just calls `.subscribe()` on the sender.  (You can just drop the initial receiver.)

libnode2/src/vsconn.rs
```diff
@@ -95,20 +110,62 @@ pub struct VSConnHandle {
...
+            Err(e) => {
+                error!(target: VS_RPC, "failed to send lifecycle event: {:?}", e);
+            }
+        }
+    }
+
+    /// Just like [VSConn::run] but will attempt to reconnect if the connection to the VS is lost, after
+    /// pausing for `reconnect_after`.
+    /// Sending a [VS2Command::Stop] will break the reconnect loop and cause this to return.
+    pub async fn run_with_reconnect(
```
cpacejo, line 156: Yay this closes #499 also!

cpacejo (commented):
Looks good!  Excited for the big switchover!

integration-test/capture-test.sh
mkolehmainen: ah yes

libnode2/src/vsconn.rs
mkolehmainen: Ah that's a good idea. Did that. And reworked the vs_worker a little to help to pick up on case where runloop somehow exits before I even get to send a connect-request.

cpacejo (approved):
Nice!!!

PR #1233, opened 2026-02-23
Mk/ln2test tui - adds a TUI to lntest

mkolehmainen:
First this is a refactor of lntest which was getting very long, so I put all the support code into a set of modules under /cli.

Second, this adds a very basic TUI to lntest so that the logging output is separated from the command output.

Screenshot_2026-02-23_11-29-11


owenrhysthomas (approved):
LGTM

PR #1234, opened 2026-02-23
Ort/unbind

### LINKED ISSUE #956: Implement unbind
Allow adapters to unbind tethers when they're done with them (e.g. on TCP FIN)

owenrhysthomas:
First pass at implementing unbind. I am not positive it is correct. I am also not quite sure how to test it.


adapter/ph/src/mgmt/handlers.rs
```diff
@@ -767,3 +767,73 @@ pub async fn handle_bind_egress_stream_response(
...
+
+    Ok(())
+}
+
+pub async fn handle_unbind_egress_stream_request(
+    asm: &Arc<Assembly>,
+    pkt: Packet,
+) -> HandleMgmtResult {
+    if !matches!(asm.ph_mode, PhMode::Adapter) {
+        error!(target: ZDP, "Link {}: received BindEgressStream message on node", pkt.metadata().ingress_link_id);
```
cpacejo: Should be "UnbindEgressStream" in the log message

adapter/ph/src/mgmt/requests.rs
```diff
@@ -372,6 +372,25 @@ pub fn send_bind_egress_stream_request<'a>(
...
+pub fn send_unbind_egress_stream_request<'a>(
+    asm: &'a Assembly,
+    link_id: LinkId,
+    txn_id: TxnId,
+) -> Sent<'a> {
+    debug!(target: ZDP, "Link {link_id}: sending UnbindEgressStreamRequest");
+
+    let req = core::new_heap_packet();
+
+    core::send_per_flow_txn_mgmt(
```
cpacejo: Unlike bind, for unbind we will want a `StreamId` (to indicate which stream to unbind), but not a `TxnId` (which is used to link requests to responses).  `send_per_flow_mgmt()` will form a message without the `TxnId` (and has an argument for stream ID).

adapter/ph/src/mgmt_processor_worker.rs
cpacejo, line 1: I think for Unbind we will actually want to simply make these Indications, not Request/Response pairs.  (Just chatted w/ Sarah & I'll have Frank update the RFC.)  So no transaction ID is needed (which is used to link Requests with Responses).

adapter/ph/src/zdp.rs
```diff
@@ -219,6 +219,18 @@ pub struct ZdpBindActorAddressResponseHeader {
     // followed by traffic classifier
 }
 
+#[derive(FromBytes, IntoBytes, Immutable, KnownLayout, Unaligned)]
+#[repr(packed)]
+// Alternativelt to the current setup, we could pass the FT instead, not sure
+// which is better for how the system is going to be set up
```
cpacejo: So for Unbind (both / all 3 of them), we're always unbinding a specific stream ID.  So these all will have empty bodies!  (No need to make an empty struct, you can just not declare one.)

adapter/ph/src/mgmt/handlers.rs
```diff
@@ -767,3 +767,73 @@ pub async fn handle_bind_egress_stream_response(
         _ => Err(HandleMgmtError::BadStructure),
     }
 }
+
+pub async fn handle_unbind_actor_address_request(
```
cpacejo: See comments below – unbind will be by stream ID, so no need for packet body or 5-tuple.

adapter/ph/src/mgmt/dock.rs
```diff
@@ -715,3 +716,76 @@ fn resolve_next_hop_bind_originator(
...
+    };
+
+    // Issue unbind request to next hop
+    if egress_link_id.get() == LOCAL_ACTOR_LINK_ID {
+        adapter::unbind_egress_stream(asm, NonZero::new(DOCK_LINK_ID).unwrap(), stream_id);
+    } else {
+        requests::send_unbind_egress_stream_request(
+            asm,
+            egress_link_id.get(),
+            // Not sure if the txn id should be different. The other place in the bind request where
```
cpacejo: No transaction ID needed!  (See my other comments)

Otherwise this logic looks good!

adapter/ph/src/mgmt/dock.rs
```diff
@@ -715,3 +716,76 @@ fn resolve_next_hop_bind_originator(
...
+            return;
+        }
+        _ => {
+            warn!(target: FLOW_MGMT, "Link {ingress_link_id}: unbind request: invalid initial packet");
+            return;
+        }
+    }
+
+    // Remove requested entry from the ELT
+    match asm.elt.remove(&five_tuple) {
```
cpacejo: The ELT lives on the ingress adapter, whereas at this point we're on the node.  So instead, we'll want to remove the stream from the PFT, which lives in the peer state for the ingress link (which you can get with `asm.peer_table.get(ingress_link_id.get())`).

I think the visa we'll actually want to leave in place, and let it age out naturally, since in general the lifetimes of visas are decoupled from streams.

However, though we'll leave the visa in place, we _will_ need to additionally un-link the stream ID from the corresponding visa.  This will need corresponding `unlink_forwarding_entry()` functions added in the `visa_table` module.

(The entry being created in the PFT, and being linked to the visa, both occur in `mgmt::dock::requested_tether_granted()`.)

cpacejo (changes requested):
This is a good first pass!  I think the main changes to make are –

* No transaction IDs needed, the unbind requests should be Indications without Responses.
* Unbind from adapter->node should just be by stream ID, no packet body or 5t needed.
* Both types (actually all 3, once we have node->node) of unbind request I think can all use the same message, `UnbindEgressStreamIndication`.  When handling it, choose the underlying logic to use based on whether it's coming from a node or adapter, and whether we are a node or adapter.
* For handling the unbind request on a node, we want to remove the stream from the PFT of the ingress adapter, and unlink the stream from the corresponding visa, but don't remove the visa itself.

& lemme think the best way to test this… could be to add a command to the admin tool.

Otherwise the logic looks good!  There's a lot of moving parts here for sure.

owenrhysthomas:
One question I had was whether the `Visa::unlink_forwarding_entry` should produce an error if there is no match found, or if it should stay as is and do nothing if there is no match

cpacejo:
> One question I had was whether the `Visa::unlink_forwarding_entry` should produce an error if there is no match found, or if it should stay as is and do nothing if there is no match

I think this is fine to produce no error.  It's definitely a logic error, but not one we can handle, and not worth panicking over.

adapter/ph/src/mgmt/dock.rs
```diff
@@ -715,3 +715,51 @@ fn resolve_next_hop_bind_originator(
...
+    match asm
+        .visa_table
+        .write()
+        .unwrap()
+        .unlink_forwarding_entry(pft_pep.visa_id, &pft_pep.next_hop)
+    {
+        Ok(()) => (),
+        Err(e) => {
+            warn!(target: FLOW_MGMT, "Link {ingress_link_id}: error removing stream from visa: {}", e);
+            return;
```
cpacejo: I think logging is the correct thing to do, but we should continue on to issue the request to the next hop.

adapter/ph/src/mgmt/dock.rs
```diff
@@ -715,3 +715,51 @@ fn resolve_next_hop_bind_originator(
...
+        }
+    }
+
+    let Ok(egress_link_id) = visa_mgmt::get_egress_link_for_visa(asm, pft_pep.visa_id) else {
```
cpacejo: No need to use the visa for the lookup, `pft_pep.next_hop.0` tells us the next hop!

adapter/ph/src/mgmt/handlers.rs
```diff
@@ -767,3 +767,56 @@ pub async fn handle_bind_egress_stream_response(
...
+        }
+        LinkType::NodeToNode => {
+            debug!(
+                target: ZDP,
+                "Link {}: node -> node UNIMPLEMENTED",
+                ingress_link_id.get()
+            );
+            return Err(HandleMgmtError::MessageNotPermitted);
+        }
+        LinkType::Internal => {
```
cpacejo: `Internal` we do need to handle, it covers the links between the node and its local adapter.  There are two cases:

If the `ingress_link_id` is `zpr::packet_info::LOCAL_ACTOR_LINK_ID`, we should treat this the same as `NodeToAdapter`: that is, the local adapter is requesting that the dock unbind the ingressing stream.

If the `ingress_link_id` is `zpr::packet_info::DOCK_LINK_ID`, we should treat this the same as `AdapterToNode`: that is, the node/dock is requesting that the local adapter unbind the egressing stream.

(This system is confusing, because the node and the local adapter share the same peer table…!)

You can do this with the match something like:

```
match (link_type, ingress_link_id) {
  (LinkType::NodeToAdapter, _) | (LinkType::Internal, LOCAL_ACTOR_LINK_ID) => { remove from PFT }
  (LinkType::AdapterToNode, _) | (LinkType::Internal, DOCK_LINK_ID) => { remove from DLT }
  (LinkType::NodeToNode, _) => { TODO }
  (LinkType::Internal, _) => { invalid }
}
```

adapter/ph/src/visa_table.rs
```diff
@@ -133,6 +133,21 @@ impl Visa {
...
+        for (i, fe) in self.streams.iter().enumerate() {
+            if forwarding_entry == fe {
+                idx = Some(i);
+            }
+        }
+
+        if idx.is_some() {
+            self.streams.remove(idx.unwrap());
+        }
+        // Do we want to error in the else case here? Or leave as is
```
cpacejo: You could return a bool, true if we did find the entry, false if not.

& return that value as the `Ok` result from `VisaTable::unlink_forwarding_entry()`.

adapter/ph/src/visa_table.rs
```diff
@@ -133,6 +133,21 @@ impl Visa {
...
+        let mut idx: Option<usize> = None;
+
+        for (i, fe) in self.streams.iter().enumerate() {
+            if forwarding_entry == fe {
+                idx = Some(i);
+            }
+        }
+
+        if idx.is_some() {
+            self.streams.remove(idx.unwrap());
```
cpacejo: Slight optimization: you can use [swap_remove()](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.swap_remove) here.  (We don't care about the order in the Vec.)

adapter/ph/src/visa_table.rs
```diff
@@ -133,6 +133,21 @@ impl Visa {
...
+    pub fn unlink_forwarding_entry(&mut self, forwarding_entry: &ForwardingEntry) {
+        let mut idx: Option<usize> = None;
+
+        for (i, fe) in self.streams.iter().enumerate() {
+            if forwarding_entry == fe {
+                idx = Some(i);
+            }
+        }
+
+        if idx.is_some() {
```
cpacejo: Would be slightly cleaner as `if let Some(idx) = idx { ... }`… then no need for the unwrap.

adapter/ph/src/visa_table.rs
```diff
@@ -133,6 +133,21 @@ impl Visa {
...
     }
+
+    pub fn unlink_forwarding_entry(&mut self, forwarding_entry: &ForwardingEntry) {
+        let mut idx: Option<usize> = None;
+
+        for (i, fe) in self.streams.iter().enumerate() {
+            if forwarding_entry == fe {
+                idx = Some(i);
+            }
+        }
```
cpacejo: `let idx = self.streams.iter().position(|fe| forwarding_entry == fe);` will do the same!

cpacejo (changes requested):
Looks good!  Just some minor changes, and also support for the internal links is needed.

adapter/ph/src/mgmt/handlers.rs
```diff
@@ -767,3 +767,56 @@ pub async fn handle_bind_egress_stream_response(
...
+        // who sent this??
+        error!(target: FLOW_MGMT, "coding error: stray packet from unknown source; dropping");
+        return Ok(());
+    };
+
+    let link_type = match asm.peer_table.get(ingress_link_id.get()) {
+        Some(peer_state) => peer_state.link_state_machine.get_link_type(),
+        None => {
+            error!(target: FLOW_MGMT, "ingress link id not recognized");
+            return Ok(());
```
cpacejo: Here we'll just want to return `Err(HandleMgmtError::LinkClosed)` without logging (the caller will take care of that).

adapter/ph/src/mgmt/dock.rs
```diff
@@ -715,3 +715,49 @@ fn resolve_next_hop_bind_originator(
 
     Ok((ingress_link_id, ingress_txn_id))
 }
+
+pub fn unbind_stream(asm: &Arc<Assembly>, ingress_link_id: NonZero<LinkId>, stream_id: StreamId) {
+    let peer_table = match asm.peer_table.get(ingress_link_id.get()) {
+        Some(pt) => pt,
+        None => {
+            warn!(target: FLOW_MGMT, "Link {ingress_link_id}: unbind request: no peer table");
```
cpacejo: I would just return without logging here (and generally in cases where the link isn't found in the peer table).  Reason being, when a link is going down, all its peer state gets deleted, so whether or not we managed to change something in its peer state before it went down or not doesn't really matter.  So if we did log, then when a link goes down, we'd get random error messages relating to whatever operation happened to be in progress, which don't actually communicate any information to us (since the end result is the same as if the operation had completed before the link went down).

adapter/ph/src/mgmt/dock.rs
```diff
@@ -715,3 +715,49 @@ fn resolve_next_hop_bind_originator(
...
+            return;
+        }
+    };
+
+    // Get visa id of next hop
+    let pft_pep = match peer_table.pft.get(stream_id) {
+        Some(pep) => pep,
+        None => {
+            warn!(target: FLOW_MGMT, "Link {ingress_link_id}: unbind request: no pft pep");
```
cpacejo: Maybe good to include `{stream_id}` in the log message.

cpacejo (changes requested):
Yes this is great!  Just a couple trivial changes I'd suggest around logging.

cpacejo (approved):
Looks great thanks for making the changes!

PR #1237, opened 2026-02-25
Mk/workspace

mkolehmainen:
Top level Cargo.toml which lists the important shared deps.
Updated workspace Cargo files to reference the workspace.

**NOTE** Big usage change is that the rust binaries now all end up $ROOT/target/debug rather than into /target/debug dirs in the individual crates.

Since binary locations have changed this is now using a new version of our dev-tools workflow.


owenrhysthomas (dismissed):
🔥

owenrhysthomas (dismissed):

owenrhysthomas (approved):
👍

PR #1242, opened 2026-02-27
Ort/wireshark coloring

### LINKED ISSUE #1070: coloring for wireshark

owenrhysthomas:
NOTE right now only the echo packets are paired with their ack. I am not sure totally of which packets have the ack. I have to do some more digging into the code/rfc to get the whole picture with the ack, and then I will update.

Checkmark for ack, along with link
image

Arrows in req/res, along with link
image


owenrhysthomas:
image
 Ack and linked request for bind actor response!

cpacejo (approved):
Nice this is awesome!!

PR #1244, opened 2026-03-02
Fix integration test

cpacejo:
Fix various integ test issues:

1. The error logic for the ping tests was wrong in a couple places, causing failures to be lost.
2. The visa revocation test is no longer valid: revoking a visa just causes a new one to be re-requested (which succeeds).
3. The valkey lookup was causing issues on Debian (where the symlink resolved to a binary with a different name, causing VK to do something different).


mkolehmainen (approved):
Great!

PR #1246, opened 2026-03-02
added abiliy to get node addr from adapter

### LINKED ISSUE #1245: get node info from adapter using CLI

owenrhysthomas:
Chris also noted the address you might be wanting is the `acror_addresses` in the `LinkStateMachine`, but the comments about that say those are for when you are on a node, not an adapter.. Let me know if the info you want is not the info I am giving here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2117,6 +2117,7 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "tokio-util",
+ "zpr",
  "zpr-ext",
 ]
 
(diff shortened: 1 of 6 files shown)


mkolehmainen (approved):
Does return the correct value! 

However I notice that when the link is shutdown calling "addr" causes the ph-cli to hang, and the ph itself goes into serious hang where I cannot even exit it (have to kill -9).  Don't know if this is an issue introduced here or if this is a new bug.  Let me know if it is new bug and I'll add an issue for it.

mkolehmainen:
Actually that hang is not your problem at all. It hangs anyway due to an auth issue.

PR #1250, opened 2026-03-03
fixed default path discrepancy, added comment

### LINKED ISSUE #1247: management socket default in ph-cli differs from ph.
On ubuntu, when I run ph, it defaults to setting it's management socket in a location but that location is not the default when I run ph-cli.

Output from ph:
```
2026-03-03T16:20:29.398791Z  INFO startup: control socket bound to "/root/.local/share/zpr/control.sock"
2026-03-03T16:20:29.398877Z  INFO startup: capture socket bound to "/root/.local/share/zpr/capture.sock"
```

Output from ph-cli:
```
Options:
  -p, --socket <SOCKET>          Path to the Packet Handler's management socket [default: /var/run/zpr/ph.sock]
```

These should be the same.

owenrhysthomas:
Changed socket to PathBuf so we could use the exact same function logic as in the config. Changed cap-socket to align with socket.

### DIFF
--- a/adapter/cli/src/main.rs
+++ b/adapter/cli/src/main.rs
@@ -18,11 +18,11 @@ use std::io::prelude::*;
 use std::io::{BufReader, Error, IoSlice};
 use std::os::fd::AsFd;
 use std::os::unix::net::UnixStream;
+use std::path::PathBuf;
 use thiserror::Error;
 use tokio::time::{Duration, sleep};
 use tokio_util::compat::*;
 use zpr_ext::std::os::unix::net::{SocketAncillary, UnixStreamExt};
-use std::path::PathBuf;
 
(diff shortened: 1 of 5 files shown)


adapter/cli/src/main.rs
```diff
@@ -66,18 +67,19 @@ impl From<std::str::Utf8Error> for CliError {
 async fn main() -> Result<(), CliError> {
     let args = CmdlineArgs::parse();
     let socket = args.socket.clone();
+    println!("{:?}", socket);
```
mkolehmainen: Is this printf left in by accident?

mkolehmainen (dismissed):
Looks good. Remove the println?

mkolehmainen (approved):
LGTM!

PR #1255, opened 2026-03-05
Address CWE-208, CWE-295, CWE-347

cpacejo:
Found by Dependabot.  Also, remove the subdirectory Cargo.lock files which are no longer used since switching to using workspaces.


owenrhysthomas (approved):
👍

PR #1257, opened 2026-03-05
upgrade ratatui to upgrade lru to address CWE-476

cpacejo:
See <https://github.com/org-zpr/zpr-core/security/dependabot/62>.  Needed to upgrade ratatui to get the latest version of lru to address this issue.


mkolehmainen (approved):
Excellent!

PR #1263, opened 2026-03-13
Teach traffic classifier how to classify traffic

cpacejo:
Just some code rearranging to move traffic classification logic into the traffic classifier…


mkolehmainen (approved):
LGTM!

PR #1267, opened 2026-03-18
Mk/156 core support

### LINKED ISSUE #146: visa service startup w/ respect to state

mkolehmainen:
This is zpr-core work (mostly libnode2) to better handle restarts of adapter, node and visa service.  Has a corresponding PR in zpr-visaservice - https://github.com/org-zpr/zpr-visaservice/pull/170.  Tests probably won't work until VS changes merged in and new release bundle created.

 ---
  1. Optional CA Certificate (config.rs, km_cert_exchange.rs, main.rs, main_argparse.rs)

  - KmCertExchange updated to accept Option<X509> as the authority cert.
  - Adapters still require a CA cert (enforced at startup with a clear error); nodes do not.
  - Certificate verification in link_state.rs is now gated on whether a CA file is configured (require_ca_signature). Special peer name assignment (e.g. identifying the visa service) works even without a CA cert.

  ---
  2. Visa Revocation on Peer Drop (assembly.rs, visa_table.rs, config.rs)

  - drop_peer() in Assembly now calls revoke_for_link() on the visa table when a peer is removed (node mode only).
  - Visas with IDs below MIN_VISA_ID = 1000 are exempt (reserved for hardcoded/static visas).
  - New disconnect_adapters() method on Assembly: resets all non-VS, non-internal peers (used when the VS reports out-of-sync).

  ---
  3. VS Connection State Management (vs_worker.rs, vsconn.rs, error.rs)

  - New StateFlag enum (NoState / HasState) tracks whether the node has established state with the visa service across reconnects.
  - vs_worker handles a new OutOfSync error from the VS: clears all adapter connections and visas, then resets state to NoState before retrying.
  - All Cap'n Proto RPC calls now wrapped with timeouts.

  ---
  4. VS Identity Check Fix (visa_mgmt.rs)

  - build_connect_request() now identifies the visa service peer via the special peer table (SpecialPeerName::VisaServiceAdapter) rather than by matching the CN string. This correctly respects whether the certificate was verified.

  ---
  5. README Rewrite (README.md)

  - Setup guide substantially rewritten to reflect current tooling: simplified key generation, optional CA cert, new VS config/startup steps, Valkey dependency noted, updated zplc invocation (no -k flag, outputs .bin2).


adapter/ph/src/assembly.rs
```diff
@@ -176,6 +176,28 @@ impl Assembly {
         }
     }
 
+    /// Disconnect all non-internal, non-vs adapters.
+    /// Used in a node context when (for example) we loose all state with the visa service.
+    ///
+    /// TODO: Does this remove visas related to the adapter/peer?
```
cpacejo: I don't think it does.

adapter/ph/src/link_state.rs
```diff
@@ -574,41 +574,52 @@ impl LinkStateWrapper {
...
-                                info!(target: LINK_STATE, "Link {link_id} assigned special name {name:?}")
-                            }
-                            Err(_) => {
-                                warn!(target: LINK_STATE, "Unable to assign link {link_id} special name {name:?}")
-                            }
+                    (cert, true)
+                }
+            };
+
+            if (require_ca_signature && is_verified) || !require_ca_signature {
```
cpacejo: Logic simplification: `is_verified || !require_ca_signature`

adapter/ph/src/visa_table.rs
```diff
@@ -284,6 +295,44 @@ impl VisaTable {
...
+            .filter_map(|(visa_id, visa)| {
+                if *visa_id >= config::MIN_VISA_ID as i32
+                    && visa.streams.iter().any(|entry| entry.0 == link_id)
+                {
+                    Some(*visa_id)
+                } else {
+                    None
+                }
+            })
+            .collect();
```
cpacejo, line 329: I wonder if this vector could potentially be very large.  Memory for the various tables we preallocate (ish... modulo OS-level lazy allocation), and I wonder if a single inter-node link could account for the majority of visas, thus resulting in us here temporarily allocating a vector of the same order of magnitude as the visa table.  Is there a way we can do this without collecting the vector first?

adapter/ph/src/visa_table.rs
```diff
@@ -284,6 +295,44 @@ impl VisaTable {
...
+    /// — a partially-cleaned visa could forward traffic over a stale route.
+    /// In the future, if visas legitimately span multiple links and partial
+    /// cleanup (retaining entries on surviving links) is needed, this method
+    /// should be split into a "remove entries for link" path followed by a
+    /// lookup table rebuild, leaving the visa alive when it still has valid
+    /// entries.
+    pub fn revoke_for_link(&mut self, link_id: LinkId, peer_table: &peer_table::PeerTable) {
+        let to_revoke: Vec<VisaId> = self
+            .table
+            .iter()
```
cpacejo, line 319: Also – this is potentially a slow operation which is holding an exclusive lock on the visa table and 5t lookup table, even for links with few visas (since we're traversing the whole table here).

I wonder if this would be better either (a) by maintaining a linked list of visas for each link, so we need only traverse those, or (b) performing this cleanup lazily or asynchronously somehow, so as not to block all other operations on it.

libnode2/src/vsconn.rs
```diff
@@ -193,141 +244,77 @@ impl VSConn {
...
+                tokio::pin!(ping_timeout);
+                let mut ping_failures = 0;
+
+                // Then loop over commands and periodically ping.
+                loop {
+                    tokio::select! {
+                        // If we get a command, handle it.
+                        cmd = self.cmd_rx.recv() => {
+                            match cmd {
+                                Some(cmd) => {
```
cpacejo: Could eliminate two layers of indentation using `let Some(cmd) = cmd else { ... };` here.

libnode2/src/vsconn.rs
```diff
@@ -546,6 +729,21 @@ impl VSConn {
             }
         }
     }
+
+    async fn do_ping(
+        &self,
+        vs_h: &vsapi2::v_s_handle::Client,
+        ping_failures: u32,
+    ) -> Result<(), VSApiError> {
+        let dur = PING_MIN_TIMEOUT + Duration::from_secs(ping_failures as u64);
```
cpacejo: The relationship between ping failure count and extra duration here seems non-obvious.  (I recognize above that the timeout for each retry is hardcoded as 1 second… I think it would be clearer if that's a defined constant which is then used here to compute the total timeout duration.)

cpacejo (dismissed):
Looks good!  A few small suggestions for improvement but nothing blocking.

libnode2/src/vsconn.rs
mkolehmainen: Ah, I forgot about (if I ever knew about...) `let Some(x) = x else {}`

mkolehmainen:
Agree that we need to look into how to keep revoke under control.  Created issue: 
https://github.com/org-zpr/zpr-core/issues/1281

cpacejo (approved):
Nice looks good thanks!

Issue #1268, opened 2026-03-18
node restarting causes connect fail from adapter.


mkolehmainen:
This is using new code from these PRs not yet in main:
* https://github.com/org-zpr/zpr-core/pull/1267
* https://github.com/org-zpr/zpr-visaservice/pull/170

### To reproduce:
1. start visa service
2. start node
3. start visa service adapter
4. wait for node and vs to sync.
5. kill the node
6. wait a few seconds for vs-adapter to settle down
7. start the node
8. the node and adapter are never able to synchronize

### Notes

What appears to happen is that the adapter never receives the HELLO-RESPONSE message from the node even though in my debugging I was able to see that the node was sending it (in code anyway, did not check the wire).  So anyway, the adapter times out in HELLOING and terminates the link.

mkolehmainen:
**LLM ANALYSIS:**

  Root cause — the adapter's PeerState (and its zdpr_recv) is reused across restarts via setup_restart. The node always starts fresh and sends HelloResponse at seq=0, but the adapter's receiver already has
  seq=0 recorded as seen from the previous session, so it classifies it as Duplicate → AckDoNotProcess. The packet is ACK'd (so the node stops retransmitting) but never processed.

  Secondary issue — zdpr_send has the same staleness problem in the other direction. It's currently masked because the node's fresh receiver accepts any first seq number up to offset+31, but once the
  adapter has restarted ~32 times the HelloRequest will also be silently dropped.

  Fix — add reset() methods to zdpr::Sender and zdpr::Receiver, and call both in process_keying_done (link_state.rs:555) just before transitioning to Helloing. That's the natural boundary between the crypto
   session and the application-level handshake.

mkolehmainen:
Initial attempt targeted wrong subsystem: https://github.com/org-zpr/zpr-core/pull/1270

But see chris notes on that PR. Sounds like the right idea is for the adapter to detect the new link on the node side and open it's own.

PR #1270, opened 2026-03-19
LLM guided fix for #1268

mkolehmainen:
Used LLM to diagnose and propose fix for #1268.  I have tested this on my latest code and it does indeed fix the problem.  LLM summary follows...

### Root cause

  When a link is restarted, the adapter reuses the existing PeerState — including its zdpr::Sender and zdpr::Receiver. After a successful first session, the receiver has recorded that seq=0 was received. When the node restarts, it sends a new HelloResponse at seq=0, but the stale receiver classifies it as a duplicate (AckDoNotProcess) and the packet is never delivered to handle_hello_response. Because the adapter does still ACK the packet, the node doesn't retransmit, and the
  adapter eventually times out.

  The sender has a symmetric issue: its last_sent counter keeps accumulating across restarts, which would eventually cause HelloRequests to be dropped by the node's fresh receiver once the counter exceeded its window (~31 restarts).

### Changes

  src/zdpr.rs

  - Added Sender::reset() — restores the sender to the same state as Sender::new(), discarding any queued or unacknowledged packets and waking their wakers.
  - Added Receiver::reset() — restores the receiver to the same state as Receiver::new(window_size), clearing the seen-sequence-number bitmap while preserving the configured window size.
  - Changed poll_ack and poll_send to return Poll::Ready (instead of panicking) when called with a seq_num/packet_id that is no longer tracked after a reset. This prevents a panic cascade where async futures awaiting the old session's packets re-poll after reset wakes them.
  - Added four unit tests covering the above: test_reset_quiesces, test_reset_discards_in_flight_packets, test_poll_ack_after_reset_returns_ready, test_poll_send_after_reset_returns_ready.

  src/link_state.rs

  - In process_keying_done, reset both zdpr_send and zdpr_recv just before transitioning to Helloing. This ensures every new authenticated session starts with fresh ZDPR sequence numbers that match the remote peer's fresh state.

### DIFF
--- a/adapter/ph/src/zdpr.rs
+++ b/adapter/ph/src/zdpr.rs
@@ -207,10 +207,10 @@ impl<Pkt> Sender<Pkt> {
     /// Implementation of `Future::poll` to wait for send of a given packet.
     pub fn poll_send(&mut self, cx: &mut Context<'_>, id: QueuedPacketId) -> Poll<Option<SeqNum>> {
         let offset = id.wrapping_sub(self.last_enqueued) as i64;
-        assert!(
-            offset <= 0,
-            "poll of packet which has not yet been enqueued"
-        );
+        if offset > 0 {
+            // The sender was reset; this packet is no longer tracked.  Treat as cancelled.
+            return Poll::Ready(None);
+        }
 
         if -offset >= self.blocked.len() as i64 {
             return Poll::Ready(self.lookup_seq_num(id));
@@ -231,7 +231,10 @@ impl<Pkt> Sender<Pkt> {
     /// of a given packet.
     pub fn poll_ack(&mut self, cx: &mut Context<'_>, sn: SeqNum) -> Poll<()> {
         let offset = sn.wrapping_sub(self.last_sent) as i64;
-        assert!(offset <= 0, "poll of packet which has not yet been sent");
+        if offset > 0 {
+            // The sender was reset; this packet is no longer tracked.  Treat as acked.
+            return Poll::Ready(());
+        }
 
         if -offset >= self.unacked.len() as i64 {
             return Poll::Ready(());
@@ -442,6 +445,12 @@ impl<Pkt> Sender<Pkt> {
                 pkt.as_mut()
                     .map(|(pkt, _waker)| (sn_base.wrapping_add(offset as u64), pkt))
             })
+    }
+
+    /// Reset to the same state as `Sender::new()`.
+    /// Any queued or unacknowledged packets are discarded.
+    pub fn reset(&mut self) {
+        *self = Self::new();
     }
 
     /// Destructs this object, returning all currently queued packets for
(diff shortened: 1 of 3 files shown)


adapter/ph/src/zdpr.rs
```diff
@@ -231,7 +231,10 @@ impl<Pkt> Sender<Pkt> {
     /// of a given packet.
     pub fn poll_ack(&mut self, cx: &mut Context<'_>, sn: SeqNum) -> Poll<()> {
         let offset = sn.wrapping_sub(self.last_sent) as i64;
-        assert!(offset <= 0, "poll of packet which has not yet been sent");
+        if offset > 0 {
+            // The sender was reset; this packet is no longer tracked.  Treat as acked.
+            return Poll::Ready(());
+        }
```
cpacejo, line 237: We shouldn't be indicating to the caller that the packet was acked if it was not in fact acked.

adapter/ph/src/zdpr.rs
```diff
@@ -207,10 +207,10 @@ impl<Pkt> Sender<Pkt> {
...
     pub fn poll_send(&mut self, cx: &mut Context<'_>, id: QueuedPacketId) -> Poll<Option<SeqNum>> {
         let offset = id.wrapping_sub(self.last_enqueued) as i64;
-        assert!(
-            offset <= 0,
-            "poll of packet which has not yet been enqueued"
-        );
+        if offset > 0 {
+            // The sender was reset; this packet is no longer tracked.  Treat as cancelled.
+            return Poll::Ready(None);
+        }
```
cpacejo, line 213: The comment is incorrect: this treats the packet as acked.  Additionally, the packet is not in fact acked, and we should not indicate such to the caller.

adapter/ph/src/zdpr.rs
cpacejo, line 1: I don't think the changes to ZDPR are warranted.  The poll functions are carefully constructed to communicate the exact state of each in-flight packet and to catch bugs due to misuse.

I am assuming these changes were made because there are active futures which re-poll the ZDPR state machines after the state is reset, resulting in panics?

I would propose instead that we maintain an epoch counter in the peer state, which is incremented on every link reconnect, and validated by the futures in `mgmt::core` before calling these ZDPR functions.

That leaves the question of how should these futures respond to the epoch changing.  Returning `MgmtSendError::LinkClosed` isn't correct, as this is used extensively as an indication that the link is in fact closed and its state no longer matters.  Returning success isn't correct, because the packets may in fact not have been reliably delivered.

See my toplevel comment for further discussion.

cpacejo:
I feel like this is the wrong solution to this issue.

Correct my analysis where I'm wrong but my understanding here is that when restarting a link, we're reusing the same peer ID and peer state.

I think, beside the state of ZDPR, this begs the overarching question of what happens to the state of every other adapter subsystem which is actively communicating over this link.

If we restart the link reusing the same ID without any indication given to any subsystem accessing the link, then we risk that those subsystems believe the peer to be in one state when it may in fact be in a different state, since we've now lost synchronization.

(Notably, for ZDPR, this materializes as us being unsure of whether specific packets were reliably delivered or not.  This undermines the core guarantee of ZDPR: that every request is delivered exactly once to the peer, _or_ the peer terminates and we learn about that.)

What's driving us to restart a link with the same link ID, vs. destroying the old link and starting a new one with a new ID?

mkolehmainen:
I agree with what you are saying. Claude has failed to understand the larger picture here.  I think using new link is probably the right thing to do.  We just need a way to signal up that it has happened, I guess it will be the KM subsystem that can detect this.  

The top most level of this is just me restarting the node.  Obviously the node starts with zero state -- it's the adapter side that tries to re-establish link using same state. 

Please ignore this "fix" and we can work on approaching it from next layer up.

PR #1273, opened 2026-03-23
update crossterm and tui-input to match ratatui dependencies & eliminate old lru dependency

cpacejo:
Addresses <https://github.com/org-zpr/zpr-core/security/dependabot/62>.


mkolehmainen (approved):
Excellent

PR #1294, opened 2026-05-20
workspace packages

owenrhysthomas:
Moved the version, edition, and license packages to the workspace toml

I picked v0.6.0 because that was the version of the ph. The most recent release is v0.3.1, so we also could revert to v0.4.0.


cpacejo (approved):
Nice thanks for doing this!

svjaffer (approved):

PR #1296, opened 2026-05-20
attempt to prevent hangs during node shutdown.

### LINKED ISSUE #1293: Node refusing to terminate gracefully
See this run: https://github.com/org-zpr/zpr-core/actions/runs/26168945898/job/76998840693

Possibly because an adapter tries to reconnect during node shutdown?

mkolehmainen:
Identified an issue where the vsconn was retrying connects and failing to respond to shutdown requests while doing that. Now we use select in vsconn to be sensitve to other things going on.

~~Also as defensive measure throw a timeout on the reset call in assembly as the system is going down.~~

### DIFF
--- a/libnode2/src/vsconn.rs
+++ b/libnode2/src/vsconn.rs
@@ -108,6 +108,15 @@ pub struct VSConn {
     connect_fn: ConnectFn,
 }
 
+/// Handle for sending commands to a running [VSConn].
+///
+/// All async methods return [`VSApiError::ConnClosed`] in two situations:
+/// - The run loop has exited (the underlying command channel is closed).
+/// - A command was sent while the run loop was busy establishing a TCP
+///   connection or waiting out a reconnect delay; in those windows the
+///   command is discarded immediately rather than queued. Callers should
+///   treat `ConnClosed` as a transient error and retry once a new
+///   [`VSConnLifecycleEvent::RunLoopStarts`] event is observed.
 #[derive(Clone)]
 pub struct VSConnHandle {
     cmd_tx: mpsc::Sender<VS2Command>,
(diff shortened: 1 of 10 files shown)


libnode2/src/vsconn.rs
```diff
@@ -185,7 +203,16 @@ impl VSConn {
                 Err(e) => {
                     error!(target: VS_RPC, "VSConn: run loop exited with error: {:?}", e);
                     info!(target: VS_RPC, "VSConn: reconnecting in {} seconds...", reconnect_after.as_secs());
-                    tokio::time::sleep(reconnect_after).await;
+                    tokio::select! {
+                        _ = tokio::time::sleep(reconnect_after) => {}
```
cpacejo: This timer will be restarted any time a non-stop command is received.  Fix is to `let mut timeout = tokio::time::sleep(reconnect_after);` before the loop and just have `_ = &mut timeout` here.

cpacejo (changes requested):
One minor issue otherwise looks good!

libnode2/src/vsconn.rs
mkolehmainen: If we get a non-stop command in the reconnect delay the select completes and it tries to connect again.  I guess that behavior is a little unexpected, but maybe good?

cpacejo: Oh, I missed that the select isn't immediately surrounded by a loop.  Humm, if there's e.g. an admin tool periodically polling a hypothetical "GetStats" command, this would result in the reconnect delay never being observed in full.  So I'd lean toward wrapping the select in a loop, with the timeout initialized outside the loop, so it's always honored in full.

mkolehmainen: Yeah and it makes sense that if you call it asking for a delay we should give it to you.

cpacejo (approved):
Looks good!

PR #1298, opened 2026-05-21
ZDPR cancel

### LINKED ISSUE #1277: ZDPR cancel
ZDPR suffers from the issue of getting blocked by an "undeliverable" packet (e.g., one which is too large for the MTU), which the sender might not actually care about, leading to the entire link going down for not a good reason.

Proposed solution is to add a notion of "cancellation": a sender may send a new message, Cancel, to its peer, instructing the peer to cancel a specified message.  The peer then responds with either Acknowledgement, which indicates that the message was nonetheless received and thus not cancelled, or a new message Cancelled, which indicates that the message was in fact not yet received.  In the latter case, the peer marks the sequence number as cancelled, thus ignoring any future receipt of the original packet, and responding idempotently to any duplicate Cancel requests.  Only upon receiving Cancelled or Acknowledged may the sender advance its window.

The ZDPR send API is then extended with two new abilities: the ability to immediately request that an unacked message is cancelled, which definitively indicates whether it was in fact cancelled, and the ability to indicate that ZDPR is allowed to cancel a specific message after a configured retry count.  In either case, the caller may choose not to wait for the cancellation response, if it does not in fact care whether the peer received and acted on the message.  (Note that, for any message which is part of a transaction, the sender must heed cancellation to adjust the transaction state appropriately.)

### LINKED ISSUE #1316: Arg order incorrect in calls to `make_tcp_visa`
The last two args to that function are both i64's are are (configuration, expiration_ms). But in `visa_table.rs` the `new_with_vs_visas` function passes in (expiration_ms, configuration). (around line 195).

The unit tests also get it wrong.

cpacejo:
Imbues ZDPR with the ability to deterministically cancel packets which have already been sent, either immediately, or after a specified number of retries.

The purpose of this ability is to account for the case where some packet is "manifestly" undeliverable, because of e.g. an MTU issue or deep-packet inspection.  Since ZDPR normally guarantees reliable delivery of every packet, were such a packet encountered, this would result in a stalled link and eventual link timeout/teardown, with a high chance of reoccurring when the link is reestablished and the operation which originated the packet performed again.  Furthermore, such behavior could be triggerable by an end-user, since some management packets (namely, bind request) encapsulate user packets.

By allowing certain packets (again, namely bind requests) to be marked for cancellation after a specified number of retries, we avoid this situation: instead, an error is returned and the specific operation which originated the packet is aborted.  By permitting control of this feature on a case-by-case basis, we retain the guarantee of reliable delivery for all sent packets which do _not_ request cancellation (which presumably are those which either are not expected to be manifestly undeliverable, and/or are required for operation of the link).

There are several changes herein:

zdp.rs:
* "Cancel" and "Canceled" ZDP packets are added.  These have no bodies, and indicate respectively that a given packet has been _requested_ to be canceled, or has been _acknowledged_ to be canceled.

zdpr.rs:
* Sender statistics were added to ZDPR, to permit users to detect when the remote side does something weird like send conflicting Ack/Canceled notifications.  (We don't currently do anything with these statistics.)
* Added a Display trait for ZDPR sender state to aid debugging.
* Two packet states were added to the ZDPR sender: CANCEL_REQUESTED (represented in the code as UnackedState::Canceled), and CANCEL_ACKED (UnackedState::CancelAcked).  The meaning of these states is described in detail the updated comment section in that file.
* Due to the addition of CANCEL_ACKED as an additional penultimate state for each packet beside ACKED, in order for the user to tell the two apart in a non-racy fashion, CANCEL_ACKED does not age into FORGOTTEN automatically like ACKED does, but must be explicitly aged by the user, so added a mechanism for that.
* The BLOCKED and UNACKED packet states were augmented with an optional retry limit.  After the specified number of retries, the packet is automatically transitioned to CANCEL_REQUESTED.
* The representation of packet states was changed from an Option to dedicated enums to better support the new states.
* Methods to request cancellation, set retry limit, query cancellation status, process cancellation requests and ACKs, and age retry counters, were all added to ZDPR.

zdpr_worker.rs:
* The ZDPR worker was modified to properly age retry counters, and send cancellation request retries.

dispatch.rs:
* Dispatch modified to correctly hand off cancellation requests and ACKs to ZDPR, and to respond to the peer as warranted.

mgmt/core.rs:
* Added methods to send cancel-request/cancel-ACK packets.
* Eliminated the distinction between `Acked` and `SentAndAcked` futures.  Previously, `Acked` meant that a packet had been sent, and so dropping it could not possibly dequeue the packet (like in `Sent`).  `SentAndAcked` provided no such guarantee.  In practice this distinction wasn't very useful, since these futures were typically awaited immediately at the instantiation site, so there is little doubt what they represent, and it complicated the changes necessary to support cancellation, so instead now there is only `Acked`, which has the semantics that `SentAndAcked` used to.
* Added the `AckedOrCanceled` future, to represent packets which have been canceled or retry-limited, which completes when a packet is acknowledged or cancel-acked.  (Like the reworked `Acked`, its type is not indicative that a packet has been sent, but canceling/retry-limiting a packet and then _not_ waiting for its final disposition is typically inadvisable.)
* Added methods to explicitly cancel a packet, or to limit retries.

adapter_manager_worker.rs:
* The issuance of bind requests was modified to enforce a retry limit.  If the limit is reached and the packet canceled, we treat this as a tether denial.

config.rs:
* Default retry limit defined.  It's 3, for no particular reason.

packet.rs:
* Unrelated change to make the macOS build happy (not sure why this is needed now).

### DIFF
--- a/adapter/ph/src/config.rs
+++ b/adapter/ph/src/config.rs
@@ -42,6 +42,10 @@ pub const DEFAULT_ZDPR_RECEIVE_WINDOW_SIZE: usize = 32;
 
 pub const DEFAULT_ZDPR_RETRY_TIMER: std::time::Duration = std::time::Duration::from_millis(600);
 
+/// Used for specific messages to limit the number of ZDPR retries before giving up.
+/// (Not a global default!)
+pub const DEFAULT_ZDPR_RETRY_LIMIT: u8 = 3;
+
 pub const DEFAULT_REQUEST_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(3);
 pub const DEFAULT_TERMINATE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(1);
 
--- a/adapter/ph/src/config.rs
+++ b/adapter/ph/src/config.rs
@@ -46,7 +46,7 @@ pub const DEFAULT_ZDPR_RETRY_TIMER: std::time::Duration = std::time::Duration::f
 /// (Not a global default!)
 pub const DEFAULT_ZDPR_RETRY_LIMIT: u8 = 3;
 
-pub const DEFAULT_REQUEST_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(3);
+pub const LINK_HELLO_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(3);
 pub const DEFAULT_TERMINATE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(1);
 
 pub const DEFAULT_VSS_PORT: u16 = 8183;
--- a/adapter/ph/src/mgmt/core.rs
+++ b/adapter/ph/src/mgmt/core.rs
@@ -1,5 +1,3 @@
-#![allow(dead_code)]
-
 //! Core management packet functions.
 //!
 //! These are low-level primitives; most code should use the higher-level
@@ -180,6 +178,7 @@ impl<'a> Sent<'a> {
     ///
     /// Once true, [try_cancel()] can no longer succeed, dropping will no
     /// longer cancel the packet, and this future will be ready immediately.
+    #[allow(dead_code)]
     pub fn is_sent(&self) -> bool {
         let PacketId::Queued(packet_id) = self.packet_id else {
             return true;
@@ -197,6 +196,7 @@ impl<'a> Sent<'a> {
     /// Explicitly try to cancel sending the packet, returning the packet if
     /// successful, or an error if unsuccessful (meaning the packet was
     /// already sent).
+    #[allow(dead_code)]
     pub fn try_cancel(mut self) -> Result<Packet, Acked<'a>> {
         match self.try_cancel_internal() {
             Ok(packet) => Ok(packet),
@@ -242,6 +242,7 @@ impl<'a> Sent<'a> {
     ///
     /// Returns a future which resolves to an indication of whether the
     /// packet was indeed canceled.
+    #[allow(dead_code)]
     pub fn request_cancel(mut self) -> AckedOrCanceled<'a> {
         self.request_cancel_internal();
         AckedOrCanceled(self)
@@ -258,8 +259,6 @@ impl<'a> Sent<'a> {
             return;
         };
 
-        let mut zdpr_send = peer_state.zdpr_send.lock().unwrap();
-
         if matches!(self.packet_id, PacketId::Queued(_)) {
             if self.try_cancel_internal().is_ok() {
                 // immediately canceled
@@ -274,10 +273,12 @@ impl<'a> Sent<'a> {
             unreachable!();
         };
 
+        let mut zdpr_send = peer_state.zdpr_send.lock().unwrap();
         let Some(_pkt) = zdpr_send.cancel_sent_packet(seq_num) else {
             // already acked or cancel-requested
             return;
         };
+        drop(zdpr_send);
 
         // note, we drop the user's packet here; we could return it to them
 
@@ -374,18 +375,22 @@ impl<'a> std::future::Future for Sent<'a> {
 pub struct Acked<'a>(Sent<'a>);
 
 impl<'a> Acked<'a> {
+    #[allow(dead_code)]
     pub fn enqueue(self) {
         self.0.enqueue()
     }
 
+    #[allow(dead_code)]
     pub fn is_sent(&self) -> bool {
         self.0.is_sent()
     }
 
+    #[allow(dead_code)]
     pub fn request_cancel(self) -> AckedOrCanceled<'a> {
         self.0.request_cancel()
     }
 
+    #[allow(dead_code)]
     pub fn limit_retries(self, retry_limit: u8) -> AckedOrCanceled<'a> {
         self.0.limit_retries(retry_limit)
     }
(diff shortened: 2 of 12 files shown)


adapter/ph/src/zdpr.rs
```diff
@@ -345,33 +765,111 @@ impl<Pkt> Sender<Pkt> {
...
+        let idx = self.lookup_ack_index(sn)?;
 
-        if offset > 0 || -offset >= self.unacked.len() as i64 {
-            return None;
+        let (pkt, waker) = replace_with_or_abort_and_return(&mut self.unacked[idx], |s| match s {
+            UnackedState::Unacked { packet, waker, .. } => {
+                (Some((Some(packet), waker)), UnackedState::Acked)
+            }
+
+            UnackedState::Canceled { waker, .. } => (Some((None, waker)), UnackedState::Acked),
```
mkolehmainen: Does `UnackedState::Canceled` actually mean `UnackedState::CancelRequested` ?

adapter/ph/src/mgmt/core.rs
```diff
@@ -147,51 +171,160 @@ impl<'a> Sent<'a> {
...
+            // already acked or canceled
+            return;
+        }
+
+        let Some(peer_state) = self.asm.peer_table.get(self.link_id) else {
+            // no more link
+            return;
+        };
+
+        let mut zdpr_send = peer_state.zdpr_send.lock().unwrap();
```
mkolehmainen: Is it a problem here that we grab the zdpr_send lock at line 261, and then we grab it again in try_cancel_internal()  (line 216)?

adapter/ph/src/zdpr.rs
```diff
@@ -323,13 +638,118 @@ impl<Pkt> Sender<Pkt> {
...
+        }
+
+        let idx = (self.blocked.len() as i64 + offset - 1) as usize;
+        self.blocked[idx].limit_retries(retry_limit);
+    }
+
+    /// Limit the number of times this packet will be retried.
+    ///
+    /// Same as [limit_retries_by_id()], but by sequence number.
+    pub fn limit_retries_by_seq_num(&mut self, sn: QueuedPacketId, retry_limit: u8) {
```
mkolehmainen: Should `sn` be of type `SeqNum` ?

adapter/ph/src/mgmt/core.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(dead_code)]
```
mkolehmainen: This feels extreme... Whole file scope?

adapter/ph/src/config.rs
```diff
@@ -42,6 +42,10 @@ pub const DEFAULT_ZDPR_RECEIVE_WINDOW_SIZE: usize = 32;
 
 pub const DEFAULT_ZDPR_RETRY_TIMER: std::time::Duration = std::time::Duration::from_millis(600);
 
+/// Used for specific messages to limit the number of ZDPR retries before giving up.
+/// (Not a global default!)
+pub const DEFAULT_ZDPR_RETRY_LIMIT: u8 = 3;
```
mkolehmainen, line 47: How is this related to `DEFAULT_REQUEST_TIMEOUT`? Should we note that changing this may eventually run into that timeout?

mkolehmainen (dismissed):
Looks good.  I like the use of enums for the state -- I would not have thought of that. Who knew you could do so much with an enum! 

 I think that the **lock thing** is the only real issue in here.

The LLM found this (see below) maybe issue that I don't fully comprehend but maybe it will be clear to you:

### `AckedOrCanceled::poll` always reports `Acked` for queued packets

  `mgmt/core.rs:497–499`

  When `limit_retries()` is called on a `Sent` future whose packet is still blocked (ZDPR window full), `AckedOrCanceled` holds `PacketId::Queued`. Its poll then delegates to `poll_send_and_ack`, which returns `()` for both `Acked` and `CancelAcked` terminal states and discards the distinction. The poll immediately returns `PacketStatus::Acked` regardless:

```rs
  PacketId::Queued(packet_id) => {
      ready!(zdpr_send.poll_send_and_ack(cx, packet_id));
      Poll::Ready(Ok(PacketStatus::Acked))   // always Acked — CancelAcked indistinguishable
  }
```

  The `PacketId::Sent` case correctly calls `is_cancel_acked` to distinguish, but `Queued` never gets that chance.

  Consequence in `adapter_manager_worker.rs`: if the ZDPR window is full when a bind request is issued, the bind packet gets queued. If it later exhausts its retry limit and is cancel-acked, the match arm `Ok(PacketStatus::Canceled)` is never reached — deny_tether is silently skipped, and the requestor stalls indefinitely.

  The fix is to resolve the `SeqNum` once `poll_send_and_ack` returns (it can be retrieved via `lookup_seq_num` before the send completes, or by restructuring `poll` to first `poll_send`, record the seq num, then `poll_ack`).

adapter/ph/src/zdpr.rs
cpacejo: Yeah it's poorly named :/  I might go through and do a renaming pass.

adapter/ph/src/mgmt/core.rs
cpacejo: Yeah that's fair!  Will scope it down a bit more

adapter/ph/src/mgmt/core.rs
cpacejo: Oh, good catch! this should move down.  Clearly I missed testing this case…

adapter/ph/src/zdpr.rs
cpacejo: Ah yest good catch!

adapter/ph/src/config.rs
cpacejo: Oh good find, hummm, `DEFAULT_REQUEST_TIMEOUT` is used only in two places, both in the link state machine.  One is the adapter waiting for the node to send a HELLO.  That I think we want to keep but should rename it.  The other is waiting for an ACK from the init auth request – that use we should replace with a limited retry count.

mkolehmainen (approved):
LGTM!

PR #1304, opened 2026-06-11
Tighten up ca_file and certificate usage

### LINKED ISSUE #1041: Is CA-CERT really useful for the adapter? If not, remove it.
Confirm the how an adapter uses the CA-KEY.  I *think* it is just used to verify the certificate sent over by the node during KM setup.  Since the adapter already has the nodes public key, who cares if it holds a cert?  Anyway, if that's the only use for it we should remove it from the config and logic.

A node should still get it ... as it needs to verify special peers like "vs.zpr".  Another question: is holding a signed noise key really enough authority to connect as the visa service?

mkolehmainen:
- Makes ca_file optional for adapters. Startup now warns when no CA file is configured instead of failing adapter startup.
- Keeps adapter certificate_file optional, but requires name when no certificate is provided so self-signed cert generation has a valid CN.

- Delays bootstrap/BAS auth object construction until after config-file values and CLI overrides are merged, fixing --name precedence for CN-dependent auth setup.

- Tightens special-peer assignment so visa-service peers are recognized only from CA-verified certificates; unverified certs claiming special peer names are logged and ignored.

- Updates README guidance for node/adapter certificate behavior, peer verification, and visa-service certificate requirements.

### DIFF
--- a/adapter/ph/src/config.rs
+++ b/adapter/ph/src/config.rs
@@ -109,8 +109,7 @@ impl ArgError for str {
 /// use [crate::main_argparse::argparse].
 #[derive(Debug, Clone)]
 pub struct Config {
-    /// Name of this node or adapter. If no signed certificate is available this is used as the CN.
-    /// This has no meaning for nodes.
+    /// Adapter CN fallback when no signed certificate is configured. Ignored by nodes.
     pub name: String,
 
     /// Path to the unix domain socket for the control interface.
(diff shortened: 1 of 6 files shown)


adapter/ph/src/config.rs
```diff
@@ -444,26 +478,11 @@ impl Config {
     ) -> Result<(), ArgsError> {
         if let Some(config) = config {
             if let Some(bas_key) = &config.bas_key {
-                let keyfile = if bas_key.is_relative() {
-                    base_dir.join(bas_key)
+                if bas_key.is_relative() {
+                    self.bas_key_path = Some(base_dir.join(bas_key));
```
cpacejo, line 477: I think this branch actually covers both the relative and absolute cases.  At least, the [docs](https://doc.rust-lang.org/nightly/std/path/struct.Path.html#method.join) say "If `path` [the argument] is absolute, it replaces the current path."

cpacejo (dismissed):
Looks good!  Just a minor suggestion to simplify a line.

adapter/ph/src/config.rs
mkolehmainen: Ah, this actually is points to some confusion about those routines from earlier in the dev process. I will clean all that up now.

cpacejo (approved):
Looks good thanks!

PR #1317, opened 2026-06-15
bump zpr-common dependency to v0.18.0; VisaId is now u64

cpacejo:
Now matches Cap'n Proto definition.

Also, fix arg order in calls to make_tcp_visa (per @mkolehmainen request).


owenrhysthomas (commented):
Nice

owenrhysthomas (approved):

PR #1318, opened 2026-06-17
Minor changes to README after setting up a ZPRnet

owenrhysthomas:
- Corrected command syntax for adding IP addresses in README.
- added `integration-test/` before one of the `zpr-pki` commands for consistency
- added `authority/` before `auth-ca.crt` in conf toml files, because all other commands assume `auth-ca.crt` is in a subdirectory called `authority`
- added name for `adapter-vs-conf.toml`

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,9 @@ matches against.  Generate and sign one:
 
 ### Create a configuration file for the visa service adapter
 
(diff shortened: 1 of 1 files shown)


mkolehmainen (approved):
Great!

PR #1324, opened 2026-06-24
Prelude module

### LINKED ISSUE #1113: PH prelude
Add a "prelude" module which can be used in any other module as `use crate::prelude::*` which pulls in some core concepts which are used "all over" the PH.  Notably the Assembly and the logging macros and constants, but likely some other things too.

ianchang1:
Added a prelude module to reduce imports. In second commit current file imports were reduced with prelude


adapter/ph/src/prelude.rs
cpacejo, line 1: Nice I like these!  I would add a couple more:

```
pub use bytes::{Buf as _, BufMut as _};
pub use zerocopy::FromBytes as _;
pub use zpr::packet_info::*;
```

(The `as _` allows us to pull in the methods of those traits everywhere without pulling in the trait name itself.  This is all we use in many files despite that we import the name.)

Maybe `zpr::packet_info::*` is contentious, since there's a lot of names there, but we do tend to use many of them in many files so I think that's OK.

libnode2/test-data/test.zplc
cpacejo, line 1: These look like they snuck into the changeset from some local testing?  Let's leave them out of this PR.

cpacejo (changes requested):
Nice this looks good!  Will save a bunch of boilerplate.  Just added a couple suggestions for additional modules to pull in.  (Also looks like one stray change from testing snuck in.)

cpacejo (approved):
Thanks for the changes!  Any code change with net negative 100 lines is a win in my book.

PR #1327, opened 2026-06-25
Created functions for formatting link ids and edited print statements…

### LINKED ISSUE #1149: Stop saying "link 2" in adapter log messages
Link 2 (i.e. the dock) is the only link.  We should just say "dock link" or "tether".

(The node I believe does not have this problem, as its dock link is virtual.)

I think the way to do this is: add a method to `Assembly`, call it `format_link_id(&self, link_id: LinkId, f: &mut Formatter<'_>) -> Result<(), Error>`, which formats link 0 as "unknown link", link 1 as "adapter link", link 2 as "dock link", and all other links simply as "link X" (where X is the link_id).

Then also add a method, `formatted_link_id(&self, link_id: LinkId) -> impl std::fmt::Display`, which simply calls `std::fmt::from_fn(move |f| asm.format_link_id(link_id, f))`.

Then, all places where we print a link ID, we can pass `asm.formatted_link_id(link_id)` instead of the link ID itself.

ianchang1:
I changed all print statements except these:

No Assembly/asm in scope. To convert I would need a header change
link_state.rs:264 (set_state)
link_state.rs:746 (process_assigned_aaa)
handlers.rs:106 (handle_report)
handlers.rs:119 (handle_discard)
visa_table.rs:335 (revoke_for_link)

### DIFF
--- a/adapter/ph/src/adapter_manager_worker.rs
+++ b/adapter/ph/src/adapter_manager_worker.rs
@@ -49,14 +49,14 @@ fn do_request_tether_id(asm: &Arc<Assembly>, pkt: Packet) {
     };
 
     if !peer_state.link_state_machine.is_ready() {
-        debug!(target: FLOW_MGMT, "Link {dock_link_id} is not ready to receive traffic yet");
+        debug!(target: FLOW_MGMT, "{} is not ready to receive traffic yet", asm.formatted_link_id(dock_link_id));
         mgmt::core::count_event(asm, ManagementCounterType::DroppedNoSA);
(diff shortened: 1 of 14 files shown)


cpacejo (approved):
Perfect thank you!  Yes those few that have no `Assembly` in scope are fine to leave as-is, they mostly only deal with allocated link IDs, not predefined link IDs, anyway.

(We could make `formatted_link_id()` a global, but it's possible/likely the meaning of the internal link IDs may become dependent on configuration.)

PR #1330, opened 2026-07-01
added various rustyline features to CLI to make interactive

### LINKED ISSUE #1328: Swap ph-cli over to RustyLine (or some other alternative)
Clap is built for command line arg parsing, not making an interactive CLI.  We should swap in a library intended for CLI development, like [RustyLine](https://docs.rs/crate/rustyline/18.0.1) or something similar.

ianchang1:
Worked on implementing rustyline in CLI for testing. Features I added

- autocomplete for current commands
- hinting for autocomplete depending on state of command
- pop up descriptions for commands once you finish typing them
- listing for all possible commands at a given point (hit tab when nothing is typed or when the prefix of multiple options has been typed)
- coloring for hints and valid commands show as green until they are fully typed/auto completed
- Validation to continue allowing typing if "" are not closed

I know the original intention for this task was to swap clap out for rustyline but it seemed like the features of rustyline are almost entirely visual so I built it on top of the existing clap structure. If there are other problems with the CLI with its current implementation I am happy to find a different library or continue building the current one out. Hope this helps

### DIFF
--- a/adapter/cli/src/rusty_helper.rs
+++ b/adapter/cli/src/rusty_helper.rs
@@ -19,6 +19,7 @@ const COMMANDS: &[&str] = &[
     "logging",
     "addr",
     "quit",
+    "help",
 ];
 
 const CAPTURE_SUBCOMMANDS: &[&str] = &[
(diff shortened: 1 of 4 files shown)


svjaffer (dismissed):
I forgot Owen had already done part of this migration!  Thanks for enabling these extra features!

svjaffer:
Just adding @owenrhysthomas since he touched this last

adapter/cli/src/rusty_helper.rs
```diff
@@ -0,0 +1,178 @@
...
+        "" => COMMANDS,
+        _ => &[],
+    }
+}
+
+// Helper function to get the parameter info for a given command
+fn get_command_parameter_info(command: &str) -> &str {
+    match command {
+        "echo" => " : Ping the Packet Handler",
+        "counters" => " -r/--reset : Display all counters + uptime; --reset to zero them instead.",
```
owenrhysthomas: this line shows the same command (--reset) for displaying all the counters and for resetting. I believe that you don't need any additional commands to just display the counters.

adapter/cli/src/rusty_helper.rs
```diff
@@ -0,0 +1,178 @@
...
+    match head_command {
+        "capture" => CAPTURE_SUBCOMMANDS,
+        "link" => LINK_SUBCOMMANDS,
+        "" => COMMANDS,
+        _ => &[],
+    }
+}
+
+// Helper function to get the parameter info for a given command
+fn get_command_parameter_info(command: &str) -> &str {
```
owenrhysthomas, line 153: Most of these seem correct to me, they are just not all formatted the same, some have capital letters at the beginning, some have lowercase. Some of them tell you what the command does (ex. logging, addr), others just give the arguments or the order of the arguments (ex. configure, start, sequence).

In `main_args.rs` there are doc comments (`///`) for each of the commands that describe what they all do. Just go through those and ensure that the descriptions here line up.

owenrhysthomas (changes requested):

owenrhysthomas (approved):
👍

PR #1338, opened 2026-07-07
migrating towards new capnproto admin interface

### LINKED ISSUE #1206: pass ancillary data through new capnproto admin interface
We use ancillary data to pass the capture socket to the PH.  We need this to finally move away from the old style text admin interface.

ianchang1:
Added all changes as outline in task. One note: Instead of deleting old implementation added "capnp-ancillary" gate to not interfere with original code base before the demo


cpacejo (approved):
Yes this looks great!

PR #1339, opened 2026-07-13
Two bugfixes

cpacejo:
First bugfix:

Mgmt ingest logic was wrong, was ACKing packets before seeing whether it had room to queue them, resulting in permanently lost mgmt packets.  Fix was easy.  Added test.  (Found via code inspection relating to ZDP hairpinning.)

Second bugfix:

A bounds check in Packet was wrong, leading to an unclear panic.  Fixed.  (Found while writing the above test.)

### DIFF
--- a/adapter/ph/src/mgmt/dispatch.rs
+++ b/adapter/ph/src/mgmt/dispatch.rs
@@ -24,7 +24,7 @@ pub fn dispatch_mgmt_packet_with_addr(
     pkt: &mut Packet,
 ) {
     match zdp::ZdpBaseHeader::ref_from_prefix(pkt.body()) {
-        Ok(base_hdr) if base_hdr.0.packet_type == zdp::ZdpPacketType::KeyManagement => {
+        Ok((base_hdr, _)) if base_hdr.packet_type == zdp::ZdpPacketType::KeyManagement => {
             pkt.advance(std::mem::size_of::<zdp::ZdpBaseHeader>());
 
             // TODO: once we have multi-node, how do we know whether this is a link or a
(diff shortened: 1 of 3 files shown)


adapter/ph/src/mgmt/dispatch.rs
```diff
@@ -103,24 +104,25 @@ pub fn dispatch_mgmt_packet_with_link(asm: &Arc<Assembly>, pkt: &mut Packet) {
...
-                }
-            }
-
             if disp.should_process() {
                 let mgmt_pkt = Packet::new_with_existing_metadata(pkt.buffer().clone());
                 match peer_state.mgmt_processor.try_enqueue_packet(mgmt_pkt) {
                     Ok(()) => (),
                     Err(queues::TryEnqueueError::Full(_mgmt_pkt)) => {
                         core::count_event(asm, ManagementCounterType::QueueBackpressure);
+                        // important: do not ACK this packet, we're dropping it!
```
mkolehmainen: We did call `receiver.process_packet` on this packet earlier, but we don't ACK now that the ack block is moved down. Is this ok? Will sender retry sending? And then won't receiver return AckDoNotProcess ?

mkolehmainen (dismissed):
Just one comment on moving the ACK block down. I don't fully understand the ZDPR so maybe it's fine so I am approving!

adapter/ph/src/mgmt/dispatch.rs
cpacejo: Oh good catch!!  Will need a slightly more intricate fix, will do & add a test for that too

cpacejo: OK, now we first query whether we _should_ process the packet, _then_ we try to enqueue it to mgmt, and then only if successful do we tell ZDPR that we've processed it.  Amended the test to retry the dropped packet from earlier in the test to ensure it now gets processed.

mkolehmainen (approved):
LGTM!

PR #1341, opened 2026-07-23
Opensslswitch

### LINKED ISSUE #1337: Remove dependency on openssl
Our dependency on openssl (libopenssl.so) can frustrate efforts to run on different distributions.  Let's look into whether we can remove this dependency (or gate it behind a feature flag) by taking advantage of rustls's capability to use different crypto providers.

ianchang1:
I completely removed the OpenSSL dependency in three parts.

1. Secure random number generation was switched to rand::OsRng. I know you initially suggested using rustls::crypto::CryptoProvider but OsRng made it so I didn't have to initialize a randomizer in each file while maintaining the same functionality (both pull from the same OS entropy).

2. RSA key loading and signing was moved to a helper file rsa_sign.rs using aws_lc_rs. Aws had the right functions I needed and an RsaKeyPair structure which kept private and public keys together. I updated all the RSA loading and signing locations across the repo to use it. Loading the X25519 noise keys is handled separately in pki.rs using pkcs8/pem, since those aren't RSA keys

3. Replaced openssl::X509 with my own Cert structure. The Cert holds DER bytes and is parsed on demand so it doesn't carry a lifetime and preserves the exact signed bytes. Reading the cert uses x509-parser, building self-signed certs uses x509-cert, and verifying signatures uses aws_lc_rs. Cert::verify handles both RSA-SHA256 (for the CA-signed certs) and Ed25519 (for self-signed ones). I updated all the callers.

Other Changes:

- Self-signed noise certs are now Ed25519-signed instead of RSA-SHA256. The x509-cert builder needs a signer that implements the RustCrypto signature traits. ed25519-dalek has that but the aws_lc_rs RsaKeyPair doesn't, and keeping RSA would mean pulling in the rsa crate which has a timing-attack advisory. I think this is fine since these certs are always Unverified and trust comes from matching the embedded key against the expected peer key, not from the signature.
- Certificate signature-verification failures are surfaced as PeerCertificate::Unverified rather than aborting the exchange.

I was very unfamiliar with these packages and coding anything cryptography related before this task so if something doesn't make sense it is definitely on my part. For choosing packages, my standards weren't very advanced, I looked for things that were secure and easy to integrate. Happy to change anything.


cpacejo (commented):
Were you able to look at the [rustls_pki_types](https://docs.rs/rustls-pki-types/latest/rustls_pki_types/) crate?  (Which is re-exported also as `rustls::pki_types`.)  We use rustls already, so if we can leverage it here that will allow us to avoid adding various new dependencies.

libnode2/src/vsconn.rs
```diff
@@ -126,7 +125,7 @@ pub struct VSConn {
     cmd_rx: mpsc::Receiver<VS2Command>,
     vs_addr: SocketAddr,
     node_cn: String,
-    node_private_key: PKey<Private>,
```
cpacejo, line 129: Wondering if we can do without the Arc here?  We oughtn't be cloning the private key anywhere but maybe we are.

adapter/ph/src/pki.rs
```diff
@@ -1,17 +1,38 @@
...
+        pkix::{KeyUsage, KeyUsages},
+    },
+    name::Name,
+    serial_number::SerialNumber,
+    spki::{AlgorithmIdentifierOwned, SubjectPublicKeyInfoOwned, SubjectPublicKeyInfoRef},
+    time::Validity,
+};
+
+use pkcs8::PrivateKeyInfoRef;
+use x509_parser::certificate::X509Certificate;
```
cpacejo, line 32: Looking at x509_cert and x509_parser, they both seem to have support for parsing PEM certificates, and inspecting the contents of certificates.  Would it be possible to use one or the other exclusively?

(Of the two, I would lean toward x509_cert, since it and pkcs8 seem to be from the same project, RustCrypto.)

adapter/ph/src/pki.rs
```diff
@@ -30,13 +51,138 @@ pub enum ParseError {
...
+    DecodeError,
+
     #[error("Key Error")]
     KeyError,
 
     #[error("I/O error {0}")]
     IOError(#[from] std::io::Error),
 }
 
+///Public key extracted from a certificate.
```
cpacejo, line 64: Rather than defining our own `PubKey` and `Cert` types, would it be possible to reuse or alias the corresponding types from the RustCrypto project?  i.e. [spki::SubjectPublicKeyInfoOwned](https://docs.rs/spki/latest/spki/type.SubjectPublicKeyInfoOwned.html) and [x509_cert::Certificate](https://docs.rs/x509-cert/latest/x509_cert/certificate/type.Certificate.html) (or TbsCertificate)?

adapter/ph/src/pki.rs
```diff
@@ -30,13 +51,138 @@ pub enum ParseError {
...
+        f.debug_struct("Cert")
+            .field("cn", &self.common_name())
+            .finish()
+    }
+}
+
+impl Cert {
+    /// Construct from DER bytes
+    pub fn from_der(der: &[u8]) -> Result<Cert, ParseError> {
+        X509Certificate::from_der(der).map_err(|e| {
```
cpacejo, line 92: Here we parse the DER and throw away the result, only to parse it again on any method call.  Can we instead store the parsed result and reuse it?

libnode2/src/rsa_sign.rs
```diff
@@ -0,0 +1,119 @@
+//! RSA signinging helpers using aws-lc-rs
```
cpacejo, line 1: Typo, "signining" -> "signing"

adapter/ph/src/address_pool.rs
```diff
@@ -69,7 +69,7 @@ impl AddressPool {
         addr_bytes[..6].copy_from_slice(&base_addr.segments()[..6]);
 
         let mut buf = [0u8; 8];
-        rand_bytes(&mut buf).unwrap();
+        OsRng.try_fill_bytes(&mut buf).expect("OS RNG Failed");
```
cpacejo, line 72: Nothing wrong with using the `rand` engine here.  I wonder though, since we are using the `aws_lc_sys` RNG elsewhere, if we should standardize on that.  (e.g. [aws_lc_sys::rand::fill()](https://docs.rs/aws-lc-rs/latest/aws_lc_rs/rand/fn.fill.html))

adapter/ph/src/link_state.rs
```diff
@@ -578,7 +578,7 @@ impl LinkStateWrapper {
             let require_ca_signature = asm.config.get().ca_file.is_some();
             let (cert, is_verified) = match peer_cert {
                 PeerCertificate::Unverified(cert) => {
-                    warn!(target: LINK_STATE, "{} has unverified name {:?}", asm.formatted_link_id(link_id), cert.subject_name());
```
cpacejo, line 587: Note that the common name is just a part of the subject name.  (In ZPR, we do require that the subject name only has this one component.)  Let's retain the behavior of printing the full subject name here though to aid in debugging.

cpacejo (dismissed):
This looks good and is very thorough!!  Thanks for your patience as I worked through the changes.  A few minor comments here and there.  Mainly I wonder if it's possible (a) to handle all the certificate parsing with x509_cert, I do see some relevant methods (e.g. `Certificate::parse_pem_chain()`), and (b) use aws_lc_sys for all random number generation (it seems we're using it in some places and `rand` in others).

This all looks correct though so feel free to rebase & address any changes in a separate PR.

cpacejo (approved):

PR #1345, opened 2026-07-27
update quinn-proto to 0.11.16 (addresses dependabot #88)

cpacejo:
Quick review to address <https://github.com/org-zpr/zpr-core/security/dependabot/88>.  This is simply the result of `cargo update quinn-proto`.

Note – I never use the dependabot proposed fixes because they update/add to Cargo.toml, which is almost always wrong in my opinion (the exception would be if a major version bump/incompatible change is required), since we have no semantic dependency on the newer version.  We should always prefer to simply bump the Cargo.lock version, especially in cases where we do not have a direct dependency like this one.


ianchang1 (approved):
Looks good! Bumping the Cargo lock instead makes sense

PR #1347, opened 2026-08-03
Ssl fixes

ianchang1:
Addresses all review comments on the opensslswitch pr request. I branched off, fixed everything, and than split into three commits. 1. Dropped the Arc around RSA private key everywhere Clone wasn't derived. 2. Removed x509-parser (functionality replaced with x509-cert) and removed custom Pubkey and Cert structs for x509_cert::Certificate and spki::SubjectPublicKeyInfoOwned. 3. Swapped out OsRNg with aws_lc_s::rand::fill.

### DIFF
--- a/adapter/ph/src/pki.rs
+++ b/adapter/ph/src/pki.rs
@@ -137,17 +137,25 @@ pub fn verify(cert: &Certificate, key: &SubjectPublicKeyInfoOwned) -> Result<boo
         ParseError::DecodeError
     })?;
     let sig = cert.signature().raw_bytes();
-    let alg = &cert.signature_algorithm().oid;
+    let sig_alg = &cert.signature_algorithm().oid;
+    let key_alg = &key.algorithm.oid;
     let raw_key = key.subject_public_key.raw_bytes();
-
-    if *alg == OID_SHA256_WITH_RSA {
+    if *key_alg == OID_RSA_ENCRYPTION {
+        if *sig_alg != OID_SHA256_WITH_RSA {
+            error!(target: KEY_MGMT, "signature algorithm {sig_alg} is not valid for an RSA issuer key");
+            return Err(ParseError::KeyError);
+        }
         let vk = UnparsedPublicKey::new(&signature::RSA_PKCS1_2048_8192_SHA256, raw_key);
         Ok(vk.verify(&tbs, sig).is_ok())
-    } else if *alg == OID_ED25519 {
+    } else if *key_alg == OID_ED25519 {
+        if *sig_alg != OID_ED25519 {
+            error!(target: KEY_MGMT, "signature algorithm {sig_alg} is not valid for an Ed25519 issuer key");
+            return Err(ParseError::KeyError);
+        }
         let vk = UnparsedPublicKey::new(&signature::ED25519, raw_key);
         Ok(vk.verify(&tbs, sig).is_ok())
     } else {
-        error!(target: KEY_MGMT, "unsupported certificate signature algorithm: {alg}");
+        error!(target: KEY_MGMT, "unsupported issuer public key algorithm: {key_alg}");
         Err(ParseError::KeyError)
     }
 }
@@ -254,7 +262,8 @@ pub fn get_cn_from_cert(cert: &Certificate) -> Option<String> {
     common_name(cert)
 }
 
-const ID_X25519: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.3.101.110");
+const OID_X25519: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.3.101.110");
+const OID_RSA_ENCRYPTION: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.840.113549.1.1.1");
 const OID_SHA256_WITH_RSA: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.840.113549.1.1.11");
 const OID_ED25519: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.3.101.112");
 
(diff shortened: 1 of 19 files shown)


libnode2/Cargo.toml
```diff
@@ -15,8 +15,7 @@ tokio-util = { workspace = true }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true, optional = true }
 zpr = { workspace = true }
-rand = "0.9.3"
-x509-parser = "0.18"
+x509-cert = "0.3.0"
```
cpacejo, line 18: Since this is common to both the adapter and libnode, let's move it up to the toplevel Cargo.toml to ensure we keep everything at the same version.  (And so here you would just use `{ workspace = true }` to reference that version.)

adapter/ph/src/pki.rs
```diff
@@ -61,125 +57,98 @@ pub enum ParseError {
     IOError(#[from] std::io::Error),
 }
 
-///Public key extracted from a certificate.
-pub struct PubKey {
-    key: Vec<u8>,
+/// Construct from DER bytes
+pub fn from_der(der: &[u8]) -> Result<Certificate, ParseError> {
+    Certificate::from_der(der).map_err(|e| {
+        error!(target: KEY_MGMT, "error parsing DER certificate: {e}");
```
cpacejo, line 63: For helper functions like these it's better not to log, and leave that decision up to the callers.  Both because emitting log messages couples the helper functions to the log system, and also prevents callers from being able to "try" the function possibly expecting failure without it chattering to the log.

(For this function e.g. consider a caller who wants to first try parsing as DER, and if that fails, try parsing as PEM.  In that case, we would only want to emit a log message if both methods failed.)

Since you've defined an error type which implements `Error` with clear error messages, the callers can easily log if they so choose with something like `from_der(...).inspect_err(|e| error!(target: KEY_MGMT, "{e}")?`.

Also, separating out logging allows us to move this functionality into a separate common crate (see my file-level comment).

libnode2/src/pki.rs
cpacejo, line 1: Since the adapter and libnode2 share a lot of this pki functionality, and it's not really ZPR-specific, this may live better in the `zpr-utils` crate (which is in the `zpr-utils` repo).  E.g. we have some IP address helper functions there too which live there.

adapter/ph/src/pki.rs
```diff
@@ -273,11 +250,13 @@ pub fn load_noise_public_key(path: &Path) -> Result<[u8; NOISE_KEY_LEN], ParseEr
 
 /// Get the CN value as a string out of the certificate. If not found or any
 /// other issue, returns None.
-pub fn get_cn_from_cert(cert: &Cert) -> Option<String> {
-    cert.common_name()
+pub fn get_cn_from_cert(cert: &Certificate) -> Option<String> {
```
cpacejo, line 261: Is there any reason to keep this and `common_name()` as two separate functions still?

adapter/ph/src/pki.rs
```diff
@@ -273,11 +250,13 @@ pub fn load_noise_public_key(path: &Path) -> Result<[u8; NOISE_KEY_LEN], ParseEr
...
+    common_name(cert)
 }
 
 const ID_X25519: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.3.101.110");
+const OID_SHA256_WITH_RSA: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.840.113549.1.1.11");
+const OID_ED25519: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.3.101.112");
```
cpacejo: Let's be consistent with `ID_` vs. `OID_` prefixes.  (I'm fine with either tho `OID` is slightly better in my opinion.)

adapter/ph/src/pki.rs
```diff
@@ -61,125 +57,98 @@ pub enum ParseError {
...
+/// Verify a certificate's signature against the presumed issuer's
+/// public key
+pub fn verify(cert: &Certificate, key: &SubjectPublicKeyInfoOwned) -> Result<bool, ParseError> {
+    let tbs = cert.tbs_certificate().to_der().map_err(|e| {
+        error!(target: KEY_MGMT, "error re-encoding TBS certificate: {e}");
+        ParseError::DecodeError
+    })?;
+    let sig = cert.signature().raw_bytes();
+    let alg = &cert.signature_algorithm().oid;
+    let raw_key = key.subject_public_key.raw_bytes();
```
cpacejo, line 142: Here we are using the certificate's declaration of signature algorithm to dictate how we interpret the issuer's public key.  This is a potential security issue if e.g. strong public keys of algorithm A happen to correspond to weak public keys of algorithm B.  An adversary could then forge a certificate "signed" by an issuer with a (strong) public key of algorithm A, by cracking that public key as interpreted as a (weak) key of algorithm B, and signing with that.

We should instead use the algorithm from the issuer's key.  And while I think this is sufficient from a security standpoint, it's probably a good idea to validate that the stated algorithms match.

libnode2/src/rsa_sign.rs
cpacejo, line 1: This is actually a good candidate to move to `zpr-utils` also.

libnode2/src/pki.rs
```diff
@@ -19,14 +19,26 @@ pub enum ParseError {
     IOError(#[from] std::io::Error),
 }
 
+/// Isolate the first PEM block
+pub fn first_pem_block(text: &str) -> Option<&str> {
```
cpacejo, line 23: I think [`pem::decode_label()`](https://docs.rs/pem-rfc7468/latest/pem_rfc7468/fn.decode_label.html) does this same operation?  (Including the initial bytes->string conversion which we're doing at the call sites of this function.)

cpacejo (changes requested):
Nice looks good!  A few minor nits here and there.  One security issue in signature verification which I missed earlier.  And one suggestion to move the non-ZPR-specific PKI/RSA stuff out into the zpr-utils crate which is intended for such things.  (This would require a separate PR first to add those methods to zpr-utils.)

Only the security issue is a blocker for checking this in, beside that feel free to address the other issues in a separate PR.

ianchang1:
Added in that one security fix in the verify function. Going to address other changes in a separate PR as you mentioned

cpacejo (dismissed):
Looks good!

cpacejo (approved):

PR #1348, opened 2026-08-04
Ingress egress symmetric keys

### LINKED ISSUE #1087: Convey A2A MICV key to egress adapter
Using new BindEgressStream message, with help from Visa Service

ianchang1:
First half of ingress egress symmetric key implementation. Work split in three commits based on steps provided by Chris.

### DIFF
--- a/adapter/ph/src/mgmt/requests.rs
+++ b/adapter/ph/src/mgmt/requests.rs
@@ -50,14 +50,15 @@ pub fn send_echo_request(asm: &Assembly, link_id: LinkId) -> Sent<'_> {
 }
 
 /// send a Hello Request (RFC 6.5 § 6.3.4)
-///
-/// Originally this was used to send the pre-configured ZPR address of the
-/// remote adapter into the node.  This is no longer necessary.
-///
-pub fn send_hello_request(asm: &Assembly, link_id: LinkId) -> Sent<'_> {
+pub fn send_hello_request(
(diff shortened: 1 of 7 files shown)


adapter/ph/src/mgmt/requests.rs
```diff
@@ -50,14 +50,15 @@ pub fn send_echo_request(asm: &Assembly, link_id: LinkId) -> Sent<'_> {
...
 /// send a Hello Request (RFC 6.5 § 6.3.4)
-///
-/// Originally this was used to send the pre-configured ZPR address of the
-/// remote adapter into the node.  This is no longer necessary.
-///
-pub fn send_hello_request(asm: &Assembly, link_id: LinkId) -> Sent<'_> {
+pub fn send_hello_request(
+    asm: &Assembly,
+    link_id: LinkId,
+    pub_key: x25519_dalek::PublicKey,
```
cpacejo, line 56: Let's name this also `a2a_dh_pubkey` so it's clear to callers which key needs to be supplied (as opposed to the Noise key used for link establishment or the RSA key used for authentication…)

Also apologies I've been inconsistent with pub_key vs. pubkey :)  feel free to make a judgement call on that one haha

cpacejo (approved):
Nice!  Just a tiny naming nit but feel free to address that bundled with a later PR in this series.

PR #1355, opened 2026-08-10
Ic/sslfinalcleanup

ianchang1:
Addressing changes from Chris following my removal of the OpenSSL dependancy. Four commits:

1. move aws-lc-rs and x509-cert deps to workspace

2. stop logging in adapter/ph/src/pki.rs. The ParseError variants now carry the underlying error text so callers can log or wrap as they choose. Also dropped get_cn_from_cert (redundant with common_name) and the tests' local ID_ED25519 which was changed to a constant OID_ED25519.

3. drop first_pem_block and libnode2 pki module. As Chris noted Pem::decode_vec already strips the preamble, so first_pem_block was unecessary. libnode2/src/pki.rs now had no callers as functionality has been replaced so I deleted the module.

4. Some fixes in error messages in commit 2


owenrhysthomas (approved):
Looks good to me

PR #1356, opened 2026-08-10
moved rsa-sign to zpr utils

ianchang1:
Second half of rsa sign -> utils.


adapter/ph/src/auth.rs
```diff
@@ -24,6 +23,7 @@ use thiserror::Error;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use x509_cert::Certificate as X509Certificate;
+use zpr_utils::rsa_sign::{load_rsa_key, sign_rsa_key};
```
owenrhysthomas, line 26: I would be inclined to just do `use zpr_utils::rsa_sign` and then call the functions with `rsa_sign::load_rsa_key()` or `rsa_sign::sign_rsa_key()`, just to make it explicit that the function is coming from an external package. But I don't think it's necessary (or even necessarily the best choice, I'd be interested what @cpacejo thinks).

owenrhysthomas (approved):
Looks good to me overall. The comment can be addressed or not.

PR #1365, opened 2026-08-19
switch KM Noise protocol from IK to XX mode

### LINKED ISSUE #1286: Use a noise pattern with no preshared key
Current noise pattern in use in our KM requires caller to have public key for receiver.  There is another pattern we can use that does not require sharing the public key.  Let's try using that, then we can drop the config requirement of a shared key.

cpacejo:
Since adapters validate the node's certificate, we have no need for adapters to validate the node's public key directly, or to have it beforehand.

To support this, we must switch Noise from an ?K mode to a ?X mode.  NX would be sufficient for most adapters, but not for the VS adapter which needs to authenticate itself to the node, and not for upcoming node-node authentication.  IX saves one message on paper, but (like XX) cannot encrypt the KM payload before the 1st message, which is necessary to hide the HMAC, so a 3rd message is needed anyway.  XX is 3 messages, but encrypts the 2nd and 3rd so we can piggyback on both to exchange KM payloads, and it provides better identity hiding than IX which may be desirable.

If we change the KM payload exchange so that only the responder generates HMAC keying material, we could use IX if we want to save one message, at the expense of identity hiding.  Link establishment isn't necessarily latency-sensitive so for now XX is the best option.

This required also changing the KM payload mechanism to understand that there are now 3 messages, and that the 1st isn't to be used to send a payload.

Also remove no-longer-needed --node-public-key-file argument.

This of course removes the ability for adapters to ensure that they're connecting to a _specific_ node, but we should do that by allowing adapters to specify a CN they are expecting, if this is desired.

Closes #1286.

### DIFF
--- a/adapter/ph/src/km_noise.rs
+++ b/adapter/ph/src/km_noise.rs
@@ -72,6 +72,9 @@ pub struct KmNoise {
     /// which message in the handshake sequence is next to send or receive
     msg_nr: u8,
     initiate: bool,
+    /// Should we abort if we cannot verify the certificate received from the peer?
+    /// (Has no effect if the KmCertificateExchange is not configured for validation.)
+    verify_cert: bool,
     local_keypair: NoiseKeypair,
     hs_sent_t: Option<Instant>,
     hs_state: Option<snow::HandshakeState>,
@@ -189,6 +192,7 @@ impl KmNoise {
     /// - `certx` is the certificate exchange creator/verifier.
     pub fn new(
         initiate: bool,
+        verify_cert: bool,
         local_keypair: Option<NoiseKeypair>,
         zpis: ZPIPair,
         certx: KmCertExchange,
@@ -211,6 +215,7 @@ impl KmNoise {
             state: KmSMState::Configuring,
             msg_nr: 0,
             initiate,
+            verify_cert,
             local_keypair: kp,
             hs_sent_t: None,
             hs_state: None,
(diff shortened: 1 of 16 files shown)


adapter/ph/src/km_noise.rs
```diff
@@ -201,8 +209,8 @@ impl KmNoise {
         Ok(KmNoise {
             settings,
             state: KmSMState::Configuring,
+            msg_nr: 0,
```
mkolehmainen, line 216: `msg_nr` set to 0 here but not in reset.

adapter/ph/src/km_noise.rs
mkolehmainen, line 420: Should set `msg_nr=0` here.

adapter/ph/src/km_noise.rs
```diff
@@ -477,24 +487,27 @@ impl KeyManagerStateMachine for KmNoise {
...
+                            return Err(KmError::CertExchangeError);
+                        }
+                    };
+                    match self.parse_km_payload(&payload[..len], peer_pubkey) {
+                        Ok(_) => {}
+                        Err(_) => {
+                            self.state = KmSMState::Error;
+                            self.hs_state = Some(hs);
+                            return Err(KmError::HandshakeError);
+                        }
```
mkolehmainen, line 531: Right here after this match consider checking if the cert received fails verification and then do not send msg3.  Eg, 
```rust
if self.initiate 
   && self.certx.requries_verification() 
   && !matches(self.peer_cert, Some(PeerCertificate::Unverified(_))) { error_state... }
```
Just needs that additional function on `km_cert_exchange.rs` (`self.authority_cert.is_some()`)

adapter/ph/src/km_noise.rs
mkolehmainen: A responder holding a post-msg2 HandshakeState has no deadline at all.
Drop `self.initiate` and set `hs_sent_t`  when `create_hs_message` succeeds up at line 525.

mkolehmainen: Not in scope of this PR but the link-state also does not have a timeout for KEYING.

mkolehmainen (changes requested):
Great!! Can't wait to use this! Just a couple tiny things.

adapter/ph/src/km_noise.rs
cpacejo: Ah good catch!

cpacejo: Opened #1370 to address the KEYING timeout.

adapter/ph/src/km_noise.rs
cpacejo: Ended up adding an extra argument to request verification or not.  For now it's the same as whether we're initiating, but we'll need this for the node-node case.

mkolehmainen (commented):
Nice! LGTM!

mkolehmainen (approved):
LGTM

PR #1366, opened 2026-08-19
Ic/a2apubkey468

ianchang1:
I realized that I had lost steps 4, 6 and 8 of adding A2A public keys when I switched tasks to working on the OpenSSL switch fixes. Thankfully Claude was able to find my old changes in my VSCode history. Each step is split into its own commit.

Step 4: Receiving the key on the node. handle_hello_request now handles the A2A_DH_PUBKEY TLV through a new process_a2a_dh_pubkey_tlv, which follows process_window_size_tlv. The key gets stored on peer_table::PeerState as a2a_dh_pubkey.

Step 6: Sending the adapter's key to the VS. process_acquire_zpr_address_request pulls the key out of peer state and passes it to build_connect_request as a new argument. build_connect_request stuffs it into a vsapi_types::PublicKey and then into the ConnectRequest. If the adapter never sent a key we send a zero length one, which is what the VS reads as absent. Also bumped the zpr dep to v0.22.1 to pick up the PublicKey and ConnectRequest changes from step 5. 

Step 8: Sending the node's own key. NodeConnect gets a new a2a_dh_pubkey field and do_connect packages it into a second Param next to ZPR_ADDR. vs_worker::launch derives the public key from asm.a2a_dh_keypair. This commit also renames the send_hello_request parameter from pub_key to a2a_dh_pubkey, which is the naming cleanup I said I would get to. NodeConnect and ConnectRequest both picked up required fields, so the two lntest construction sites in libnode2/src/cli/handler.rs had to be filled in as well. I gave them random 32 byte keys using aws_lc_rs::rand::fill, which is already used in that file for the fake substrate address.

Added one more additional commit changing pub_key -> pubkey in different locations.


.github/workflows/adapter.yml
cpacejo, line 1: Can you re-check-in without the formatting changes?  They will obscure line-by-line history when using `git blame`.

adapter/ph/src/peer_table.rs
```diff
@@ -39,6 +39,7 @@ pub struct PeerState {
     pub mgmt_processor: queues::MgmtProcessor,
     pub mgmt_processor_worker: task::JoinHandle<()>,
     pub auth_key: [u8; 32], // set in ::new and never changed.
+    pub a2a_dh_pubkey: Mutex<Option<x25519_dalek::PublicKey>>,
```
cpacejo: `Mutex` is fine since this isn't used in high performance places, but since it is write-rarely/once, read-often, `RcuBox` would be a better match (see `transport_sa` below which is an `RcuBox`).

(RCU / read-copy-update is a synchronization mechanism with very cheap / free reads, which do not block each other or writers.  It's used often in networking code.)

cpacejo (changes requested):
Looks good thanks!  Just a couple small changes.

ianchang1:
Think I fixed everything my auto formatter must have swapped the single for double quotation marks

cpacejo (approved):
Nice looks good!

PR #1367, opened 2026-08-19
use fwd_pep next_hop when routing between nodes

cpacejo:
Fix route lookup to correctly heed the `next_hop` value from the `fwd_pep` of a Visa if present.  This is required for multinode.


owenrhysthomas (approved):
👍

PR #1371, opened 2026-08-21
step 12

ianchang1:
Step 12 of A2A PubKey. 

Ingress and egress node need to share these keys with their respective adapters. On both ingress and egress side, added ZDP protocol changes. In mgmt::requests::send_bind_actor_address_success_response() (and egress equivalent), added an extra parameter peer_a2a_dh_pubkey: Option<&x25519_dalek::PublicKey>. Wrote this to the end of the packet with first key len and then key itself. 

In mgmt::handlers::handle_bind_actor_address_response() (and egress equivalent), deserialized the key.  Pass the key as a new argument to adapter::install_tether().

### DIFF
--- a/adapter/ph/src/mgmt/handlers.rs
+++ b/adapter/ph/src/mgmt/handlers.rs
@@ -46,6 +46,8 @@ pub enum HandleMgmtError {
     LinkClosed,
 }
 
+const X25519_KEY_LEN: usize = 32;
+
 impl From<&HandleMgmtError> for counters::ManagementCounterType {
     fn from(err: &HandleMgmtError) -> Self {
         match err {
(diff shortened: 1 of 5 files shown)


adapter/ph/src/mgmt/handlers.rs
```diff
@@ -679,6 +681,26 @@ pub async fn handle_bind_egress_stream_request(
         return Err(HandleMgmtError::BadStructure);
     };
 
+    if pkt.remaining() < size_of::<u16>() {
+        return Err(HandleMgmtError::BadStructure);
+    }
+
+    let key_len = pkt.get_u16() as usize;
+
+    let peer_a2a_dh_pubkey = if key_len == 0 {
```
cpacejo, line 690: Another way you can write this is to put `let peer_a2a_dh_pubkey;` on a line by itself.  Then in the if branches, you can initialize `peer_a2a_dh_pubkey = XXX;`.  This can help avoid the weird nesting, indentation, and control flow that crop up in cases like this.

cpacejo (approved):
Looks good!  Just a minor formatting suggestion which is OK to address in a subsequent PR.

PR #1372, opened 2026-08-24
Encrypt short packets in place, into their own tailroom (#385)

### LINKED ISSUE #385: eliminate memcpy during encryption for short packets
"Short" packets (i.e. practically all packets) can simply be encrypted into a different location in the source buffer.

ZprBot1:
Closes #385.

`encrypt_full()` allocated a 12 KiB zeroed stack buffer (`PACKET_BUFFER_SIZE`) per packet, encrypted into it, then copied the ciphertext back over the packet body. Both costs are avoidable for any packet whose own tailroom can hold the ciphertext — which, as the issue notes, is practically all of them.

The ciphertext is now written directly into the packet's tailroom and the one-byte ZPI header is relocated in front of it; nothing else moves:

```
before:  [ metadata ][ headroom ][ ZPI ][ plaintext ][ tailroom ........... ]
after:   [ metadata ][ .............. dead .......... ][ ZPI ][ ciphertext  ]
```

Packets too large for their own tailroom keep the old scratch-buffer path, and a test asserts both paths produce byte-identical output.

Measured on a 64-byte body with a codec that expands by 24 bytes (release build, buffer allocation excluded from both paths):

```
in-place        28 ns/pkt
scratch buffer 192 ns/pkt
```

### Supporting changes

- `Codec` gains `max_overhead()` — the bound on how much a ciphertext may exceed its plaintext. Callers previously had no way to size an output buffer; `decrypt_with_sa()` works around that today by hardcoding `NOISE_PADLEN` behind a `km_impl == KM_ID_NOISE` check, which breaks the moment another codec exists. (That hardcode is left in place here; removing it is a natural follow-up.)
- `Packet` gains `body_offset()`, `tailroom_available()`, `body_and_tailroom_mut()`, `set_body_extent()` and `adopt_tailroom_with_header()` to express the read-body/write-tailroom transform safely (the body and tailroom are disjoint `split_at_mut` halves of the backing buffer, so plaintext/ciphertext overlap is impossible by construction).
- `encrypt_full()` / `decrypt_full()` lose their unused `&Assembly` parameter.
- `encrypt_full()` now rejects a body shorter than the ZPI header instead of underflowing on the length subtraction.

### Deviations from the plan posted on the issue

None of substance. The plan's open question (does anything downstream of egress rely on the packet retaining tailroom/offset?) got no answer; I proceeded on 'no' — `substrate_egress()` queues the packet and only `body()` is written to the socket. Flagging it here for review attention.

### Deliberately out of scope

`decrypt_full()` and `km::encrypt_transport_zdp()`/`decrypt_transport_zdp()` carry the identical 12 KiB scratch buffer. The issue is scoped to encryption; the decrypt side is a follow-up.

### Testing

- `test_encrypt_full_uses_tailroom_in_place` — ciphertext lands in the former tailroom, ZPI byte survives, offset moved as expected.
- `test_encrypt_full_falls_back_when_tailroom_too_small` — fallback fires and produces byte-identical output to the in-place path.
- `test_encrypt_full_decrypt_full_round_trip` — with an expanding codec (8-byte nonce + 16-byte tag).
- `test_encrypt_full_rejects_empty_packet`, `test_encrypt_full_zpi_only_packet` — edge cases, including the header self-copy degenerate case.
- `bench_encrypt_full_paths` (`#[ignore]`, run with `--ignored --nocapture`) — source of the numbers above; keeps the old implementation in the test module for comparison.

Full gate: `cargo build --all-targets`, `cargo fmt --check`, `cargo test` (222 passed), `-D warnings` clean.

### DIFF
--- a/adapter/ph/src/fastpath.rs
+++ b/adapter/ph/src/fastpath.rs
@@ -600,8 +600,12 @@ pub fn encrypt_hmac(send_hmac_key: [u8; 32], pkt: &mut Packet) {
 ///
 /// ```text
 /// before:  [ metadata ][ headroom ][ ZPI ][ plaintext ][ tailroom ........... ]
-/// after:   [ metadata ][ .............. dead .......... ][ ZPI ][ ciphertext  ]
+/// after:   [ metadata ][ ....... dead .........][ ZPI ][ ciphertext ][ tail.. ]
 /// ```
+///
+/// The ciphertext starts exactly where the old tailroom started, the relocated
+/// ZPI byte overwrites the tail of the old plaintext, and whatever tailroom the
+/// ciphertext (plaintext + AEAD overhead) does not consume remains after it.
 ///
 /// Only the one-byte ZPI header is moved.  Packets too large for their own
 /// tailroom to hold the ciphertext fall back to the scratch-buffer path; for
--- a/adapter/ph/src/fastpath.rs
+++ b/adapter/ph/src/fastpath.rs
@@ -592,17 +592,48 @@ pub fn encrypt_hmac(send_hmac_key: [u8; 32], pkt: &mut Packet) {
     pkt.put(&link_mac[..zdp::ZDP_PACKET_MAC_SIZE]);
 }
 
-pub fn encrypt_full(
-    _asm: &Assembly,
-    codec: &dyn Codec,
-    pkt: &mut Packet,
-) -> Result<(), km::EncryptionError> {
+/// Encrypt the body of a ZDP packet in place, leaving the ZPI header intact.
+///
+/// The ciphertext is written into the packet's own tailroom rather than into a
+/// scratch buffer, which avoids both a `PACKET_BUFFER_SIZE` stack zeroing and a
+/// copy of the ciphertext back into the packet:
+///
+/// ```text
+/// before:  [ metadata ][ headroom ][ ZPI ][ plaintext ][ tailroom ........... ]
+/// after:   [ metadata ][ .............. dead .......... ][ ZPI ][ ciphertext  ]
+/// ```
+///
+/// Only the one-byte ZPI header is moved.  Packets too large for their own
+/// tailroom to hold the ciphertext fall back to the scratch-buffer path; for
+/// the packet sizes seen in practice the fast path always applies.
+pub fn encrypt_full(codec: &dyn Codec, pkt: &mut Packet) -> Result<(), km::EncryptionError> {
     // TODO: Could do some length checks here on the packet body.  Is it too short? Too long? Etc.
 
     let zpi_hdr_len = std::mem::size_of::<zdp::ZdpZpiHeader>(); // = 1
 
-    let mut enc_buf = [0u8; config::PACKET_BUFFER_SIZE];
+    if pkt.body().len() < zpi_hdr_len {
+        return Err(km::EncryptionError::ParseError);
+    }
+
     let encr_len = pkt.body().len() - zpi_hdr_len; // Everything except the ZPI byte
+
+    // Can the ciphertext fit in this packet's own tailroom?  If so we can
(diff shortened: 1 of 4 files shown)


adapter/ph/src/packet.rs
```diff
@@ -476,6 +476,67 @@ impl Packet {
...
+    pub fn adopt_tailroom_with_header(&mut self, header_len: usize, written: usize) {
+        assert!(header_len <= self.metadata().len);
+        assert!(written <= self.tailroom_available());
+
+        let old_offset = self.metadata().offset;
+        let output_offset = old_offset + self.metadata().len;
+        let new_offset = output_offset - header_len;
+
+        // Regions cannot overlap unless the body is entirely header, in which
+        // case source and destination coincide and the copy is a no-op.
```
cpacejo: This comment seems wrong.  I'd think that whenever the header is more than half the length of the old body, the regions would overlap.  (Which is OK, since `.copy_within()` correctly handles overlapping regions.)

adapter/ph/src/fastpath.rs
```diff
@@ -938,6 +964,200 @@ mod test {
...
+                pkt.put(&enc_buf[0..len]);
+                Ok(())
+            }
+            Err(e) => Err(e),
+        }
+    }
+
+    #[test]
+    #[ignore]
+    fn bench_encrypt_full_paths() {
```
cpacejo, line 1109: Nice, glad to see a benchmark!

adapter/ph/src/fastpath.rs
```diff
@@ -592,18 +592,49 @@ pub fn encrypt_hmac(send_hmac_key: [u8; 32], pkt: &mut Packet) {
...
-) -> Result<(), km::EncryptionError> {
+/// Encrypt the body of a ZDP packet in place, leaving the ZPI header intact.
+///
+/// The ciphertext is written into the packet's own tailroom rather than into a
+/// scratch buffer, which avoids both a `PACKET_BUFFER_SIZE` stack zeroing and a
+/// copy of the ciphertext back into the packet:
+///
+/// ```text
+/// before:  [ metadata ][ headroom ][ ZPI ][ plaintext ][ tailroom ........... ]
+/// after:   [ metadata ][ .............. dead .......... ][ ZPI ][ ciphertext  ]
```
cpacejo: I think this diagram isn't quite right.  `ciphertext` should be left-aligned with the old `tailroom`, no?  And there may/will be remaining tailroom after `ciphertext`.

cpacejo (changes requested):
Overall looks good!  Just a couple comments which don't seem quite right.

adapter/ph/src/fastpath.rs
ZprBot1: You are right on both counts — fixed in c6e3cbf: the `after` diagram now shows `ciphertext` left-aligned with the old tailroom, with the remaining tailroom shown after it.

adapter/ph/src/packet.rs
ZprBot1: Correct — the comment was wrong. Fixed in c6e3cbf: it now states that the regions overlap whenever the header is more than half the old body length (new_offset < old_offset + header_len iff len < 2 * header_len), which `copy_within` handles safely via memmove semantics.

adapter/ph/src/fastpath.rs
ZprBot1: Thanks!

ZprBot1:
Both doc issues fixed in c6e3cbf (diagram alignment in `encrypt_full`, overlap comment in `adopt_tailroom_with_header`). Full gate re-run clean (build, fmt, 222 tests, -D warnings). @cpacejo re-requested review.

cpacejo (approved):
Looks great thanks!

PR #1373, opened 2026-08-24
Ic/a2 a final steps

ianchang1:
Final steps of A2APubKey! 

Step 13: Stored A2A key material is adapter_tables. When using blake3::derive_key() I wasn't sure what to use for the context parameter so I added a const to zdp.rs based on the derive_key() specs

Step 14: Grabbed key from visa and passed it to send_bind_actor_address_success_response() (did similar for egress side). Performed diffie_hellman. I also added in two helpers into visa_table.rs to assist in grabbing from visa. 

Step 15:  Added the blake3::keyed_hash() to both sides if a key is present and if None defaults back to the blake3::hash() code

### DIFF
--- a/adapter/ph/src/fastpath.rs
+++ b/adapter/ph/src/fastpath.rs
@@ -310,14 +310,17 @@ impl FastpathWorker {
         match &*entry {
             EltEntry::Active(pep) => {
                 // compute A2A MAC
-                // TODO: use actual A2A SAID & keyed hash
+                // TODO: use actual A2A SAID
                 // See https://github.com/org-zpr/zpr-core/issues/1349
                 let a2a_said: A2aSaid = 0;
                 let a2a_mac_size = zdp::ZDP_A2A_MAC_SIZE; // TODO: may be smaller depending on A2A SAID
                 let mut a2a_mac = [0u8; zdp::ZDP_A2A_MAC_SIZE];
+                let a2a_micv = match pep.a2a_micv_key {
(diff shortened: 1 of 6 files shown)


adapter/ph/src/fastpath.rs
```diff
@@ -310,14 +310,17 @@ impl FastpathWorker {
         match &*entry {
             EltEntry::Active(pep) => {
                 // compute A2A MAC
-                // TODO: use actual A2A SAID & keyed hash
+                // TODO: use actual A2A SAID
                 // See https://github.com/org-zpr/zpr-core/issues/1349
```
cpacejo, line 314: This issue link is also no longer relevant (it being a dup of the issue you're addressing with this commit).

adapter/ph/src/zdp.rs
```diff
@@ -302,6 +302,9 @@ pub mod traffic_classifier_flags {
 /// Config-specified size of A2A MAC.  Algorithm-specified MAC may be smaller (but not larger).
 pub const ZDP_A2A_MAC_SIZE: usize = 8;
 
+/// BLAKE3 key-derivation context for the A2A MICV key.
+pub const ZDP_A2A_MICV_KEY_CONTEXT: &str = "org-zpr zpr-core 2026-08-21 a2a micv key";
```
cpacejo, line 306: Yes this seems fine to me!

cpacejo (approved):
Yes looks great!  I'm so excited for this.  Just one comment nit (since you're closing out the issue it's referencing), feel free to grab that in another PR if that's more convenient.

PR #1377, opened 2026-08-25
Decrypt short packets in place, into their own tailroom (#1376)

### LINKED ISSUE #1376: eliminate memcpy during decryption for short packets
As with encryption in #385, we should likewise be able to eliminate memcpy during decryption for "short" packets by decrypting them to a different location in the source buffer.

ZprBot1:
Closes #1376.

Decrypt-side counterpart of #1372 (#385): `decrypt_full()` now decrypts the ciphertext into the packet's own tailroom instead of a `PACKET_BUFFER_SIZE` scratch buffer, then relocates the one-byte ZPI header in front of the plaintext with `adopt_tailroom_with_header()`. This removes, per non-transit encrypted ingress packet:

* the 12 KiB `[0u8; PACKET_BUFFER_SIZE]` stack zeroing, and
* the full memcpy of the plaintext back into the packet.

Net data movement on the fast path is 1 byte (the ZPI header).

Because decryption never expands (plaintext = ciphertext − overhead), `encr_len` bytes of tailroom always bound the output — no `max_overhead()` term is needed on this side. Packets without that much tailroom keep the previous scratch-buffer behaviour, and codec failure on either path maps to `DecryptError::DecryptionFailure` exactly as before.

No new `Packet` or `Codec` API: this reuses `tailroom_available()`, `body_and_tailroom_mut()`, `adopt_tailroom_with_header()` and the structure checks (`padlen`) unchanged. `km::decrypt_transport_zdp()` is out of scope per the plan (dead code, off the fast path).

Downstream safety was confirmed on #385: nothing after ingress decryption relies on tailroom or packet position ("The same is true in the ingress direction").

Tests (mirroring the encrypt-side suite, same `ExpandingCodec`):

* `test_decrypt_full_uses_tailroom_in_place` — plaintext lands where the tailroom was, ZPI byte preserved;
* `test_decrypt_full_falls_back_when_tailroom_too_small` — identical bytes via the scratch path;
* `test_decrypt_full_rejects_short_packet`, `test_decrypt_full_codec_failure_is_reported`;
* existing `test_encrypt_full_decrypt_full_round_trip` now exercises the new path.

Gate: `cargo build --all-targets`, `cargo fmt --check`, `cargo test` (226 passed in `ph`, full workspace green), `-D warnings` clean. No plan deviations.

### DIFF
--- a/adapter/ph/src/fastpath.rs
+++ b/adapter/ph/src/fastpath.rs
@@ -713,19 +713,60 @@ fn decrypt_hmac(recv_hmac_key: [u8; 32], pkt: &mut Packet) -> Result<(), Decrypt
     Ok(())
 }
 
-/// Decrypt a packet using the given encryption codec.
+/// Decrypt a packet using the given encryption codec, leaving the ZPI header
+/// intact.
+///
+/// Mirrors `encrypt_full()`: the plaintext is written into the packet's own
+/// tailroom rather than into a scratch buffer, which avoids both a
+/// `PACKET_BUFFER_SIZE` stack zeroing and a copy of the plaintext back into
+/// the packet:
+///
+/// ```text
+/// before:  [ metadata ][ headroom ][ ZPI ][ ciphertext ][ tailroom ......... ]
+/// after:   [ metadata ][ ........ dead .........][ ZPI ][ plaintext ][ tail. ]
+/// ```
+///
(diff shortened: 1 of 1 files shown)


cpacejo (approved):
Looks good thanks!

PR #1378, opened 2026-08-25
Allow adapters to use self-generated keys (#1374)

### LINKED ISSUE #1374: Allow adapters to use self-generated keys
Nodes do not verify the certificates of non-special adapters.  They are therefore unimportant, as is the Noise keypair of such an adapter.

Let's make the following changes:

- keying (in km_noise / km_cert_exchange) should optionally allow no certificate to be specified
- adapters should be permitted not to specify --certificate-file (in which case, none is sent during keying)
- if --certificate-file is not specified, adapters should be permitted not to specify --private-key-file (in which case, a new static key is generated during keying – I believe km_noise already handles this)
- the integration tests can then be updated to no longer specify the above arguments for the zpr-a/-b/-c adapters (zpr-vs adapter still requires a cert and private key)

A nice (but unnecessary) next step will be to use NX instead of XX Noise handshake during keying instead of generating a temporary (and unneeded) static key.  This would require a new KM type number to distinguish the two cases.  (This should be handled as a separate issue.)

ZprBot1:
Closes #1374.

Since the IK→XX switch (6c7b02a), the Noise handshake authenticates the peer's static key on its own; a certificate exists only to bind a name to that key. Nodes never verify the certificates of non-special adapters, so those adapters no longer need a certificate at all.

## Changes

- **km_cert_exchange**: `KmCertExchange.local_cert` is now `Option<Certificate>`. With no local cert, `write_payload` emits an empty payload (`cert_len == 0`); the wire format is otherwise unchanged. `process_payload` returns `Ok(Option<PeerCertificate>)` — `cert_len == 0` yields `Ok(None)`; a present cert goes through the same parse / CA-verify / key-match checks as before.
- **km_noise**: `parse_km_payload` stores the `Option` directly in `peer_cert`. The verification gate is untouched: with `verify_cert && requires_verification()`, a missing or unverified peer cert still aborts the handshake, so an adapter connecting to a node still requires the node's verifiable certificate.
- **main.rs**: no longer fabricates a self-signed certificate when `--certificate-file` is absent — none is sent. `generate_self_signed_noise_cert` and its helpers are now `#[cfg(test)]`.
- **Config**: unchanged. Adapters could already omit `certificate_file` and `private_key_file` (validation still requires `name` in that case, for bootstrap CN derivation); a fresh static key is generated during keying when no private key is given. Node mode still hard-requires cert + key.

Downstream consumers of `sa.peer_cert` (`link_state`, `visa_mgmt`) already handled `None`.

## Tests

- `test_km_cert_no_local_cert_round_trip`: empty payload round-trip, accepted as "no cert".
- `test_noise_handshake_certless_initiator`: full XX handshake with a cert-less adapter against a node; node sees `peer_cert == None`, adapter still verifies the node's cert.
- `test_noise_handshake_aborts_when_required_cert_missing`: initiator that requires verification aborts with `CertExchangeError` when the responder presents no cert.
- Existing cert-exchange and handshake tests updated for the new signatures; full gate (build, fmt, 228 tests, `-D warnings`) clean.

## Out of scope

Switching to an NX handshake to avoid the throwaway static key — per the issue, that is a separate change requiring a new KM type number.

### DIFF
--- a/adapter/ph/src/visa_mgmt.rs
+++ b/adapter/ph/src/visa_mgmt.rs
@@ -80,17 +80,6 @@ pub fn build_connect_request(
         }
     }
 
-    let mut cn = get_common_name(asm, id)?;
-    if cn.is_empty() {
-        // Self-generated-key adapters (no certificate exchange) leave the SA
-        // without a peer cert, so there is no cert CN. Fall back to the CN
-        // asserted in the self-signed auth blob; the visa service
-        // authenticates that CN against the blob's RSA signature.
-        if let AuthBlob::SelfSigned(ss) = blob {
-            cn = ss.cn.clone();
-        }
-    }
-
     // Convert the adapter's AuthBlob to the vsapi_types AuthBlob for the v2 protocol.
     let vsapi_blob = match blob {
         AuthBlob::SelfSigned(ss) => vsapi_types::AuthBlob::SS(vsapi_types::SelfSignedBlob {
@@ -118,10 +107,17 @@ pub fn build_connect_request(
             value: addr.to_string(),
         });
     }
-    request_claims.push(vsapi_types::Claim {
-        key: claims::key::CN.into(),
-        value: cn,
-    });
+    // The CN claim always comes from the AuthBlob certificate, never from the
+    // SA cert: the SA cert only validates the node-adapter link, while the
+    // AuthBlob cert authenticates an actor to the VS (and multiple actors may
+    // sit behind one adapter, so the two CNs must be independent). With an
+    // AuthCode blob there is no AuthBlob cert, so no CN claim is sent.
+    if let AuthBlob::SelfSigned(ss) = blob {
+        request_claims.push(vsapi_types::Claim {
+            key: claims::key::CN.into(),
+            value: ss.cn.clone(),
+        });
+    }
 
     let connect_req = vsapi_types::ConnectRequest {
         blobs: vec![vsapi_blob],
@@ -130,33 +126,6 @@ pub fn build_connect_request(
         dock_interface: 0,
     };
     Ok(Some(connect_req))
-}
-
-/// WARNING! Using this ignores whether the certificate was verified or not.
-fn get_common_name(asm: &Arc<Assembly>, id: LinkId) -> Result<String, LinkStateError> {
-    let Some(peer_state) = asm.peer_table.get(id) else {
-        return Err(LinkStateError::NotFound(id));
-    };
-
-    let Some(sa) = peer_state.get_established_transport_association() else {
-        return Err(LinkStateError::InvalidOperation(
-            "Attempted to Register Actor Address when SA not established".to_owned(),
-        ));
-    };
-
-    // TODO: validate that DN *only* has CN, since this is what VS expects
-    // (or, teach VS about DNs)
-    let cn: String;
-
-    if let Some(ref peer_cert) = sa.peer_cert {
-        cn = peer_cert.common_name().unwrap_or_default();
-    } else {
-        cn = String::new();
-    }
-
-    debug!(target: VISA_MGMT, "{} CN is {cn}", asm.formatted_link_id(id));
-
-    Ok(cn)
 }
 
 /// This uses "RemoteDisconnect" as the reason passed to the visa service.
--- a/adapter/ph/src/visa_mgmt.rs
+++ b/adapter/ph/src/visa_mgmt.rs
@@ -80,7 +80,16 @@ pub fn build_connect_request(
         }
     }
 
-    let cn = get_common_name(asm, id)?;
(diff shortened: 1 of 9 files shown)


cpacejo (changes requested):
This all looks good thank you!  Can we additionally amend the integration tests not to provide the --certificate-file or --private-key-file arguments when starting the adapters (other than the VS adapter)?  The files those refer to can be removed from the repository also.  (I updated the issue to include this ask also.)

ZprBot1:
Done in f1b3d3f: one-node-test.sh and capture-test.sh no longer pass --certificate-file/--private-key-file for the zpr-a/-b/-c adapters, and the corresponding create_actor_key_and_cert calls for adapter1-3 are removed (vs.zpr and the node keep theirs). On removing files from the repository: those adapter certs/keys were generated at test runtime in a tmpdir by create_actor_key_and_cert, never committed — integration-test/pregen only carries the actor*-rsa.key bootstrap keys, the node cert/key, and the CA pair, all still in use. So there is nothing to delete; if you had specific files in mind, point me at them. Full gate (build, fmt, test, -D warnings) is clean locally.

ZprBot1:
CI fix in a5b22ac: dropping --certificate-file left ph with no CN source, so the zpr-a/-b/-c adapters failed argument parsing ("certificate_file or name (required for CN derivation)") and the 1-node/capture tests timed out. The tests now pass --name adapter1/2/3 (the CNs the removed certs carried). Verified locally that ph parses past CN derivation with --name; full gate (build, fmt, test, -D warnings) clean.

ZprBot1:
The 1-node and capture test failures at a5b22ac were a real bug this PR exposed, fixed in 016d1c2: with self-generated keys the node's SA has no peer cert, but `check_self_signed_blob` still hard-required one to bind the blob CN, so `AcquireZprAddress` failed ("no peer cert found, cannot validate blob") and the adapters timed out in Helloing. `verify_blob_challenge` now takes `Option<&X509Certificate>`: the cert-CN match still runs when a cert was presented; without one it is skipped and the blob CN is authenticated by the VS RSA signature check, as it already was. HMAC and age checks are unchanged. Local gate clean (build, fmt, 225 tests, -D warnings); CI re-running.

ZprBot1:
Third CI round, fixed in 8f69909: 016d1c2 let the acquire-address blob through without a peer cert, but build_connect_request still sourced the device.zpr.adapter.cn claim only from the peer certificate — with self-generated keys the claim went out empty and the VS rejected authorize_connect ("device.zpr.adapter.cn claim cannot be empty"), stalling the adapters at RegisterAA. It now falls back to the CN asserted in the self-signed auth blob, which the VS authenticates against the blob's RSA signature. Local gate (build, fmt, 225 tests, -D warnings) clean; watching the integration jobs.

cpacejo:
> On removing files from the repository: those adapter certs/keys were generated at test runtime in a tmpdir by create_actor_key_and_cert, never committed — integration-test/pregen only carries the actor*-rsa.key bootstrap keys, the node cert/key, and the CA pair, all still in use. So there is nothing to delete; if you had specific files in mind, point me at them. Full gate (build, fmt, test, -D warnings) is clean locally.

Ah good catch.  I was mistaking the Noise keys for the bootstrap keys.  You are correct there is nothing to delete.

adapter/ph/src/visa_mgmt.rs
cpacejo, line 1: I think we want to do something different here.  Going forward the SA cert/CN and the AuthBlob cert/CN should have no relation to each other:  the SA cert should be used only for validating the node-adapter link, whereas the AuthBlob cert is used for authenticating an actor to the VS.  (And in the future there may be multiple actors behind a single adapter, hence the CNs _must_ be different in this case.)

So, I think the right thing to do here, is to _always_ use the CN from the AuthBlob cert in this function (ignore the SA cert/CN entirely).  If we don't have an AuthBlob cert (that is, the AuthBlob::AuthCode case), then we simply don't include the CN as a claim to the VS.

Let me know what you think & implement this change if you agree.

cpacejo (changes requested):
This all looks good, good finds with the issues I missed.  I think we want to do something slightly different in visa_mgmt which I've outlined in a comment.

adapter/ph/src/visa_mgmt.rs
ZprBot1: Agreed — implemented in 6b3bc01. `build_connect_request` now sources the CN claim exclusively from the AuthBlob: for `SelfSigned` blobs the claim carries the blob cert CN, and for `AuthCode` blobs no CN claim is sent at all. The SA cert/CN is ignored entirely here (the now-unused `get_common_name` helper is removed), so the SA cert is used only for validating the node-adapter link. Full local gate (build, fmt, 225 tests, -D warnings) passes.

cpacejo (approved):
Perfect thanks!

PR #1381, opened 2026-08-26
Node sends its advertised substrate address to the VS on connect/open (#1375)

### LINKED ISSUE #1375: Node sends its advertised substrate address to the VS on connect/open
Part of org-zpr/zpr-visaservice#299 (node reports its advertised substrate address to the VS). **Comes after** the zpr-common `pname::SUBSTRATE_ADDR` constant (see sibling zpr-common issue) and the VS-side changes; once this ships, node and VS must deploy together — the VS requires the param on both `connect` and `open`.

## Context

The value sent is the node's **advertised** substrate address — the address peers dial, not necessarily the bind address (they differ behind NAT). It is trusted operator config; the VS cross-checks it against policy topology (mismatch → `paramError`) and rejects unspecified IPs.

## Changes

**New config option** `advertised_substrate_addr` (optional, `"host:port"` string, next to `self_addr` in `adapter/ph/src/config.rs:125-127`). Accepts hostname or IP; resolved once at startup via `ToSocketAddrs` (policy `NetAddr` supports hostnames, so operators can write the same string in both places). Required in practice for NAT'd nodes — the mapped public address/port cannot be auto-discovered.

**Advertised-address resolution order** (compute once at startup; fail startup if none yields a usable address):
1. `advertised_substrate_addr` config, if set.
2. `self_addr`, if its IP is specified — the common LAN/single-host case, no extra config needed.
3. Last resort for wildcard binds: local address of the VSAPI TCP socket (`stream.local_addr()`) combined with the dock port. Correct only when node and peers share a network; a NAT'd node must use the config option.

**Wiring:**
- `libnode2/src/vsconn.rs`: add `substrate_addr: SocketAddr` (the advertised value) to `NodeConnect` (~59-64); include `Param::new_str(pname::SUBSTRATE_ADDR.into(), addr.to_string())` in the connect params (~610-620); `do_open` (~700-712) sends `params: Some(vec![substrate_addr param])` instead of `None`.
- Plumb the advertised address into `VSConn` at `adapter/ph/src/main.rs:507-511`, and through the `NodeConnect` construction sites (`adapter/ph/src/vs_worker.rs:27-30`, `libnode2/src/cli/handler.rs:43-46`).

## Verification

- Single-node deploy: VS stores the node's dock `ip:port`, not the TCP peer address with an ephemeral port.
- Set `advertised_substrate_addr` different from the bind address; confirm the advertised value is what the VS stores.
- Misconfigure the advertised address vs policy topology; confirm connect is rejected with `paramError`.

See the full plan in org-zpr/zpr-visaservice#299.

ZprBot1:
Implements org-zpr/zpr-core#1375 (node side of org-zpr/zpr-visaservice#299). The VS side is merged and requires the `substrate_addr` param on both `connect` and `open`, so node and VS must deploy together once this lands.

## Changes

- **Dep bump**: `zpr` v0.25.0 → v0.25.1 (picks up `pname::SUBSTRATE_ADDR` from zpr-common#65). `cargo update -p zpr@0.25.0 --precise 0.25.1`; the lockfile also deduped a few `windows-sys` entries as a side effect (Windows-only, no Linux build impact).
- **New config option** `advertised_substrate_addr` (optional, `"host:port"` string) in a new `[node]` section of the node config file, per the issue follow-up comment. Hostnames accepted; resolved once at config-parse time via `ToSocketAddrs`. Unresolvable values, unspecified IPs, and port 0 are rejected at startup.
- **Resolution order** (computed once at node startup in `adapter/ph/src/main.rs`):
  1. `advertised_substrate_addr`, if set;
  2. `self_addr`, if its IP is specified (common LAN/single-host case);
  3. deferred to `VSConn`: local address of the VSAPI TCP socket + the dock port, captured on each (re)connect. Correct only when node and peers share a network; a NAT'd node must use the config option.
  Startup fails (with a message pointing at the config option) when `self_addr` is wildcard **and** its port is 0, since fallback 3 could never form a usable address.
- **libnode2**: `VSConn::new` gains `advertised_substrate: Option<SocketAddr>` and `dock_port: u16`; the run loop computes the effective substrate after TCP connect; `do_connect` sends `Param::new_str(pname::SUBSTRATE_ADDR, ...)` alongside `ZPR_ADDR`; `do_open` now sends `params: Some(vec![substrate_addr])` instead of `None`.
- `lntest` passes `(None, 0)` → advertises its own socket address (test harness).

## Deviation from the issue text (flagged in the plan comment before implementing)

The issue wires `substrate_addr: SocketAddr` through `NodeConnect` and its construction sites (`vs_worker.rs`, `cli/handler.rs`). Instead the value is held inside `VSConn`, set at construction:

- `do_open` needs the same value and `NodeOpen` carries nothing;
- fallback 3 needs the TCP stream's `local_addr()`, which the `NodeConnect` construction sites don't have;
- keeping it in one place guarantees connect and open advertise the same address.

`NodeConnect`/`NodeOpen` and their construction sites are unchanged. Happy to rework to the issue's wiring if preferred.

## Verification

- `cargo build --all-targets`, `cargo build -p libnode2 --all-features`, `cargo fmt --check`, `cargo test` (229 passed + new config tests), `-D warnings` build: all clean.
- New unit tests: `[node]` section TOML parsing, `resolve_advertised_addr` (IP pass-through, hostname resolution, rejection of unspecified/port-0/garbage).
- End-to-end checks from the issue (VS stores dock `ip:port`; advertised ≠ bind honored; policy mismatch → `paramError`) need a node+VS deploy, which this environment doesn't have — they should be covered when this and the VS side are deployed together.

### DIFF
--- a/adapter/ph/src/config.rs
+++ b/adapter/ph/src/config.rs
@@ -640,7 +640,7 @@ pub struct NodeConfig {
 #[derive(Deserialize, Debug, Clone)]
 pub struct NodeConfigSection {
     /// The substrate address this node advertises to the visa service, as a
-    /// `"host:port"` string. Hostnames are resolved once at parse time.
+    /// literal `"IP:port"` string (IPv6 in square-bracket notation).
     pub advertised_substrate_addr: Option<String>,
 }
 
@@ -734,25 +734,19 @@ fn check_file_exists(desc: &str, path: &Path) -> Result<(), ArgsError> {
     }
 }
 
-/// Resolve an `advertised_substrate_addr` config string (`"host:port"`, host may be a
-/// hostname or an IP) to a concrete [SocketAddr]. Resolution happens once, here at
-/// config-parse time. Unresolvable strings, unspecified IPs (e.g. `0.0.0.0`) and
-/// port 0 are rejected: the value is what peers dial, so it must be concrete.
+/// Parse an `advertised_substrate_addr` config string to a concrete [SocketAddr].
+/// The value must be a literal `IP:port` — IPv6 in square-bracket notation, e.g.
+/// `"[2001:db8::1]:5000"`. Hostnames are deliberately not accepted: substrate
+/// claims are always for a specific IP address and adapters dial nodes by IP, so
+/// the advertised value must be the concrete address itself. Unspecified IPs
+/// (e.g. `0.0.0.0`) and port 0 are rejected: the value is what peers dial.
 fn resolve_advertised_addr(addr_str: &str) -> Result<SocketAddr, ArgsError> {
-    use std::net::ToSocketAddrs;
-    let resolved = addr_str
-        .to_socket_addrs()
-        .map_err(|e| {
-            ArgsError::ParseError(format!(
-                "failed to resolve advertised_substrate_addr {addr_str:?}: {e}"
-            ))
-        })?
-        .next()
-        .ok_or_else(|| {
-            ArgsError::ParseError(format!(
-                "advertised_substrate_addr {addr_str:?} resolved to no addresses"
-            ))
-        })?;
+    let resolved: SocketAddr = addr_str.parse().map_err(|e| {
+        ArgsError::ParseError(format!(
+            "advertised_substrate_addr {addr_str:?} is not a literal IP:port \
+             (IPv6 in square brackets, e.g. \"[2001:db8::1]:5000\"): {e}"
+        ))
+    })?;
     if resolved.ip().is_unspecified() {
         return Err(ArgsError::ParseError(format!(
             "advertised_substrate_addr {addr_str:?} is an unspecified address; peers cannot dial it"
@@ -837,15 +831,27 @@ mod test {
             resolve_advertised_addr("203.0.113.7:5000").unwrap(),
             "203.0.113.7:5000".parse::<SocketAddr>().unwrap()
         );
-        // Hostname resolves (localhost is safe to assume in CI).
-        let resolved = resolve_advertised_addr("localhost:5000").unwrap();
-        assert_eq!(resolved.port(), 5000);
-        assert!(resolved.ip().is_loopback());
+        // V6 in square-bracket notation passes through.
+        assert_eq!(
+            resolve_advertised_addr("[2001:db8::7]:5000").unwrap(),
+            "[2001:db8::7]:5000".parse::<SocketAddr>().unwrap()
+        );
+        // V6 with a scope-free full form.
+        assert_eq!(
+            resolve_advertised_addr("[2001:db8:0:0:0:0:0:9]:6000").unwrap(),
+            "[2001:db8::9]:6000".parse::<SocketAddr>().unwrap()
+        );
+        // V6 without brackets rejected.
+        assert!(resolve_advertised_addr("2001:db8::7:5000").is_err());
+        // Hostnames rejected: the claim is always for a specific IP address.
+        assert!(resolve_advertised_addr("localhost:5000").is_err());
+        assert!(resolve_advertised_addr("node1.example.com:5000").is_err());
         // Unspecified IP rejected.
         assert!(resolve_advertised_addr("0.0.0.0:5000").is_err());
         assert!(resolve_advertised_addr("[::]:5000").is_err());
-        // Port 0 rejected.
+        // Port 0 rejected (V4 and V6).
         assert!(resolve_advertised_addr("203.0.113.7:0").is_err());
+        assert!(resolve_advertised_addr("[2001:db8::7]:0").is_err());
         // Garbage rejected.
         assert!(resolve_advertised_addr("not an address").is_err());
         assert!(resolve_advertised_addr("203.0.113.7").is_err());
--- a/adapter/ph/src/config.rs
+++ b/adapter/ph/src/config.rs
@@ -126,6 +126,11 @@ pub struct Config {
     /// For a node this is the nodes dock listening address.
     pub self_addr: SocketAddr,
 
+    /// Node only, optional - the substrate address this node advertises to the visa
+    /// service (the address peers dial). Needed when it differs from the bind address,
+    /// e.g. behind NAT. Resolved from the config-file string at parse time.
+    pub advertised_substrate_addr: Option<SocketAddr>,
+
     /// Path to a PEM file containing the Certificate Authority certificate.
     /// Optional and if present is used by the link management system to verify passed noise certificates.
     pub ca_file: Option<PathBuf>,
@@ -251,6 +256,7 @@ impl Config {
             let base_dir = config_file.config_path.parent().unwrap();
             config.set_from_globals(&config_file.global, base_dir)?;
             config.set_from_authentication(&config_file.authentication, base_dir)?;
+            config.set_from_node(&config_file.node)?;
         }
         Ok(config)
     }
@@ -478,6 +484,17 @@ impl Config {
         Ok(())
     }
 
+    // Overwrite our internal state with the values present in the node section
+    // of TOML config.
+    fn set_from_node(&mut self, config: &Option<NodeConfigSection>) -> Result<(), ArgsError> {
+        if let Some(config) = config {
+            if let Some(advertised) = &config.advertised_substrate_addr {
+                self.advertised_substrate_addr = Some(resolve_advertised_addr(advertised)?);
+            }
+        }
+        Ok(())
+    }
+
     // Overwrite our internal state with the values present in the CommonArgs (from command line)
     fn set_from_common(&mut self, common: &CommonArgs) -> Result<(), ArgsError> {
         if let Some(control_path) = &common.control_path {
@@ -574,6 +591,7 @@ impl Default for Config {
             certificate_file: None,
             private_key_file: None,
             private_key_data: None,
+            advertised_substrate_addr: None,
             auth_private_key: None,
             tun_if: None,
             logging: Vec::new(),
@@ -607,7 +625,16 @@ pub struct NodeConfig {
     pub config_path: PathBuf,
 
     pub global: GlobalConfigSection,
+    pub node: Option<NodeConfigSection>,
     pub authentication: Option<AuthenticationConfigSection>,
+}
+
+// Node only bits.
+#[derive(Deserialize, Debug, Clone)]
+pub struct NodeConfigSection {
+    /// The substrate address this node advertises to the visa service, as a
+    /// `"host:port"` string. Hostnames are resolved once at parse time.
+    pub advertised_substrate_addr: Option<String>,
 }
 
 // Global section is shared by nodes and adapters.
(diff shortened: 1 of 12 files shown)


libnode2/src/vsconn.rs
```diff
@@ -608,6 +646,19 @@ impl VSConn {
...
     }
 
+    /// The `substrate_addr` param the VS requires on both `connect` and `open`.
+    /// Set by [VSConn::run] before any command is processed; the error arm is
+    /// defensive.
+    fn substrate_addr_param(&self) -> Result<Param, VSApiError> {
+        let addr = self.substrate_addr.get().ok_or_else(|| {
+            VSApiError::CommandFailed("substrate address not yet determined".to_string())
+        })?;
+        Ok(Param::new_str(
```
mkolehmainen: Add a debug log either here or at another reasonable location so that we can check the debug logs to know what substrate address the node is advertising.

mkolehmainen (changes requested):
Add a helpful log mesasge.

libnode2/src/vsconn.rs
ZprBot1: Done in 5cefdc5: a debug log now records the advertised substrate address when it is determined, at the point the effective address is computed in VSConn::run. The configured-address case logs the address directly; the derived case also logs the local socket and dock port it was derived from, so a wrong advertisement is traceable to its source. Full gate (build, fmt, 229 tests, -D warnings) clean.

mkolehmainen:
@codex review

libnode2/src/vsconn.rs
```diff
@@ -295,6 +316,34 @@ impl VSConn {
...
+                addr
+            }
+            None => {
+                let local = sock.local_addr()?;
+                let port = if self.dock_port != 0 {
+                    self.dock_port
+                } else {
+                    local.port()
+                };
+                let addr = SocketAddr::new(local.ip(), port);
```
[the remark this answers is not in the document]
ZprBot1: Fixed in 58ff816. You are right — the VSAPI TCP connection rides the ZPR network, so its local IP is the node's ZPR/TUN address and can never serve as a substrate fallback. The fallback is removed entirely: VSConn now takes a mandatory SocketAddr, and ph fails startup when self_addr has an unspecified IP and advertised_substrate_addr is unset, pointing the operator at the [node] config option. Full gate (build, fmt, 229 tests, -D warnings) clean.

libnode2/src/bin/lntest.rs
```diff
@@ -27,7 +27,7 @@ async fn main() {
     let private_key = libnode2::cli::crypto::load_private_key(&args.private_key)
         .expect("failed to load private key");
 
-    let mut vsc = VSConn::new(8, vs_sa, args.node_cn, private_key);
+    let mut vsc = VSConn::new(8, vs_sa, args.node_cn, private_key, None, 0);
```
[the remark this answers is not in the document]
ZprBot1: Fixed in 58ff816. With the socket-derived fallback removed from VSConn, lntest now takes a required --substrate-addr argument (validated to be non-wildcard with a non-zero port) and passes it straight through, instead of advertising the ephemeral VSAPI source port. The README example is updated accordingly.

libnode2/src/vsconn.rs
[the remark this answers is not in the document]
mkolehmainen: @ZprBot1 address this issue.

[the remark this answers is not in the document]
ZprBot1: Addressed in 58ff816 (the current head, which CI is running against). The VSAPI-socket fallback is removed entirely: VSConn now requires a concrete SocketAddr, and ph fails startup with a pointer to [node].advertised_substrate_addr when self_addr is wildcard and no advertised address is configured — so a ZPR/TUN address can no longer be advertised as the substrate endpoint. Full gate (build, fmt, 229 tests, -D warnings) clean.

ZprBot1:
The Capture Test failure at 58ff816 was real fallout from removing the wildcard fallback: the integration tests start the node with `--self-addr 0.0.0.0:12345` (capture test) or the wildcard default (1-node test), so the new startup check correctly refused to pick an advertised address. Fixed in ce0143b: `ph node` gains an `--advertised-substrate-addr` CLI option (same resolve/validate path as the `[node]` config key, CLI overrides config), and both test scripts now pass the node's VS-facing substrate address (10.0.0.1:12345 for capture-test, 10.0.0.1:5000 for one-node-test). Full local gate (build, fmt, test, -D warnings) is clean; CI is re-running.

mkolehmainen:
Ok looks good to me. Am asking for @cpacejo to take a look too.

mkolehmainen (dismissed):
Looks good to me. Waiting for feedback from chris.

cpacejo:
I guess my main question on this is do we want nodes to be able to advertise multiple substrate addresses?  We did a bunch of work in the PH to allow nodes to accept and respond to connections on multiple physical substrate addresses, and this would be necessary to allow a node to replace an IP router (which I understood to be a goal of ZPR routing).  (Else, to bridge substrate IP subnets, we'd require an IP router also.)

EDIT: Ah yes I'm seeing this is optional and only needed for NATted nodes.  So I guess the question is do we want to allow NATted nodes to span multiple substrate subnets but I guess that's not a necessary thing.

ZprBot1:
Good question — laying out where the implementation stands so the decision has the facts in front of it:

**Today it is single-address end-to-end, per the #1375 / org-zpr/zpr-visaservice#299 design.** The wire format is one string param (`zpr.substrate_addr`) on `connect` and `open`; the merged VS side parses exactly one `SocketAddr` (rejecting unspecified IPs and port 0), cross-checks it against the policy topology's declared substrate when one exists, and stores it as the node's single `key::SUBSTRATE_ADDR` claim. This PR computes one advertised address at node startup. So even where the PH can accept/respond on multiple physical substrate addresses, peers are only ever told to dial the one advertised address — a node standing in for an IP router would indeed need more than that.

**Multi-address would be an additive extension, and nothing here forecloses it:**
- Wire: repeat the `SUBSTRATE_ADDR` param or add a list variant. (Note the current VS parsers take a single value and do not collect repeats — `connect` keeps the last occurrence, `open` the first — so that loop is the natural place to start if we widen it.)
- VS: `key::SUBSTRATE_ADDR` becomes multi-valued (or one claim per address), the topology cross-check becomes per-address, and whatever hands the address to dialing peers hands out the set.
- Node: `advertised_substrate_addr` config becomes a list; the fallbacks (`self_addr`, VSAPI-socket local address) still yield a single-entry list.

We're pre-release with breaking changes expected, so shipping single-address now and widening to a list when the router use case lands is a viable path — but now-vs-later is a design call for you and @mkolehmainen. If the answer is "now", I'd suggest a follow-up issue spanning zpr-visaservice + zpr-core rather than growing this PR, since the VS side is already merged with the single-address contract.

adapter/ph/src/config.rs
```diff
@@ -607,9 +632,18 @@ pub struct NodeConfig {
...
+    pub node: Option<NodeConfigSection>,
     pub authentication: Option<AuthenticationConfigSection>,
 }
 
+// Node only bits.
+#[derive(Deserialize, Debug, Clone)]
+pub struct NodeConfigSection {
+    /// The substrate address this node advertises to the visa service, as a
+    /// `"host:port"` string. Hostnames are resolved once at parse time.
+    pub advertised_substrate_addr: Option<String>,
```
cpacejo, line 644: Why do we allow for hostnames here?  The claim will always be for a specific IP address, and adapters connect to nodes by specifying an IP address.  It feels odd to allow a node to specify its _own_ IP address by supplying a hostname.

adapter/ph/src/config.rs
```diff
@@ -723,4 +789,65 @@ mod test {
             "#;
         let _config: AdapterConfigSection = toml::from_str(toml_str).unwrap();
     }
+
+    #[test]
+    fn test_deserialize_node_config_section() {
+        let toml_str = r#"
+            advertised_substrate_addr = "203.0.113.7:5000"
```
cpacejo, line 790: Let's test V6 addresses also.  They should be specified using square-bracket notation.

adapter/ph/src/main.rs
```diff
@@ -500,11 +500,40 @@ fn main() -> ExitCode {
         let auth_private_key = zpr_utils::rsa_sign::load_rsa_key(auth_key_pem.as_bytes())
             .unwrap_or_else(|e| panic!("failed to parse auth_private_key {auth_key_path:?}: {e}"));
 
+        // Advertised substrate address resolution, in order:
+        //   1. advertised_substrate_addr config, if set.
+        //   2. self_addr, if its IP is specified (common LAN/single-host case).
+        // A wildcard self_addr with no configured advertised address is a startup
+        // error: the VSAPI socket traverses the ZPR network, so its local IP is the
+        // node's ZPR/TUN address and cannot serve as a substrate fallback.
```
cpacejo: This comment is incorrect.  The VSAPI socket's local IP is just one of potentially many local IPs assigned to the local machine, which a wildcard self_addr would bind to.  (And only happens to be present due to the current ordering of things in main which is somewhat arbitrary.)

I think the true issue is that there may be more than one local IP, but we can advertise only one.  (And furthermore, that the wildcard also binds to _future_ local IPs added, but we have no mechanism to claim those IPs if/when they are added.)

cpacejo (commented):
Overall looks good, just a couple comment nits and a concern about whether we actually need to accept hostnames for this argument.

adapter/ph/src/config.rs
ZprBot1: Agreed — dropped hostname support in 9467007. `resolve_advertised_addr` now requires a literal IP:port via `SocketAddr::parse` (no `ToSocketAddrs`), so the config/CLI value must be the concrete address the claim will be for. The error message tells the operator to use a literal IP (square-bracket notation for V6). Doc comments and CLI help updated to match.

adapter/ph/src/config.rs
ZprBot1: Done in 9467007: added V6 cases to `test_resolve_advertised_addr` — square-bracket notation accepted (`[2001:db8::7]:5000`, and the uncompressed form), bracket-less V6 rejected, plus V6 unspecified (`[::]:5000`) and V6 port-0 rejection.

adapter/ph/src/main.rs
ZprBot1: You are right — corrected in 9467007. The comment (and the matching `VSConn::substrate_addr` doc in libnode2) now states the true issue: a wildcard self_addr binds all local IPs, present and future, of which the machine may have several, but only one address can be advertised, so there is no sound automatic choice and the operator must set advertised_substrate_addr. The socket-derived phrasing is gone.

mkolehmainen:
I agree with the just one address for now.  The TLV format supports multiple, the ZPLC (configuration) format for links supports multiple addresses.  At the moment we just don't support a node actually advertising multiple addresses.  Will create a follow up issue.

mkolehmainen (approved):
Great!

PR #1383, opened 2026-08-27
deferred vs connect 

ianchang1:
Implementing Deferred VS connect. This is zpr core side. Step 3 I am going to finish in VS

Step 1

- build_connect_request stops returning Ok(None) for the VS link —
- The "is this the VS link?" check moved out to the caller in link_state, which already had to branch on exactly that.
- The request gets parked on Assembly as Mutex<Option<(LinkId, ConnectRequest)>> and the producer (link state machine) and consumer (vs_worker) have no other way to talk.
- Startup self-authorization is byte-for-byte identical

Step 2

- After the node connects and registers its VSS endpoint, it sends the parked request.
- Ordering is guaranteed rather than hoped for since VSAPI runs over ZPR, so the VS adapter link is up and granted before the node's connect can succeed. 
- Deliberately does not reuse authorize_connect: on failure that fires LinkEvent::Error, which on the VS link tears down the VSConn. New send-and-log function instead.

Also updated policy files confirmed with Mathias should be vs.zpr


adapter/ph/src/visa_mgmt.rs
```diff
@@ -13,6 +12,30 @@ use std::num::NonZero;
...
+    match asm
+        .vsconn
+        .as_ref()
+        .unwrap()
+        .authorize_connect(connect_req)
+        .await
+    {
+        Ok(cr) => {
+            info!(target: VISA_MGMT, "{}: visa service adapter registered with the visa service as {}",
+                asm.formatted_link_id(link_id), cr.zpr_addr);
```
cpacejo: Let's check that the returned `zpr_addr` matches the address we requested (since we've already assigned it at this point).

adapter/ph/src/visa_mgmt.rs
```diff
@@ -13,6 +12,30 @@ use std::num::NonZero;
...
+        }
+        Err(e) => {
```
cpacejo: In the error case we should probably bring down the link.  So here we could propagate the error to the caller and have the caller bail out (panicking is fine).

cpacejo (changes requested):
Looks good, just a couple changes I'd make to handle the case that the deferred connection request fails.

cpacejo (approved):
Looks good thanks Ian!

PR #1387, opened 2026-09-02
A2A pubkey testing

ianchang1:
Security testing to demonstrate that the A2A keyed MICV actually protects forwarded traffic from a malicious node. The idea is to have a node try to tamper with a transit packet it's forwarding between two adapters and re-forge the MICV - a correct receiver rejects it, the old unkeyed hash does not. All of this is behind a Cargo feature + CLI flags so it can never ship on by accident.

Step 1
Added a Cargo feature enable-security-testing and three #[cfg]-gated flags

Step 2
Big warning banner at startup (main.rs) whenever the feature is compiled in

Step 3
The mangle itself in fastpath.rs. In forward()'s Node arm, if both ingress and egress links are adapters, call the new security_testing_mangle_ping(). It pulls the five-tuple, compression mode, and the egress adapter's A2A pubkey from the visa (get_egress_a2a_dh_pubkey()), confirms the packet is an ICMP echo request (v4 type 8 / v6 type 128), and overwrites the payload with "YOU HAVE BEEN PWNED" (bails if there's no room). To recompute the MICV it expands a throwaway copy so it hashes the same bytes the receiver will. The forged key is diffie_hellman(node_priv, adapter_pub) run through blake3::derive_key(ZDP_A2A_MICV_KEY_CONTEXT, ...) but the wrong shared secret since the node can't produce the real adapter-to-adapter one.

Step 4
--security-testing-unkeyed-a2a-micv forces the old unkeyed behavior. Added a guard arm at both MICV sites. The node's forge picks the unkeyed hash to match. Has to be set on the node and all adapters otherwise a keyed verifier stares at an unkeyed forgery and just rejects it normally.

Step 5
Integration test integration-test/a2a-pubkey-test.sh, copied from one-node-test. IPv6 (uses the existing v6 ping policy), 2 adapters + VS, node launched with the mangle flag, pings only zpr-a → zpr-b. tcpdump in the zpr-b namespace on tun0 (the decrypted overlay side, so only packets that passed the MICV land there) watches for the injected string and fails the test if it ever arrives. Does not check whether pings succeed.

Step 6
--security-testing-recompute-micvs (default true). With false the node mangles the payload but keeps the original MICV instead of re-forging it, so a receiver whose MICV covers the payload rejects it regardless of keying. 

Ran all three ways against the merged tree: keyed → PASS (forgery rejected), --unkeyed → FAIL (forgery accepted), --unkeyed --reuse → PASS.

### DIFF
--- a/adapter/ph/src/fastpath.rs
+++ b/adapter/ph/src/fastpath.rs
@@ -47,6 +47,18 @@ pub struct FastpathWorkerConfig {
     pub batch_io_engine: &'static BatchIoEngine,
     pub buffer_count: usize,
     pub batch_size: usize,
+
+    /// Security Testing: node corrupts forwarded ICMP echo requests.
+    #[cfg(feature = "enable-security-testing")]
+    pub mangle_forwarded_pings: bool,
+
+    /// Security Testing: force old-style unkeyed A2A MICVs.
+    #[cfg(feature = "enable-security-testing")]
+    pub unkeyed_a2a_micv: bool,
+
+    /// Security Testing: recompute the A2A MICV after mangling (vs. reuse it).
+    #[cfg(feature = "enable-security-testing")]
+    pub recompute_micvs: bool,
 }
 
 pub struct QueuedEgressPacket {
@@ -315,6 +327,8 @@ impl FastpathWorker {
                 let a2a_mac_size = zdp::ZDP_A2A_MAC_SIZE; // TODO: may be smaller depending on A2A SAID
                 let mut a2a_mac = [0u8; zdp::ZDP_A2A_MAC_SIZE];
                 let a2a_micv = match pep.a2a_micv_key {
+                    #[cfg(feature = "enable-security-testing")]
+                    Some(_) if self.config.unkeyed_a2a_micv => blake3::hash(pkt.body()),
                     Some(micv_key) => blake3::keyed_hash(&micv_key, pkt.body()),
                     None => blake3::hash(pkt.body()),
                 };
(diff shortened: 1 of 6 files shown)


adapter/ph/src/main.rs
```diff
@@ -139,6 +139,32 @@ fn main() -> ExitCode {
 
     info!(target: STARTUP, "starting with PID {}", process::id());
 
+    #[cfg(feature = "enable-security-testing")]
+    {
+        error!(
+            target: STARTUP,
+            "\n\
+             ################################################################\n\
```
cpacejo, line 147: LOVE this warning!

adapter/ph/src/fastpath.rs
```diff
@@ -397,6 +427,117 @@ impl FastpathWorker {
...
+        }
+        let echo_data_len = temp.body().len() - echo_data_off;
+        if echo_data_len < PWNED.len() {
+            return; // not enough room; leave the packet alone
+        }
+
+        // Overwrite the echo payload in the real packet -- this is the mangle.
+        let mac_start = body_len - mac_len;
+        let echo_start = mac_start - echo_data_len;
+        pkt.body_mut()[echo_start..echo_start + PWNED.len()].copy_from_slice(PWNED);
```
cpacejo, line 505: Oh!  Running the tests now I realize I forgot to include a step!  It's not crucial, but for completeness we should also recompute the ICMP(v6) checksum field here, or the packets get rejected by the OS before being delivered to the receiving application (ping).

BUT, don't worry about that for this PR, what we're got here is still sufficient to validate that the MICV does what it should (since the packets still show up in the packet capture regardless of the ICMP checksum).  Since today is your last day anyway I will just add this in later!

cpacejo (approved):
This looks great!!  Love seeing this actual demonstration of the keyed hash actually doing its thing end-to-end.

I actually realized there's one minor missing piece (which causes the forged packets to be rejected by the OS before they get delivered to the ping process) but it doesn't actually affect the validity of the tests since the packet capture happens before that, so go ahead and check this in!  Great work.

